PR #1, opened 2015-10-21
Filter out this extra data and leading zeroes in "track" and "disc" tags.

a-mair:


MaxKellermann:
Merged with a few minor corrections

PR #2, opened 2016-02-28
deduplicate identical tags (same tag type and value)

jreinert:
this fixes problems with duplicate values for track/tracknumber, disc/discnumber etc. without side effects.

### DIFF
--- a/src/tag/TagBuilder.cxx
+++ b/src/tag/TagBuilder.cxx
@@ -199,12 +199,15 @@ TagBuilder::AddItemInternal(TagType type, StringView value)
 		value = { f.data, f.size };
 
 	tag_pool_lock.lock();
-	auto i = tag_pool_get_item(type, value);
+	auto item = tag_pool_get_item(type, value);
 	tag_pool_lock.unlock();
 
 	free(f.data);
 
-	items.push_back(i);
+	if (std::find(items.begin(), items.end(), item) == items.end()) {
+		/* the tag isn't a duplicate */
+		items.push_back(item);
+	}
 }
 
 void
--- a/src/tag/TagItem.hxx
+++ b/src/tag/TagItem.hxx
@@ -22,6 +22,7 @@
 
 #include "TagType.h"
 #include "Compiler.h"
+#include <string.h>
 
 /**
  * One tag value.  It is a mapping of #TagType to am arbitrary string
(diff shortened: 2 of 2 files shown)


bastelfreak:
:+1:

carnager:
oh seems to be exactly what I requested once :+1:

zeltak:
+1

MaxKellermann:
This adds overhead by making this O(n^2). If you don't want duplicate tag values, don't store duplicate tag values in the file. MPD is the wrong place to fix this up. Fix your files instead of adding kludges to MPD.

jreinert:
I think you need to see the overhead in context: it adds overhead only if you actually have multiple values for a tag and even _then_ only on db updates.

jreinert:
It's a very small change, and I agree that one shouldn't store duplicate tag values but sometimes you don't have a choice (mpd not being the only player accessing the library etc.) I really don't see the performance as an issue. And if it's not an issue i don't see why this shouldn't just be merged and be done with it. It would solve so many problems.

jreinert:
how would you feel about using std::set instead of std::vector for the values then? This would have the complexity of O(log(n))

MaxKellermann:
> it adds overhead only if you actually have multiple values for a tag
Completely not true. Your code walks the whole tag item list for each new item.

> and even then only on db updates
Also not true. This happens every time a Tag object gets assembled. That happens during database updates, during MPD startup while loading the database, and when a song/stream gets played.

> how would you feel about using std::set instead of std::vector for the values then? This would have the complexity of O(log(n))
This adds even more overhead. The complexity is smaller, but the constant factor is a lot bigger.

And it's still the wrong place to do it. MPD reads the tags as-is, and if the tags are wrong, you see wrong tags - if there are duplicate tags, you'll see duplicate tags. Fix them. No lame excuses.

jreinert:
> > it adds overhead only if you actually have multiple values for a tag
> Completely not true. Your code walks the whole tag item list for each new item.

The whole tag item list being one element long in 99.9% of the cases. I'd say that's neglegtable performance wise. Your only reason for not merging is being unwilling to fix problems that don't have their roots in mpd per se. I understand your position, but I honestly fail to see any harm in adding support for a problem many mpd users experience when there are virtually no downsides. 

I'd say most if not all CSV-parsers support using other separators than `,` for example even though that's not real CSV. Same thing.

> MPD reads the tags as-is, and if the tags are wrong, you see wrong tags - if there are duplicate tags, you'll see duplicate tags.
This is _not_ true. You're already ignoring "album artist" tags for example.

jreinert:
I'm happy to provide benchmarks to prove my point if that will persuade you.

MaxKellermann:
No, I'm not interested at all. Let's stop discussing this.

PR #3, opened 2016-04-02
Provide audio format to pipe output command

frankl-audio:
So far, the pipe output of MPD had the problem that the commands
in the pipe had no information about the format of the raw audio data
sent into the pipe.

With this patch mpd substitutes the strings "$MPDPIPE_BITS",
"$MPDPIPE_CHANNELS", "$MPDPIPE_RATE", "$MPDPIPE_FORMAT"
within the command started for the pipe by the number of bits per sample,
the number of channels, the sample rate, and the sample format (one of
"S8", "S16", "S24_P32" or "S32"), respectively.

Example: if in mpd.conf an audio_output of type pipe contains the line

        command         "/usr/local/bin/mympdpipe $MPDPIPE_BITS $MPDPIPE_CHANNELS $MPDPIPE_RATE $MPDPIPE_FORMAT"

then mpd starts a command like
        /usr/local/bin/mympdpipe 16 2 44100 S16
for this pipe.

### DIFF
--- a/doc/user.xml
+++ b/doc/user.xml
@@ -3076,7 +3076,7 @@ buffer_size: 16384</programlisting>
                   <parameter>CMD</parameter>
                 </entry>
                 <entry>
-                  This command is invoked with the shell.
+                  This command is invoked with the shell. The command may contain one or more of the strings $MPDPIPE_BITS, $MPDPIPE_CHANNELS, $MPDPIPE_RATE, $MPDPIPE_FORMAT. Before the command is invoked MPD will substitute these strings by the number of bits per sample, the number of channels, the sample rate, and the sample format (one of "S8", "S16", "S24_P32", "S32", "FLOAT" or "DSD"), respectively. 
                 </entry>
               </row>
             </tbody>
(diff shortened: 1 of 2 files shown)


frankl-audio:
Various people seem to use a patched version of MPD to resolve the problem of missing information about the audio format within the pipe command.

A very similar patch like this one was for example proposed here:
     https://forum.musicpd.org/viewtopic.php?f=11&t=3820

MaxKellermann:
Wrong indentation, excessive use of `std::string`, duplicate code, and I'd rather pass environment variables instead of this obscure string replacement.

frankl-audio:
I'm fine with suggestions for better solutions. 

But why is this request closed?

Are you going to implement a better solution of the problem addressed here?

Or are you not interested in solving this problem (which prevents people from using mpd together with any processing of the audio data, e.g., a filter for room correction)?

MaxKellermann:
The pipe output is the wrong tool to apply filters, so your example doesn't convince me that this is necessary at all. And no, I'm not going to implement this - there are tons of other things with a higher priority, and the pipe plugin has no priority at all for me. I told you how I'd do it, if ever, and that would be how I'd accept a pull request.

PR #4, opened 2016-04-05
Provide audio format to pipe output command

frankl-audio:
So far, the pipe output of MPD had the problem that the commands
in the pipe had no information about the format of the raw audio data
sent into the pipe.

With this patch mpd sets the environment variables
    MPDPIPE_BITS,
    MPDPIPE_CHANNELS,
    MPDPIPE_RATE
to the number of bits per sample,
the number of channels, and the sample rate of the raw PCM sent to the
pipe, respectively. These variables can then be used in the command or
within a script which is started for the pipe.

### DIFF
--- a/src/output/plugins/PipeOutputPlugin.cxx
+++ b/src/output/plugins/PipeOutputPlugin.cxx
@@ -28,7 +28,11 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#ifndef WIN32
 #include <unistd.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#endif
 
 class PipeOutput {
 	friend struct AudioOutputWrapper<PipeOutput>;
@@ -37,6 +41,9 @@ class PipeOutput {
 
 	std::string cmd;
 	FILE *fh;
+#ifndef WIN32
+	pid_t childpid;
+#endif
 
 	PipeOutput()
 		:base(pipe_output_plugin) {}
@@ -49,7 +56,13 @@ public:
 	bool Open(AudioFormat &audio_format, Error &error);
 
 	void Close() {
+#ifdef WIN32
 		pclose(fh);
+#else
+		int status;
+		fclose(fh);
+		waitpid(childpid, &status, 0);
+#endif
 	}
 
 	size_t Play(const void *chunk, size_t size, Error &error);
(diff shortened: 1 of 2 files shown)


frankl-audio:
This is  a nicer version to achieve the same as the rejected pull request #3.

Thanks for the hint to set environment variables; I was not aware of the rules for the environment of a process started with `popen` but checked that it will always see the variables set here.

The code works fine for me (and is indeed much nicer than in #3), an adjustment of the documentation is included.

MaxKellermann:
This implementation is not acceptable, because it leaks the environment variables to the MPD process as well. The problem is, of course, that `popen()` doesn't allow setting environment variables for the child process. This is where you should use `fork()`/`execve()` instead of `popen()` if available.

frankl-audio:
I cannot see any  problem with setting these environment variables for the MPD process itself.
Nevertheless, with the new commit the environment is only changed for the child process.

MaxKellermann:
Your code will not compile on Windows, because Windows has no fork().

frankl-audio:
I wasn't aware that MPD can be used on Windows. I don't have any Windows system and do not know how to find out what is available there. (Have only used cygwin which does have fork().)

You didn't mention Windows after commit d1f170e9b9fe6e. Does that work on Windows?
I need a suggestion how to proceed. 
From other files I now guess that Windows specific code can be specified with the help of `#ifdef WIN32`, is this what should be used?

MaxKellermann:
I wrote "This is where you should use `fork()`/`execve()` instead of `popen()` **if available**". You missed the last part. Yes, `#ifdef WIN32` would be an acceptable compile-time test.

frankl-audio:
Ok, thanks for the hint.

src/output/plugins/PipeOutputPlugin.cxx
```diff
@@ -84,12 +97,30 @@ PipeOutput::Create(const ConfigBlock &block, Error &error)
 	return po;
 }
 
+#define set_MPDPIPE_vars(audio_format) {\
```
MaxKellermann: Don't write preprocessor macros when you can write C++.

src/output/plugins/PipeOutputPlugin.cxx
```diff
@@ -84,12 +97,30 @@ PipeOutput::Create(const ConfigBlock &block, Error &error)
...
+	setenv("MPDPIPE_RATE", strbuf, 1);\
+	snprintf(strbuf, sizeof(strbuf), "%u", audio_format.channels);\
+	setenv("MPDPIPE_CHANNELS", strbuf, 1);\
+}
+
 inline bool
 PipeOutput::Open(AudioFormat &audio_format, Error &error)
 {
+#ifdef WIN32
+	set_MPDPIPE_vars(audio_format);
```
MaxKellermann: This sets environment variables in the parent process, which I already told you is not acceptable. (Don't argue with me again. I don't accept it.)

src/output/plugins/PipeOutputPlugin.cxx
```diff
@@ -84,12 +97,30 @@ PipeOutput::Create(const ConfigBlock &block, Error &error)
...
+	fh = popen(cmd.c_str(), "w");
```
MaxKellermann: This commit re-adds code that has been removed before. Since I don't merge your original (bad) commit, I can't merge this one, because it conflicts.

MaxKellermann:
It's been 3 months without a reply. Closing for now.

PR #6, opened 2016-04-12
support sacdiso  ,solve the playback dsd, dff,sacd-iso file incorrect speed,background white noise.

hicaoc:
Play music under DSD_U32_BE mode dsd x4 faster , this modified dsd music sample_rate divided by four.
In addition , under DSD_U32_BE mode , background white noise ,four byte order reversed , this should be a bug, together with amendments submitted , we have not tested the compatibility of other DAC .

The request also includes support for the other contributors of sacdiso

I github newcomers are trying to learn in.  
If there is something wrong place , very much like to get corrected . thank you very much


MaxKellermann:
This pull request contains a big load of commits which are not relevant for your goal. Please remove them. If you really want to resubmit the SACD plugin, clean up the code first. The code is bad, and I rejected merging it already long ago, but the author never fixed it.

PR #7, opened 2016-04-20
Added decoder plugin for usf/miniusf (Nintendo 64 sound files)

habbbe:
This includes a decoder plugin for the miniusf sound format, which is used for storing information about music and sounds from Nintendo 64 games.

The decoder uses the lazyusf library which can be found at
https://bitbucket.org/kode54/lazyusf2

### DIFF
--- a/src/decoder/plugins/UsfDecoderPlugin.cxx
+++ b/src/decoder/plugins/UsfDecoderPlugin.cxx
@@ -1,0 +1,318 @@
+/*
+ * Copyright (C) 2003-2015 The Music Player Daemon Project
+ * http://www.musicpd.org
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "config.h"
+#include "UsfDecoderPlugin.hxx"
+#include "../DecoderAPI.hxx"
+#include "../DecoderInternal.hxx"
+#include "tag/TagHandler.hxx"
+#include "fs/Path.hxx"
+#include "util/Domain.hxx"
+#include "Log.hxx"
+#include "util/ScopeExit.hxx"
+
+#include <usf.h>
+#include <psflib.h>
+
+static constexpr Domain usf_domain("usf");
+static constexpr unsigned USF_BUFFER_FRAMES = 2048;
+static constexpr unsigned USF_CHANNELS = 2;
+static constexpr unsigned USF_BUFFER_SAMPLES = USF_BUFFER_FRAMES*USF_CHANNELS;
+
+static void *
+stdio_fopen( const char * path )
+{
+    return fopen( path, "rb" );
+}
+
+static size_t
+stdio_fread( void *p, size_t size, size_t count, void *f )
+{
+    return fread( p, size, count, (FILE*) f );
+}
+
+static int
+stdio_fseek( void * f, int64_t offset, int whence )
+{
+    return fseek( (FILE*) f, offset, whence );
+}
+
+static int
+stdio_fclose( void * f )
+{
+    return fclose( (FILE*) f );
+}
+
+static long
+stdio_ftell( void * f )
+{
+    return ftell( (FILE*) f );
+}
+
+static constexpr psf_file_callbacks stdio_callbacks =
+{
+    "\\/:",
+    stdio_fopen,
+    stdio_fread,
+    stdio_fseek,
+    stdio_fclose,
+    stdio_ftell
+};
+
+struct UsfLoaderState
+{
+    bool enableCompare; // The _enablecompare tag is present in the file; passed to usf_set_compare
+    bool enableFIFOFull; // The _enableFIFOfull tag is present in the file; passed to usf_set_fifo_full
+    void * emu; // The emulator state
+};
+
+static int
+usf_loader(void * context, const uint8_t * exe, size_t exe_size,
+               const uint8_t * reserved, size_t reserved_size)
+{
+    struct UsfLoaderState * state = ( struct UsfLoaderState * ) context;
+
+    if ( exe && exe_size > 0 ) return -1;
+
+    return usf_upload_section( state->emu, reserved, reserved_size );
+}
+
+static int
+usf_info(void * context, const char * name, const char * value)
+{
+    struct UsfLoaderState * state = ( struct UsfLoaderState * ) context;
+    if (!value) return 0;
+
+    if ( strcmp( name, "_enablecompare" ) == 0)
+        state->enableCompare = 1;
+    else if ( strcmp( name, "_enableFIFOfull" ) == 0)
+        state->enableFIFOFull = 1;
+
+    return 0;
+}
+
+struct UsfTags {
+    double length = -1; // Track duration. -1 represents looping infinitely
+    double fade = 0;   // Fade out duration
+    bool setTags = false;  // If set, set the tags with the below variables
+    const TagHandler * tagHandler;
+    void * handler_ctx;
+};
+
+static double
+get_length_from_string(const char * string)
+{
+    size_t len = strlen(string);
+    int32_t total = 0; // Total time in milliseconds
+    int32_t finalMult = 1000; // Multiplier for the final unit. If no delimiter is used, use seconds.
+    int32_t localMult = 1; // Multiplier in the currently parsed unit (millisecond, second, minute).
+    int32_t tmp = 0; // 
+    for (int i = len - 1; i >= 0; i--) {
+        char c = string[i];
+        if (c >= '0' && c <= '9') {
+            tmp += (c - '0') * localMult;
+            localMult *= 10;
+        } else {
+            localMult = 1;
+            int mult = 0;
+            if (c == '.') {
+                mult = 1;
+            } else if (c == ':') {
+                mult = 1000;
+                finalMult = 60000;
+            } else {
+                return -1; // Error in parsing duration. Return -1 (looping).
+            }
+            total += tmp * mult;
+            tmp = 0;
+        }
+    }
+    total += finalMult*tmp;
+
+    return total/1000;
+}
+
+static bool
+usf_set_tag(const char * field, const char * name, const char *value, TagType type, UsfTags * tags)
+{
+    if (strcmp(name, field) == 0) {
+        tag_handler_invoke_tag(*tags->tagHandler, tags->handler_ctx, type, value);
+        return true;
+    }
+    return false;
+}
+
+static int
+usf_tags_target(void * context, const char * name, const char * value)
+{
+    struct UsfTags * tags = ( struct UsfTags * ) context;
+    if (!value) return 0;
+
+    if (tags->setTags) {
+        usf_set_tag("title", name, value, TAG_TITLE, tags) ||
+            usf_set_tag("artist", name, value, TAG_ARTIST, tags) ||
+            usf_set_tag("composer", name, value, TAG_COMPOSER, tags) ||
+            usf_set_tag("game", name, value, TAG_ALBUM, tags) ||
+            usf_set_tag("year", name, value, TAG_DATE, tags) ||
+            usf_set_tag("genre", name, value, TAG_GENRE, tags) ||
+            usf_set_tag("track", name, value, TAG_TRACK, tags);
+    }
+
+    if (strcmp("length", name) == 0) {
+        tags->length = get_length_from_string(value);
+    } else if (strcmp("fade", name) == 0) {
+        tags->fade = get_length_from_string(value);
+    }
+    return 0;
+}
+
+
+inline static double
+track_lengths(double length, double fade)
+{
+    return length + fade;
+
+}
+
+static void
+usf_file_decode(Decoder &decoder, Path path_fs)
+{
+    /* Load the file */
+
+    UsfTags tags;
+    UsfLoaderState state;
+    state.emu = malloc(usf_get_state_size());
+    usf_clear(state.emu);
+    AtScopeExit(state) {
+        free(state.emu);
+    };
+
+    const char* path = path_fs.c_str();
+    const int psf_version_tags = psf_load(path, &stdio_callbacks, 0, 0, 0, usf_tags_target, &tags, 0);
+    const int psf_version_state = psf_load( path, &stdio_callbacks, psf_version_tags, usf_loader, &state, usf_info, &state, 0 );
+
+    if ( psf_version_tags != 0x21 || psf_version_state <= 0 ) {
+        LogWarning(usf_domain, "Error loading usf file");
+        return;
+    }
+
+    usf_set_compare( state.emu, state.enableCompare );
+    usf_set_fifo_full( state.emu, state.enableFIFOFull );
+
+    int32_t sample_rate;
+    usf_render( state.emu, 0, 0, &sample_rate );
+
+    /* initialize the MPD decoder */
+
+    const AudioFormat audio_format(sample_rate, SampleFormat::S16, USF_CHANNELS);
+    assert(audio_format.IsValid());
+
+    double totalLength =  track_lengths(tags.length, tags.fade);
+
+    // Duration
+    decoder_initialized(decoder, audio_format, true, SongTime::FromS(totalLength));
+
+    /* .. and play */
+    DecoderCommand cmd;
+    do {
+        int16_t buf[USF_BUFFER_SAMPLES];
+        const char* result = usf_render(state.emu, buf, USF_BUFFER_FRAMES, 0);
+        if (result != 0) {
+            LogWarning(usf_domain, result);
+            break;
+        }
+
+        // Simple fading
+        double fadeTime = tags.fade;
+        double trackLength = tags.length;
+        if (fadeTime > 0 && totalLength >= 0 && decoder.timestamp > trackLength) {
+            const double vol = 1.0 - ((decoder.timestamp - trackLength) / fadeTime);
+            const double normalizedVol = vol < 0 ? 0 : vol;
+            for (unsigned int i = 0; i < USF_BUFFER_SAMPLES; i++) {
+                buf[i] *= normalizedVol;
+            }
+        }
+
+        cmd = decoder_data(decoder, nullptr, buf, sizeof(buf), 0);
+
+        // Stop song manually
+        if (totalLength >= 0 && decoder.timestamp > totalLength+2)
+            break;
+
+        if (cmd == DecoderCommand::SEEK) {
+            // If user seeks during the fade period. Disable fading and play forever.
+            // Hacky way to give user posibility to enable looping on the fly
+            if (decoder.timestamp > trackLength) {
+                totalLength = -1;
+            }
+            // Seek manually by restarting emulator and discarding samples.
+            const double target_time = decoder_seek_time(decoder).ToDoubleS();
+            usf_restart(state.emu);
+            const unsigned frames_to_throw = (unsigned) (sample_rate*target_time+0.5);
+            usf_render(state.emu, 0, frames_to_throw, 0);
+
+            // Time correction after seek. Decided by trial and error.
+            decoder_command_finished(decoder);
+            double newTime = (frames_to_throw/(double)sample_rate)-0.5;
+            decoder_timestamp(decoder, newTime);
+        }
+
+    } while (cmd != DecoderCommand::STOP);
+    usf_shutdown(state.emu);
+}
+
+static bool
+usf_scan_file(Path path_fs, const struct TagHandler &handler, void *handler_ctx) 
+{
+    const char* path = path_fs.c_str();
+    UsfTags tags;
+    tags.setTags = true;
+    tags.tagHandler = &handler;
+    tags.handler_ctx = handler_ctx;
+    const int psf_version = psf_load(path, &stdio_callbacks, 0, 0, 0, usf_tags_target, &tags, 0);
+    if (psf_version < 0) {
+        return false;
+    }
+
+    // Duration
+    double totalLength = track_lengths(tags.length, tags.fade);
+
+    tag_handler_invoke_duration(handler, handler_ctx, SongTime::FromS(totalLength));
+    return true;
+}
+
+static const char *const usf_suffixes[] = {
+	"usf",
+	"miniusf",
+	nullptr
+};
+
+extern const struct DecoderPlugin usf_decoder_plugin;
+const struct DecoderPlugin usf_decoder_plugin = {
+	"usf",
+	nullptr,
+	nullptr,
+	nullptr, /* stream_decode() */
+	usf_file_decode,
+	usf_scan_file,
+	nullptr, /* stream_tag() */
+	nullptr,
+	usf_suffixes,
+	nullptr, /* mime_types */
+};
--- a/Makefile.am
+++ b/Makefile.am
@@ -796,6 +796,7 @@ libarchive_a_CPPFLAGS = $(AM_CPPFLAGS) \
 
 ARCHIVE_LIBS = \
 	libarchive.a \
+	$(PSFLIB_LIBS) \
 	$(BZ2_LIBS) \
 	$(ISO9660_LIBS) \
 	$(ZZIP_LIBS)
@@ -910,6 +911,7 @@ libdecoder_a_CPPFLAGS = $(AM_CPPFLAGS) \
 	$(AUDIOFILE_CFLAGS) \
 	$(LIBMIKMOD_CFLAGS) \
 	$(GME_CFLAGS) \
+	$(LAZYUSF_CFLAGS) \
 	$(SIDPLAY_CFLAGS) \
 	$(FLUIDSYNTH_CFLAGS) \
 	$(WILDMIDI_CFLAGS) \
@@ -929,6 +931,7 @@ DECODER_LIBS = \
 	$(SNDFILE_LIBS) \
 	$(AUDIOFILE_LIBS) $(LIBMIKMOD_LIBS) \
 	$(GME_LIBS) \
+	$(LAZYUSF_LIBS) \
 	$(SIDPLAY_LIBS) \
 	$(FLUIDSYNTH_LIBS) \
 	$(WILDMIDI_LIBS) \
@@ -1037,6 +1040,12 @@ if ENABLE_MIKMOD_DECODER
 libdecoder_a_SOURCES += \
 	src/decoder/plugins/MikmodDecoderPlugin.cxx \
 	src/decoder/plugins/MikmodDecoderPlugin.hxx
+endif
+
+if ENABLE_LAZYUSF
+libdecoder_a_SOURCES += \
+	src/decoder/plugins/UsfDecoderPlugin.cxx \
+	src/decoder/plugins/UsfDecoderPlugin.hxx
 endif
 
 if ENABLE_MODPLUG
--- a/configure.ac
+++ b/configure.ac
@@ -425,6 +425,11 @@ AC_ARG_ENABLE(un,
 		[disable support for clients connecting via unix domain sockets (default: enable)]),,
 	[enable_un=$host_is_unix])
 
+AC_ARG_ENABLE(lazyusf,
+	AS_HELP_STRING([--enable-lazyusf],
+		[enable usf support via lazyusf]),,
+	enable_lazyusf=auto)
+
 AC_ARG_ENABLE(vorbis,
 	AS_HELP_STRING([--enable-vorbis],
 		[enable Ogg Vorbis decoder]),,
@@ -889,6 +894,13 @@ if test x$enable_mikmod = xyes; then
 fi
 
 AM_CONDITIONAL(ENABLE_MIKMOD_DECODER, test x$enable_mikmod = xyes)
+
+dnl -------------------------------- usf --------------------------------------
+MPD_ENABLE_AUTO_LIB(lazyusf, LAZYUSF, lazyusf, usf_render, [-llazyusf -lpsflib], [],
+        [usf decoder plugin], [liblazyusf or psflib not found])
+
+#MPD_ENABLE_AUTO_LIB(psflib, PSFLIB, psflib, psf_load, [-lpsflib], [],
+                    #[psf file plugin], [psflib not found])
 
 dnl -------------------------------- libmodplug -------------------------------
 MPD_ENABLE_AUTO_PKG(modplug, MODPLUG, [libmodplug],
@@ -1391,6 +1403,7 @@ printf '\n\t'
 results(sndfile, [libsndfile])
 results(mikmod, [MikMod])
 results(modplug, [MODPLUG])
+results(lazyusf, [lazyusf])
 results(mad, [MAD])
 results(mpg123, [MPG123])
 results(mpc, [Musepack])
--- a/src/decoder/DecoderList.cxx
+++ b/src/decoder/DecoderList.cxx
@@ -43,6 +43,7 @@
 #include "plugins/MpcdecDecoderPlugin.hxx"
 #include "plugins/FluidsynthDecoderPlugin.hxx"
 #include "plugins/SidplayDecoderPlugin.hxx"
+#include "plugins/UsfDecoderPlugin.hxx"
 #include "util/Macros.hxx"
 
 #include <string.h>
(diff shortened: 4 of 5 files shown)


MaxKellermann:
Is there a special reason you're submitting so many small commits? This looks like you have been working on your code incrementally, but this should really be one big commit.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+
+#include <usf.h>
+#include <psflib.h>
+
+static constexpr Domain usf_domain("usf");
+static constexpr unsigned USF_BUFFER_FRAMES = 2048;
+static constexpr unsigned USF_CHANNELS = 2;
+static constexpr unsigned USF_BUFFER_SAMPLES = USF_BUFFER_FRAMES*USF_CHANNELS;
+
+static void * stdio_fopen( const char * path )
```
MaxKellermann: Please follow the coding style like the rest of MPD; e.g.

    static void *
    stdio_fopen(const char *path)

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+{
+    return fclose( (FILE*) f );
+}
+
+static long stdio_ftell( void * f )
+{
+    return ftell( (FILE*) f );
+}
+
+psf_file_callbacks stdio_callbacks =
```
MaxKellermann: This should be `static` + `constexpr`

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+    stdio_fopen,
+    stdio_fread,
+    stdio_fseek,
+    stdio_fclose,
+    stdio_ftell
+};
+
+struct usf_loader_state
+{
+    uint32_t enable_compare;
```
MaxKellermann: Why `uint32_t`? Should this be `bool`?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+    uint32_t enable_fifo_full;
```
MaxKellermann: Likewise. `bool`?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+{
+    "\\/:",
+    stdio_fopen,
+    stdio_fread,
+    stdio_fseek,
+    stdio_fclose,
+    stdio_ftell
+};
+
+struct usf_loader_state
```
MaxKellermann: Use CamelCase and document the variables.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+    return 0;
+}
+
+static int
+usf_tags_target(void * context, const char * name, const char * value)
+{
+    struct usf_tags * tags = ( struct usf_tags * ) context;
+    if (!value) return 0;
+
+    if (usf_save_tag("title", name, value, tags->title));
```
MaxKellermann: Empty block (only `;`) after `if` is bad style, and is an invitation for hard-to-find bugs.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+    std::string genre;
+    std::string length;
+    std::string fade;
+    std::string track;
+};
+
+static int
+usf_save_tag(const char * field, const char * name, const char *value, std::string& tag)
+{
+    if (strcasecmp(name, field) == 0) {
```
MaxKellermann: Does this really need to be case-insensitive?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+    else if (usf_save_tag("usfby", name, value, tags->usfby));
+    else if (usf_save_tag("genre", name, value, tags->genre));
+    else if (usf_save_tag("length", name, value, tags->length));
+    else if (usf_save_tag("fade", name, value, tags->fade));
+    else if (usf_save_tag("track", name, value, tags->track)){}
+    return 0;
+}
+
+static double
+get_length_from_string(std::string length)
```
MaxKellermann: Passing a std::string by value is a lot of overhead.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+    if ( strcasecmp( name, "_enablecompare" ) == 0 && strlen( value ) )
+        state->enable_compare = 1;
+    else if ( strcasecmp( name, "_enablefifofull" ) == 0 && strlen( value ) )
+        state->enable_fifo_full = 1;
+
+    return 0;
+}
+
+struct usf_tags {
+    std::string title;
```
MaxKellermann: Why `std::string`? Why not pass everything right into `TagHandler`?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+    const char * str_length = length.c_str();
+    double ttime = 0, temp = 0, mult = 1, level = 1;
+    char Source[1024];
+    char * src = Source + strlen(str_length);
+    strcpy(&Source[1], str_length);
+    Source[0] = 0;
+
+    while(*src)
+    {
+        if((*src >= '0') && (*src <= '9'))
```
MaxKellermann: Useless parantheses.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+}
+
+static double
+get_length_from_string(std::string length)
+{
+    const char * str_length = length.c_str();
+    double ttime = 0, temp = 0, mult = 1, level = 1;
+    char Source[1024];
+    char * src = Source + strlen(str_length);
+    strcpy(&Source[1], str_length);
```
MaxKellermann: Never never never ever use `strcpy()` unless you really really know what you're doing! This is an exploitable bug you just wrote.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+        }
+        src--;
+    }
+
+    ttime += (temp * (1000 * level));
+    return ttime/1000;
+}
+
+static void
+track_lengths(std::string length, std::string fade, double& track_length, double& fade_time, double& total_length)
```
MaxKellermann: More `std::string` bloat.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+    double fade_time;
+    double total_length;
+    track_lengths(tags.length, tags.fade, track_length, fade_time, total_length);
+
+    // Duration
+    decoder_initialized(decoder, audio_format, true, SongTime::FromS(total_length));
+
+    /* .. and play */
+    DecoderCommand cmd;
+    signed short buf[USF_BUFFER_SAMPLES];
```
MaxKellermann: Move this buffer declaration to where it's really used.

MaxKellermann: And don't use `short` unless you do NOT care about how large this really is. Did you mean `int16_t`?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+    state.emu = malloc(usf_get_state_size());
+    usf_clear(state.emu);
+
+    const char* path = path_fs.c_str();
+    const int psf_version_tags = psf_load(path, &stdio_callbacks, 0, 0, 0, usf_tags_target, &tags, 0);
+    const int psf_version_state = psf_load( path, &stdio_callbacks, psf_version_tags, usf_loader, &state, usf_info, &state, 0 );
+
+    if ( psf_version_tags != 0x21 || psf_version_state <= 0 ) {
+        LogWarning(usf_domain, "Error loading usf file");
+        free(state.emu);
```
MaxKellermann: Use `AtScopeExit()` to make freeing the buffer reliable, if you really need `malloc()`. Adding `free()` to every `return` is error prone.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,342 @@
...
+
+    } while (cmd != DecoderCommand::STOP);
+    usf_shutdown(state.emu);
+    free(state.emu);
+}
+
+static void
+set_tag_from_tag(const struct TagHandler &handler, void *handler_ctx, TagType type, std::string value)
+{
+    if (!value.empty()) tag_handler_invoke_tag(handler, handler_ctx, type, value.c_str());
```
MaxKellermann: Why this empty check?

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: I'm not completely sure about the variables "enable_compare" and "enable_fifo_full" are doing, as they work on a lower level (in the emulator). I'm not completely sure what their purpose are, but they are needed to decode properly. What would be a reasonable comment for this?

MaxKellermann: What, you don't know what they're doing? You wrote that code! How can you not know how your own code works!

habbbe: The variables are passed into the third party library's usf_set_compare and usf_set_fifo_full functions. I didn't write the libraries responsible for the actual decoding. In this plugin they are merely read from the tags of the input file and passed through to the emulator.

MaxKellermann: So you're saying you are using undocumented features in the library your code uses?

habbbe: The documentation for these functions are:
/* These are both required functions before calling usf_render.
   Their values are assumed to be zero, unless the respective
   _enablecompare or _enablefifofull tags are present in the file. */

The documentation for the actual file format (and for theses flags) can be found at:
https://www.hcs64.com/usf/usf.txt
The two flags can be found on the bottom of this text.

I would argue that including this low level information would take away focus from how the (decoder plugin's) actual code works.
Would it be enough to give the comment "Flags needed for by lazyusf for correct rendering"?

MaxKellermann: I'd say "The _enablecompare tag is present in the file; passed to usf_set_compare()." And it should be bool. That lazyusf API is badly designed!

habbbe: I agree that the documentation leaves things to be desired.
Unfortunately it's the only available library for this file format :/

I will perform the other changes you suggest.

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: Would it be preferable to set the tag to the empty string in this case?
All tags are not always present in the input files, which is why they can be empty.

I take it I should also pass the value string by reference?

MaxKellermann: No, not use `std::string` at all!

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: For the parent comment:
I worked on this over a period of time, resulting in many commits. Would you rather I squash it to one and submit a new pull request?

MaxKellermann:
Please fold all commits. I can't review these micro-commits, because by itself, they don't make sense - these commits are not self-contained, they just fix mistakes in previous unmerged commits.

habbbe:
Squashed commits

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+#include <usf.h>
+#include <psflib.h>
+
+static constexpr Domain usf_domain("usf");
+static constexpr unsigned USF_BUFFER_FRAMES = 2048;
+static constexpr unsigned USF_CHANNELS = 2;
+static constexpr unsigned USF_BUFFER_SAMPLES = USF_BUFFER_FRAMES*USF_CHANNELS;
+
+static void *
+stdio_fopen( const char * path )
```
MaxKellermann: Coding style. These spaces before and after parantheses and after the * are everywhere!

MaxKellermann:
Fix your indentation. MPD uses tabs, not 4 spaces.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+    void * emu; // The emulator state
+};
+
+static int
+usf_loader(void * context, const uint8_t * exe, size_t exe_size,
+               const uint8_t * reserved, size_t reserved_size)
+{
+    struct UsfLoaderState * state = ( struct UsfLoaderState * ) context;
+
+    if ( exe && exe_size > 0 ) return -1;
```
MaxKellermann: What does this check do? (Code comment?)

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+
+    return usf_upload_section( state->emu, reserved, reserved_size );
+}
+
+static int
+usf_info(void * context, const char * name, const char * value)
+{
+    struct UsfLoaderState * state = ( struct UsfLoaderState * ) context;
+    if (!value) return 0;
```
MaxKellermann: Why check value for begin non-nullptr when you don't use the value?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+        state->enableFIFOFull = 1;
+
+    return 0;
+}
+
+struct UsfTags {
+    double length = -1; // Track duration. -1 represents looping infinitely
+    double fade = 0;   // Fade out duration
+    bool setTags = false;  // If set, set the tags with the below variables
+    const TagHandler * tagHandler;
```
MaxKellermann: Make this a reference, not a pointer.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+    void * handler_ctx;
+};
+
+static double
+get_length_from_string(const char * string)
+{
+    size_t len = strlen(string);
+    int32_t total = 0; // Total time in milliseconds
+    int32_t finalMult = 1000; // Multiplier for the final unit. If no delimiter is used, use seconds.
```
MaxKellermann: coding style: no CamelCase in variable names.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+    bool setTags = false;  // If set, set the tags with the below variables
+    const TagHandler * tagHandler;
+    void * handler_ctx;
+};
+
+static double
+get_length_from_string(const char * string)
+{
+    size_t len = strlen(string);
+    int32_t total = 0; // Total time in milliseconds
```
MaxKellermann: Why force the compiler to use 32 bit integers?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+    int32_t finalMult = 1000; // Multiplier for the final unit. If no delimiter is used, use seconds.
+    int32_t localMult = 1; // Multiplier in the currently parsed unit (millisecond, second, minute).
+    int32_t tmp = 0; //
```
MaxKellermann: What does `tmp` mean? I mean, this is not a temporary value/variable - it is the one variable that is really important in this function, its lifetime spans the whole function body.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+struct UsfTags {
+    double length = -1; // Track duration. -1 represents looping infinitely
+    double fade = 0;   // Fade out duration
+    bool setTags = false;  // If set, set the tags with the below variables
+    const TagHandler * tagHandler;
+    void * handler_ctx;
+};
+
+static double
+get_length_from_string(const char * string)
```
MaxKellermann: API documentation!

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+            tmp = 0;
+        }
+    }
+    total += finalMult*tmp;
+
+    return total/1000;
+}
+
+static bool
+usf_set_tag(const char * field, const char * name, const char *value, TagType type, UsfTags * tags)
```
MaxKellermann: API documentation!

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+{
+    if (strcmp(name, field) == 0) {
+        tag_handler_invoke_tag(*tags->tagHandler, tags->handler_ctx, type, value);
+        return true;
+    }
+    return false;
+}
+
+static int
+usf_tags_target(void * context, const char * name, const char * value)
```
MaxKellermann: API documentation!

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+{
+    struct UsfTags * tags = ( struct UsfTags * ) context;
+    if (!value) return 0;
+
+    if (tags->setTags) {
+        usf_set_tag("title", name, value, TAG_TITLE, tags) ||
```
MaxKellermann: Make this a table, instead of repeating many `usf_set_tag()` calls.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+        tags->fade = get_length_from_string(value);
+    }
+    return 0;
+}
+
+
+inline static double
+track_lengths(double length, double fade)
+{
+    return length + fade;
```
MaxKellermann: What does that formula mean? (If that's not obvious, a code comment is missing.)

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+static void
+usf_file_decode(Decoder &decoder, Path path_fs)
+{
+    /* Load the file */
+
+    UsfTags tags;
+    UsfLoaderState state;
+    state.emu = malloc(usf_get_state_size());
+    usf_clear(state.emu);
+    AtScopeExit(state) {
```
MaxKellermann: Capturing `state` means a full copy of the whole struct is copied for the scope. Better capture only a reference to it!

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+        free(state.emu);
+    };
+
+    const char* path = path_fs.c_str();
+    const int psf_version_tags = psf_load(path, &stdio_callbacks, 0, 0, 0, usf_tags_target, &tags, 0);
+    const int psf_version_state = psf_load( path, &stdio_callbacks, psf_version_tags, usf_loader, &state, usf_info, &state, 0 );
+
+    if ( psf_version_tags != 0x21 || psf_version_state <= 0 ) {
```
MaxKellermann: Obscure "magic" integer values! What do they mean?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+        LogWarning(usf_domain, "Error loading usf file");
+        return;
+    }
+
+    usf_set_compare( state.emu, state.enableCompare );
+    usf_set_fifo_full( state.emu, state.enableFIFOFull );
+
+    int32_t sample_rate;
```
MaxKellermann: Why 32 bit?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "config.h"
+#include "UsfDecoderPlugin.hxx"
+#include "../DecoderAPI.hxx"
+#include "../DecoderInternal.hxx"
```
MaxKellermann: Don't include `DecoderInternal.hxx`!

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+    // Duration
+    decoder_initialized(decoder, audio_format, true, SongTime::FromS(totalLength));
+
+    /* .. and play */
+    DecoderCommand cmd;
+    do {
+        int16_t buf[USF_BUFFER_SAMPLES];
+        const char* result = usf_render(state.emu, buf, USF_BUFFER_FRAMES, 0);
+        if (result != 0) {
+            LogWarning(usf_domain, result);
```
MaxKellermann: Better prefix this log message, or the user may be confused by some generic message, not knowing where it came from.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+
+    int32_t sample_rate;
+    usf_render( state.emu, 0, 0, &sample_rate );
+
+    /* initialize the MPD decoder */
+
+    const AudioFormat audio_format(sample_rate, SampleFormat::S16, USF_CHANNELS);
+    assert(audio_format.IsValid());
+
+    double totalLength =  track_lengths(tags.length, tags.fade);
```
MaxKellermann: Use `SongTime` instead of floating point.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+    DecoderCommand cmd;
+    do {
+        int16_t buf[USF_BUFFER_SAMPLES];
+        const char* result = usf_render(state.emu, buf, USF_BUFFER_FRAMES, 0);
+        if (result != 0) {
+            LogWarning(usf_domain, result);
+            break;
+        }
+
+        // Simple fading
```
MaxKellermann: Simple fading? What does that mean?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+        double fadeTime = tags.fade;
+        double trackLength = tags.length;
+        if (fadeTime > 0 && totalLength >= 0 && decoder.timestamp > trackLength) {
+            const double vol = 1.0 - ((decoder.timestamp - trackLength) / fadeTime);
+            const double normalizedVol = vol < 0 ? 0 : vol;
+            for (unsigned int i = 0; i < USF_BUFFER_SAMPLES; i++) {
+                buf[i] *= normalizedVol;
```
MaxKellermann: Lots and lots of floating point, when MPD may run on machines which have no FPU! Why that?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+            }
+        }
+
+        cmd = decoder_data(decoder, nullptr, buf, sizeof(buf), 0);
+
+        // Stop song manually
```
MaxKellermann: Huh, what??

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+        if (totalLength >= 0 && decoder.timestamp > totalLength+2)
+            break;
+
+        if (cmd == DecoderCommand::SEEK) {
+            // If user seeks during the fade period. Disable fading and play forever.
+            // Hacky way to give user posibility to enable looping on the fly
```
MaxKellermann: WTF??

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+            if (decoder.timestamp > trackLength) {
+                totalLength = -1;
+            }
+            // Seek manually by restarting emulator and discarding samples.
+            const double target_time = decoder_seek_time(decoder).ToDoubleS();
+            usf_restart(state.emu);
+            const unsigned frames_to_throw = (unsigned) (sample_rate*target_time+0.5);
+            usf_render(state.emu, 0, frames_to_throw, 0);
```
MaxKellermann: Does this mean that by seeking to the end of a large file, the decoder will decode the whole file completely, only to go to that position? I'd say this decoder is not seekable.

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+
+            // Time correction after seek. Decided by trial and error.
+            decoder_command_finished(decoder);
+            double newTime = (frames_to_throw/(double)sample_rate)-0.5;
```
MaxKellermann: What is this magic `-0.5`?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+        if (cmd == DecoderCommand::SEEK) {
+            // If user seeks during the fade period. Disable fading and play forever.
+            // Hacky way to give user posibility to enable looping on the fly
+            if (decoder.timestamp > trackLength) {
+                totalLength = -1;
+            }
+            // Seek manually by restarting emulator and discarding samples.
+            const double target_time = decoder_seek_time(decoder).ToDoubleS();
+            usf_restart(state.emu);
+            const unsigned frames_to_throw = (unsigned) (sample_rate*target_time+0.5);
```
MaxKellermann: What is this magic `+0.5`?

src/decoder/plugins/UsfDecoderPlugin.cxx
```diff
@@ -0,0 +1,318 @@
...
+	"usf",
+	nullptr,
+	nullptr,
+	nullptr, /* stream_decode() */
+	usf_file_decode,
+	usf_scan_file,
+	nullptr, /* stream_tag() */
+	nullptr,
+	usf_suffixes,
+	nullptr, /* mime_types */
```
MaxKellermann: Are there no MIME types for this file format?

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: The usf format supports looping (a lot of overworld music etc.). But the files also has a set track length in the tags, probably calculated from the time passed after a certain number of loops. For this kind of music, there is (at least for me) a desire to loop indefinitely, which this is gives an opportunity to do.

This particular way to achieve looping was inspired by a decoder plugin for foobar2000.

I would be glad to just keep this in a separate fork if you consider it out of place.

MaxKellermann: It is a very obscure way to enable a feature. Too obscure.

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: That is correct. Would you rather I remove seeking altogether?

MaxKellermann: This kind of "clumsy" seeking could as well be implemented in the MPD core for all unseekable plugins; however there may be one advantage on your approach: by passing 0, the library may skip the actual decoder, which may save lots of CPU (and memory). So there is a good use for your code.

Oh, please don't use `0` when you really mean `nullptr`.

MaxKellermann: I think one thing should be added: when seeking forward, you don't need to restart the decoder. (You can implement that in a separate commit, as an incremental optimization, after I have merged this one.)

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: Should this rather be a flag that is passed to the server, or do you reckon the functionality should be removed altogether? Maybe this functionality could also be added later?

MaxKellermann: Add later, or add a configuration option. (And document your plugin in doc/user.xml).

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: Are you suggesting I use std::map for this? Or would you rather have an if-else series mapping a string to a TagType?

MaxKellermann: No! Something like https://github.com/MaxKellermann/MPD/blob/master/src/decoder/plugins/FfmpegMetaData.cxx

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: Is there a way to get the current timestamp (in ms)? Or do I have to keep track myself by counting samples?

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: This is for the linear fading when a song reaches the end. The multiplication in
buf[i] *= normalizedVol;
 is where I multiply the current volume with the ratio (between 0.0 and 1.0 where 1.0 is full volume).

When I tried a method without floating point operations I got rather poor performance (about 40% worse than with floating point) after a quick benchmark. Is there a define I could use to check for the presence of an FPU?

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: The library (lazyusf) doesn't stop decoding/emulating automatically, which is why I need to stop the song through a timestamp check (this will be rewritten to use SongTime).

src/decoder/plugins/UsfDecoderPlugin.cxx
MaxKellermann: This is duplicate code, and this doesn't belong in a decoder plugin at all! The decoder is not supposed to apply software volume.

src/decoder/plugins/UsfDecoderPlugin.cxx
MaxKellermann: And what if totalLength<0? What is this `2`? Your code is full of arbitrary constant numbers, with no explanation!

src/decoder/plugins/UsfDecoderPlugin.cxx
MaxKellermann: You did it the wrong way: you're the decoder, you should know the timestamp, but you didn't tell the MPD core via `decoder_timestamp()`. In spite of that, you ask the MPD core for the timestamp, which it can only guess by counting samples.

It is up to you (as the decoder) to know the time stamp. There can be various ways to keep track of that; some codecs contain a time stamp in every packet/block, and others need to count samples.

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: Then the song will loop indefinitely. There are some files that don't specify the length tag where this is the desired behavior. I will add a comment for this.

The number 2 will be removed.

src/decoder/plugins/UsfDecoderPlugin.cxx
habbbe: A fading function is necessary (and expected) in a format like this. It is also implemented in the GmeDecoderPlugin, although here it is implemented through a call to the third party library, that applies the fading before returning the samples to GmeDecoderPlugin. This would still be in the decoder, no?
Such a function does not exist in lazyusf, which is why I needed to implement it here.

MaxKellermann: In the GME decoder, this call is required as per libgme API (see commit8b2f4fc823496c9e473616487e6dc572d77df346), and it does not duplicate any code.

habbbe: I'm talking about what you said that you are not supposed to apply software volume in the decoder, when that it is clearly happening in libgme. The call is not required either (at least not in the latest revision); it merely makes it simpler to check when the track has ended (in addition to applying the fading), but correct me if I'm wrong.
I hear you on he code duplication. When this is taken care of, I would like to refer you to my second comment in this chain.

MaxKellermann: If it's not necessary, then it should be removed from the gme plugin. I didn't write this plugin - I relied on the author's comment.

habbbe: I understand. I would, however, beg you to reconsider allowing the fading for the emulated formats, as this functionality is very much expected by the users.

MaxKellermann: No.

habbbe: Then please disregard this pull request.

MaxKellermann:
Closing this pull request upon author's wish. Annoyingly, that guy has wasted lots of my time for code reviews. Gnaaa!

PR #9, opened 2016-07-03
OSX output plugin: get rid of compile time and runtime deprecation warnings

jacobvosmaer:
Hi! The OS X output plugin creates about 15 deprecation warnings during compilation, and it also emits a deprecation warning at runtime. This change gets rid of the warning by using what appears to be the current APIs.

I am not an experienced OS X or C++ developer but I hope you will consider this change for review anyway. Thanks!

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -127,6 +127,8 @@ osx_output_set_device(OSXOutput *oo, Error &error)
 		goto done;
 	}
 
+	FormatDebug(osx_output_domain, "found %u OS X audio output devices", size);
+
 	/* what are the available audio device IDs? */
 	numdevices = size / sizeof(AudioDeviceID);
 	deviceids = new AudioDeviceID[numdevices];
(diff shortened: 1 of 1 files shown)


jacobvosmaer:
Example runtime deprecation message:

```
decoder: audio_format=44100:16:1, seekable=true
2016-07-03 14:21:33.521 mpd[17081:65097] 14:21:33.521 WARNING:  140: This application, or a library it uses, is using the deprecated Carbon Component Manager for hosting Audio Units. Support for this will be removed in a future release. Also, this makes the host incompatible with version 3 audio units. Please transition to the API's in AudioComponent.h.
output: opened plugin=osx name="echo" audio_format=44100:16:1
```

MaxKellermann:
I can't compile or even run this code, so I have to trust the only guy who's apparently interested in keeping MPD running on OS X. In any case, thanks for working on this.

jacobvosmaer:
Gee, thanks @MaxKellermann !

I am new to MPD, I see now that you do more work on the mpd-devel list. I will use `git format-patch` next time.

MaxKellermann:
On 2016/07/05 16:14, Jacob Vosmaer <notifications@github.com> wrote:
> Gee, thanks @MaxKellermann !
> 
> I am new to MPD, I see now that you do more work on the mpd-devel list. I will use `git format-patch` next time.

Actually, pull requests are easier for me.  Doesn't matter if it's
GitHub or any other git server.

pushrax:
Thanks @jacobvosmaer and @Wang-Yue for updating the OS X code. Your efforts are appreciated!

PR #10, opened 2016-07-08
output/osx rewrite render callback

jacobvosmaer:
This change makes the OS X render callback more explicit (and pedanctic). It also makes sure the output buffers get filled with silence in case of an input buffer underrun.

To make life easier for future contributors I have tried to annotate the render method.

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -41,6 +41,7 @@ struct OSXOutput {
 	const char *device_name;
 
 	AudioComponentInstance au;
+	AudioStreamBasicDescription asbd;
 	Mutex mutex;
 	Cond condition;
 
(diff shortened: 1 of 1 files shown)


src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -220,43 +221,116 @@ osx_output_set_device(OSXOutput *oo, Error &error)
 	return ret;
 }
 
+
+/*
+	This function (the 'render callback' osx_render) is called by the
```
jacobvosmaer: @MaxKellermann is this an OK style for multi-line comments? If not, could you point me to a file in the tree that shows the right style for such comments?

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -347,43 +421,43 @@ osx_output_open(AudioOutput *ao, AudioFormat &audio_format,
 	char errormsg[1024];
 	OSXOutput *od = (OSXOutput *)ao;
 
-	AudioStreamBasicDescription stream_description;
-	stream_description.mSampleRate = audio_format.sample_rate;
-	stream_description.mFormatID = kAudioFormatLinearPCM;
-	stream_description.mFormatFlags = kLinearPCMFormatFlagIsSignedInteger;
+	memset(&od->asbd, 0, sizeof(od->asbd));
+	od->asbd.mSampleRate = audio_format.sample_rate;
```
jacobvosmaer: The new code in the render callback needs information about the stream format, so I figured I'd save this AudioStreamBasicDescription in the OutputDevice struct.

jacobvosmaer:
To be honest this PR does not fix any problems; the old code was working. I rewrote this render method to try and solve a different 'problem' but in the end it turned out to be a red herring.

Still, I think this new code makes more sense, and it does handle more cases than the old code (possible de-interleaving requests from the output hardware, input buffer underruns).

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -220,43 +221,116 @@ osx_output_set_device(OSXOutput *oo, Error &error)
...
 
+	// Acquire mutex when accessing input_buffer
 	od->mutex.lock();
 
-	auto src = od->buffer->Read();
-	if (!src.IsEmpty()) {
-		if (src.size > buffer_size)
-			src.size = buffer_size;
-
-		memcpy(buffer->mData, src.data, src.size);
```
jacobvosmaer: This old memcpy assumed that the output buffer was interleaved in the exact same way as the input buffer.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -220,43 +221,116 @@ osx_output_set_device(OSXOutput *oo, Error &error)
...
+		the output buffers we divide the input buffer frames into 'sub frames'
+		that fit into the output buffers.
+	*/
+	size_t sub_frame_offset = 0;
+	for (unsigned int i = 0 ; i < buffer_list->mNumberBuffers; ++i) {
+		output_buffer = &buffer_list->mBuffers[i];
+		output_buffer_frame_size = output_buffer->mNumberChannels * sample_size;
+		for (UInt32 current_frame = 0; current_frame < available_frames; ++current_frame) {
+				dest = (size_t) output_buffer->mData + current_frame * output_buffer_frame_size;
+				memcpy(
```
jacobvosmaer: If we are asked to fill 1 output buffer with 512 frames (these are numbers I observed in practice) we will call memcpy 512 times.

MaxKellermann:
Next time, please describe on your change in the commit message, not on GitHub comments. GitHub comments are volatile and get lost, only git commit messages (and code comments) persist.

jacobvosmaer:
Thanks @MaxKellermann !

PR #14, opened 2016-09-18
A lot of improvements to OS X output plugin

Wang-Yue:
Add sample rate synchronization, which ensures mpd play bit perfectly on OS X. User can enable this by setting sync_sample_rate to "true".

Add device hogging capability. This forbids other process interfere with selected user devices. User may enable this by setting "hog_device" to "true".

Kill the mutex, locks, extra buffer in original code. Now you'll see greater than 3X performance improvement.

Set the right buffer length for the device.

Fix an audio unit initialization problem. 

Code borrowed from the CMUS project. I am the author of cmus coreaudio output plugin. So I have full rights to redistribute the code. The coding style may be slightly different, but we can always polish/refactor it in another PR. 

@jacobvosmaer Please take a look.

@MaxKellermann please merge.

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -30,7 +30,7 @@
 #include <CoreAudio/CoreAudio.h>
 #include <AudioUnit/AudioUnit.h>
 #include <CoreServices/CoreServices.h>
-
+#include <libkern/OSAtomic.h>
 #include<boost/lockfree/spsc_queue.hpp>
 
 struct OSXOutput {
@@ -48,12 +48,7 @@ struct OSXOutput {
 	AudioComponentInstance au;
 	AudioStreamBasicDescription asbd;
 
-	Mutex mutex;
-	Cond condition;
-
 	boost::lockfree::spsc_queue<uint8_t> *ring_buffer;
-	size_t render_buffer_size;
-	uint8_t *render_buffer;
 
 	OSXOutput()
 		:base(osx_output_plugin) {}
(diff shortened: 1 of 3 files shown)


Wang-Yue:
@jacobvosmaer Please take a look

Wang-Yue:
@MaxKellermann  hogging should be considered an advanced feature, and it's better to only enable this feature when user specify it. does configuration file has a flag to specify system dependent features?

Wang-Yue:
FYI, hogging on OS X == exclusive mode on Windows (https://msdn.microsoft.com/en-us/library/windows/desktop/dd370844(v=vs.85).aspx)

Wang-Yue:
@MaxKellermann  please merge

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -101,6 +103,8 @@ osx_output_configure(OSXOutput *oo, const ConfigBlock &block)
 	}
 
 	oo->channel_map = block.GetBlockValue("channel_map");
+	oo->hog_device = block.GetBlockValue("hog_device", false);
+	oo->sync_sample_rate = block.GetBlockValue("sync_sample_rate", false);
```
MaxKellermann: These must be documented in doc/user.xml

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -665,9 +616,9 @@ osx_output_disable(AudioOutput *ao)
 
 	AudioComponentInstanceDispose(oo->au);
 
-        if (oo->hog_device) {
+	if (oo->hog_device) {
```
MaxKellermann: What's this?

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -665,9 +616,9 @@ osx_output_disable(AudioOutput *ao)
...
 		osx_output_hog_device(oo->dev_id, false);
-        }
+	}
```
MaxKellermann: And this?

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: a tab. I previously used spaces for indent.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: see above

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Done.

jacobvosmaer:
Thanks @Wang-Yue ! I am too late to review this but I am going to ask a few questions so I can learn.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -381,79 +558,13 @@ osx_render(void *vdata,
...
-					output_buffer_frame_size
-				);
-				output_buffer->mDataByteSize += output_buffer_frame_size;
-		}
-		sub_frame_offset += output_buffer_frame_size;
-	}
-
-	return noErr;
+	int count = in_number_frames * od->asbd.mBytesPerFrame;
+	buffer_list->mBuffers[0].mDataByteSize =
```
jacobvosmaer: How do you know there is only one output buffer? Is this documented by Apple anywhere?

To be clear, I am not asking because I think this is wrong, but because I wonder how one is supposed to find out about these things.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -381,79 +558,13 @@ osx_render(void *vdata,
...
+		od->ring_buffer->pop((uint8_t *)buffer_list->mBuffers[0].mData,
+				     count);
```
jacobvosmaer: Maybe I missed it but when reading [the spsc_queue docs](http://www.boost.org/doc/libs/1_59_0/doc/html/boost/lockfree/spsc_queue.html) I nowhere saw any promises about how many elements will be popped. How do we know we get `count` bytes, or at least a multiple of `mBytesPerFrame`?

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -244,6 +259,179 @@ osx_output_set_channel_map(OSXOutput *oo, Error &error)
...
+			      "Cannot get the buffer frame size: %d",
+			      err);
+		return err;
+	}
+
+	buffer_frame_size *= desc.mBytesPerFrame;
+
+	// We set the frame size to a power of two integer that
+	// is larger than buffer_frame_size.
+	while (*frame_size < buffer_frame_size + 1) {
```
jacobvosmaer: Would it make sense to call this variable `buffer_size` instead of `frame_size`? When I read `frame_size` I think of the size of a single frame. The buffer (spsc_queue) holds multiple frames.

src/output/plugins/OSXOutputPlugin.cxx
jacobvosmaer: I know about the comments in /System/Library/Frameworks/CoreAudio.framework/Headers/*.h but it is hard to learn from those comments if you don't know what to look for.

Wang-Yue: This is undocumented  (The entire CoreAudio is poorly document) but I was told by an Apple engineer.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Feel free to rename it.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: You don't get this promise. But if it's not multiple mBytesPerFrame it won't cause any trouble. During the next call CoreAudio will get the remaining bytes.

PR #15, opened 2016-09-19
use osx_output_delay to wait until ring buffer is able to write

Wang-Yue:
This further optimize the performance. 

This is now working properly with a conditional variable bug being fixed (8bbfb5cda12e105460c1e2bf22066873ecc29b0f).

### DIFF
--- a/src/output/OutputPlugin.hxx
+++ b/src/output/OutputPlugin.hxx
@@ -93,10 +93,11 @@ struct AudioOutputPlugin {
 	void (*close)(AudioOutput *data);
 
 	/**
-	 * Returns a positive number if the output thread shall delay
-	 * the next call to play() or pause().  This should be
-	 * implemented instead of doing a sleep inside the plugin,
-	 * because this allows MPD to listen to commands meanwhile.
+	 * Returns a positive number if the output thread shall further
+	 * delay the next call to play() or pause(), which will happen
+	 * until this function returns 0.  This should be implemented
+	 * instead of doing a sleep inside the plugin, because this 
+	 * allows MPD to listen to commands meanwhile.
 	 *
 	 * @return the number of milliseconds to wait
 	 */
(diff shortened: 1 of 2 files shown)


MaxKellermann:
Are you sure this works? There is no guarantee attached with the `delay` method. Previously, this was a loop, and the push was only done when there was room available; now the push can happen at any time, without any runtime checks.

Wang-Yue:
as far as I understand the code:

https://github.com/MaxKellermann/MPD/blob/1c07f197debe044a09c9ba6cb9d79377f40d541a/src/output/OutputThread.cxx#L337

It has a while loop until 
- the timeout specify by the _delay is zero
or
- a command has been issued, in which case it returns a false. 

In the first case, the function will keep waiting. This is exactly what we want.

For the second case, it will break here:
https://github.com/MaxKellermann/MPD/blob/1c07f197debe044a09c9ba6cb9d79377f40d541a/src/output/OutputThread.cxx#L479

so the _play will never happen in this case. This ensures _play will never return 0. 

I think we're good?

MaxKellermann:
You looked at the code, and the code does what you say it does; but the API documentation does not give you the guarantee - quite contrary, it says the opposite! That means the code can change at any time, for example due to optimization, without breaking the (documented) contract. At this point, your code will probably crash; not because the code change was bad, but because your code change was bad, because it relied on guarantees that simply did not exist.
(One way out of here could be changing the documentation to give that guarantee.)

Wang-Yue:
Is this doc what you're pointing to? 

https://github.com/MaxKellermann/MPD/blob/f254831aa6251af2aa9b6052953f59439035b939/src/output/OutputPlugin.hxx#L95

MaxKellermann:
Yes. This documentation states that the next `play` call shall be delayed - it does not specify that `delay` will be called again after one delay. Therefore, it is legal to call `play` without checking `delay` again.

Wang-Yue:
The doc is updated. From a design perspective a repeated call of _delay should be expected. Otherwise if it's called only once, there's no guarantee the resource/buffer/file would become writeable after the single delay.

MaxKellermann:
I'm not happy with your commit message. It's wrong.
You're not "updating" documentation - you're changing the established contract! It is a serious API change. The commit message is misleading.

Wang-Yue:
the commit message is updated.

PR #16, opened 2016-10-07
Fix comment in mpdconf.example ("can setting can" -> "setting can")

dennisschagt:
There is a small error in the sample configuration file. This commits fixes it.


MaxKellermann:
Cherry-picked to v0.19.x

PR #17, opened 2016-10-11
add playnext and playnextid - jump in playlist after current track finishes

hawken93:
I was trying to convert my friend from foobar2000 when he noticed that MPD doesn't allow you to automatically jump in the playlist after the current track finishes playing.
The important changes here are in Queue.cxx and Queue.hxx, both of which may seem like an ugly hack if you know better ways to do this. Improvements could be to try to keep track of where the next track is going to be after either positions or the order changes.
Another way to do this could be to rewrite the order map.

### DIFF
--- a/doc/protocol.xml
+++ b/doc/protocol.xml
@@ -892,6 +892,36 @@
             </para>
           </listitem>
         </varlistentry>
+        <varlistentry id="command_playnext">
+          <term>
+            <cmdsynopsis>
+              <command>playnext</command>
+              <arg choice="req"><replaceable>SONGPOS</replaceable></arg>
+            </cmdsynopsis>
+          </term>
+          <listitem>
+            <para>
+              Goes to the song number
+              <varname>SONGPOS</varname>
+              in the playlist after the current song.
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry id="command_playnextid">
+          <term>
+            <cmdsynopsis>
+              <command>playnextid</command>
+              <arg choice="req"><replaceable>SONGID</replaceable></arg>
+            </cmdsynopsis>
+          </term>
+          <listitem>
+            <para>
+              Goes to the song
+              <varname>SONGID</varname>
+              in the playlist after the current song.
+            </para>
+          </listitem>
+        </varlistentry>
         <varlistentry id="command_previous">
           <term>
             <cmdsynopsis>
--- a/src/Partition.hxx
+++ b/src/Partition.hxx
@@ -137,6 +137,14 @@ struct Partition final : QueueListener, PlayerListener, MixerListener {
 		return playlist.PlayId(pc, id);
 	}
 
+	void PlayNextPosition(int position) {
+		return playlist.PlayNextPosition(pc, position);
+	}
+
+	void PlayNextId(int id) {
+		return playlist.PlayNextId(pc, id);
+	}
+
 	void PlayNext() {
 		return playlist.PlayNext(pc);
 	}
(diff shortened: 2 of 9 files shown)


hawken93:
There should probably also be commands to look up if there is such a jump queued so that players could indicate it

hawken93:
Also, this functionality seems to be interrupted half the time.. Maybe the queue is automatically modified / reset or something.

TheCrueltySage:
Since you raised the topic, I guess I should bring it up too.

I actually kinda faced the same problem of wanting to jump around the playlist, which no player seems to be able to do on Linux. And it's, for whatever reason, the way I like to listen to music one giant playlist, and jumping around it chaotically.
I figured nobody but me wanted it, and solved it by adding a custom patch for a new "temporary" song variable (alike priority, only makes sense for the song in a playlist, default if not manually changed, reset to 0 when the song is played) which allowed changing MPD modes on song beginning to be played according to a bitmask number. I set the mode to random, or sometimes some song to "toggle random", set a priority on every song I wanted to jump to, and another "toggle random" or sometimes "toggle random and single" on the song I wanted to end jumping on.
It was easy to implement and patch, fit the current MPD pretty well, was conceptually pretty simple and allowed for quite a great flexibility (not just playing songs in needed order, but also switching to random automatically on certain track, or even stopping after you reach a certain song, with repeat and consume also being possible to manipulate), but it was quite uncomfortable to work with (need to set a value for every song I want to jump to, and also figure out the jumps chain I want to use, which is a bit of a hassle) and just overall felt hacky as hell and not straightforward enough for my actual primary use case.

Don't know if anyone cares or needs this, but could drop the patches I wrote if it's something upstream is actually interested in merging, and I'm not the only guy who needs this. And if it doesn't complicate things too much.
Not right now, unfortunately, since they're on another machine, but it's not like the idea is hard to implement anyway. My hack was mostly a copy-paste of priority implementation code with some variable names changed, and one function of actually doing stuff (which I'm not sure was optimal, but made for easy patching with no API breakage).

hawken93:
Thanks for commenting on this :)
What my code does is of course only one track, and has the disadvantage that it's done totally wrong..
The foobar people seem to have this one figured out using the following interactions:
 - Every marked track gets some kind of a sequential number (next track is 001, track after that is 002, ...)
 - You are able to push tracks to the end of the queue (003, 004, ..)
 - You are able to remove a track from the queue, which causes the tracks under it to decrement by one (001, 002, 003, 004) -> remove 003 -> (001, 002, 003) where 003 was formerly 004

What I realized is that we are basically making a short-term playlist in addition to the big one, with the quirk that after this short term playlists finishes, it continues on the next track in the big playlist..
This overrides any random mode in the big playlist.

If MPD was to implement this functionality, maybe it would be best if MPD tried to provide maximal flexibility, and rather let the clients decide what kind of operations they would provide their users to provide this queue.

If you make a pull request out of your patch set, I think it will be helpful in figuring out the final solution :)

hawken93:
It sounds like your way of doing this is much better than mine anyway. Basically I think you have made a solution so flexible that it gets hard to present every possible operation to the user..

That should be relatively easy to fix :)

MaxKellermann:
I believe this implementation is bad, because it adds a lot of code complexity for a useless feature. It is useless because if you want to change the order in which songs play, you can just move songs around in the queue. Or, if you're in random mode, adjust queue item priorities.

hawken93:
I agree that my implementation is bad but I don't think the concept is bad.
You may prefer to sort your playlist by album and wish to temporarily enqueue things without having to break your playlist..
The fact that other players have the feature and people actually use it, suggests that the feature would not be useless..
Would it be okay to open this as a feature request?

MaxKellermann:
Break your playlist? Managing the order of playback is the whole point of the queue. Even more: that's its sole purpose, it does nothing else. Using the queue for what it's designed is not "breaking".

What you apparently want is turn the queue into a complex multi-layered beast. But that violates MPD's paradigm of maximum simplicity.

"The fact that other players have the feature and people actually use it" suggests that you should use other players.

PR #19, opened 2016-11-04
doc/protocol: Turn the link to the UTF-8 FAQ into a ulink element

mineo:
(and replace `UTF=8` with `UTF-8`).


MaxKellermann:
Cherry-picked to v0.19.x

PR #22, opened 2017-01-28
Fix build on FreeBSD

riggs-:
Two patchsets are required to make mpd build successfully on FreeBSD. These are likely issues on other platforms as well.

-  src/util/VarSize.hxx requires cstdlib to be included to use free() . Probably on Linux this is pulled in via some other header and hence complies.
- During the overhaul of the DB methods, several method signatures no longer match the corresponding types in Interface.hxx. clang36 on FreeBSD-10 refuses to auto-convert the 'static bool' methods to the required 'void' and exits.

### DIFF
--- a/src/db/Count.cxx
+++ b/src/db/Count.cxx
@@ -62,7 +62,7 @@ Print(Response &r, TagType group, const TagCountMap &m)
 	}
 }
 
-static bool
+static void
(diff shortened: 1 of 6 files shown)


MaxKellermann:
Cherry-picked into v0.20.x (stable)

PR #23, opened 2017-02-08
sidplay: require at least libsidplayfp 1.8.0

olafhering:
With sidplayfp ->sidChips is used, which was introduced with version 1.8.

Signed-off-by: Olaf Hering <olaf@aepfle.de>


MaxKellermann:
This is the wrong approach to solve this problem, as I already told you on IRC.

PR #25, opened 2017-02-26
Add missing include for cstdlib, otherwise free() is undefined

riggs-:
On FreeBSD, and probably also on other BSDs, cstdlib is required to have free() defined.


MaxKellermann:
Cherry-picked to v0.20.x

PR #26, opened 2017-03-23
Fix user doc typo

cotko:


MaxKellermann:
Cherry-picked into v0.20.x

PR #27, opened 2017-03-26
Update libs.py

martinarielhartmann:
upgrade zlib to 1.2.11


martinarielhartmann:
The link to zlib 1.2.8 is broken, so I wasn't able to compile for win32 with build.py. After this change in libs.py I was able to compile MPD 0.20.6 in Ubuntu 16.10. Kudos for your awesome project!

MaxKellermann:
Cherry-picked into v0.20.x

PR #28, opened 2017-04-16
Remove some redundant code

tremby:
I noticed this while looking for where to add code for another patch I'm about to submit. It looks like this code is not needed, since it's identical to the code in the next case. Falling through should achieve the same ends.


MaxKellermann:
Cherry-picked to v0.20.x 0a7d612f41bc61a7218a711e48653bc178186421

PR #29, opened 2017-04-16
Add support for the OriginalDate tag

tremby:
See https://picard.musicbrainz.org/docs/mappings/

This tag is useful when the user would like all releases of the same album to be sorted next to each other.


tremby:
Please ignore this for now; I somehow missed that there is an OriginalDate tag, with an equivalent ID3 mapping, which is more useful.

tremby:
Ready for review.

src/tag/Id3Scan.cxx
```diff
@@ -55,6 +55,10 @@
 #define ID3_FRAME_ALBUM_ARTIST "TPE2"
 #endif
 
+#ifndef ID3_FRAME_ORIGINAL_RELEASE_DATE
+#define ID3_FRAME_ORIGINAL_RELEASE_DATE "TDOR"
+#endif
```
tremby: Some ID3 tag identifiers are in definitions up here, others aren't. I wasn't sure which style to mimic. Also, some definitions are indented, others are not; ditto.

MaxKellermann:
Fails to build:

    src/db/plugins/ProxyDatabasePlugin.cxx:161:23: error: 'MPD_TAG_ORIGINAL_DATE' was not declared in this scope

This needs to check the libmpdclient version (at compile time).

tremby:
Could you please give a further hint at how to do this? Should I check for libmpdclient at one version greater than the currently-released one?

MaxKellermann:
No, just greater than the current one.

tremby:
Ah, of course! I'll fix this stuff up after work today. Thanks.

tremby:
As you see in my libmpdclient patch I've moved the constants so the binary API doesn't suffer. I also updated configure.ac in my local working copy to have a 2.12.0 version number, reconfigured it and rebuilt and reinstalled. I've switched back and forth between stock 2.11 and this version and tested building MPD both ways.

When I was adding the preprocessor version checks I noticed that other such checks use `LIBMPDCLIENT_CHECK_VERSION`. This macro checks that the version being compiled against is equal to or greater than the specified version, but you were hinting that I should check for strictly greater than, and then specify the current version (2.11.0). I'm wondering if perhaps you misremembered how this macro works, or perhaps you were encouraging me to add a new macro, or perhaps to explicitly use the `LIBMPDCLIENT_(MAJOR|MINOR|PATCH)_VERSION` constants.

For now I've used the macro, and assumed that 2.12.0 will be the first to support these tags.

Please let me know if you'd prefer it done another way.

src/tag/Type.h
```diff
@@ -46,6 +46,7 @@ enum TagType
 	TAG_NAME,
 	TAG_GENRE,
 	TAG_DATE,
+	TAG_ORIGINAL_DATE,
```
tremby: I wasn't able to get a preprocessor check here without getting an error. If this line of code needs a check, please help me with it.

MaxKellermann: Why would this need a preprocessor check? This is MPD's internal definition, not one imported from libmpdclient (which in turn mirrors MPD's internal definitions to use MPD's features).

MaxKellermann:
> This macro checks that the version being compiled against is equal to or greater than the specified version, but you were hinting that I should check for strictly greater than
You're right; while my suggestion was correct, it is more important to use libmpdclient's version check macro, which is not designed to use not-yet-released features.

src/output/plugins/HaikuOutputPlugin.cxx
```diff
@@ -27,6 +27,8 @@
 #include "system/Error.hxx"
 #include "Log.hxx"
 
+#include <mpd/client.h>
```
MaxKellermann: This adds a dependency from Haiku on libmpdclient. That's not good!

src/output/plugins/HaikuOutputPlugin.cxx
```diff
@@ -395,6 +397,9 @@ HaikuOutput::SendTag(const Tag &tag)
 			break;
 		case TAG_GENRE:
 		case TAG_DATE:
+#if LIBMPDCLIENT_CHECK_VERSION(2,12,0)
```
MaxKellermann: ... and this condition is unnecessary, because this has nothing to do with libmpdclient.

src/tag/Id3Scan.cxx
```diff
@@ -55,6 +57,12 @@
 #define ID3_FRAME_ALBUM_ARTIST "TPE2"
 #endif
 
+#if LIBMPDCLIENT_CHECK_VERSION(2,12,0)
```
MaxKellermann: Same here. This new tag should always be available, even when not building with libmpdclient.

src/tag/Names.c
```diff
@@ -20,6 +20,8 @@
 #include "config.h"
 #include "Type.h"
 
+#include <mpd/client.h>
```
MaxKellermann: ... and so on... you get the picture?

src/tag/Type.h
tremby: I don't know the answer; I wasn't sure where it would be needed and where it would not, and was trying to err on the side of caution. C++ is not my strong suit.

tremby:
Thanks for explaining. I guess I was overcomplicating things! Hopefully I have it looking right this time.

tremby:
Thanks!

PR #49, opened 2017-05-19
support for mime-type `audio-mpegurl` for M3U playlists

diraimondo:
It looks that there are several mime-types for M3U playlists around (see https://en.wikipedia.org/wiki/M3U); this PR should help mpd to manage a wider range of playlists.

As specific case see: https://github.com/diraimondo/gmusicproxy/issues/93


MaxKellermann:
Cherry-picked into v0.20.x: ffa676f577a202f3b78b69a52cf327fa095541be

Issue #82, opened 2017-07-18
MusicBrainz tags not being added to DB


matthewleon:
Here is a sample mp3 with ID3v2.3 tags:
https://s3.amazonaws.com/www.flactunesapp.com/01+L'Amour+Toujours+1.mp3

Inspecting the file with `id3v2` at the console shows that it has the appropriate MusicBrainz tags. For example:
```
TXXX (User defined text information): (MusicBrainz Artist Id): 09c88499-fd98-4f0e-a4c2-7e763140b611
TXXX (User defined text information): (MusicBrainz Artist Id): 09c88499-fd98-4f0e-a4c2-7e763140b611
TXXX (User defined text information): (MusicBrainz Album Artist Id): 09c88499-fd98-4f0e-a4c2-7e763140b611
TXXX (User defined text information): (MusicBrainz Release Group Id): d71f0ee6-5c77-4adf-99a0-488f3e4502be
TXXX (User defined text information): (MusicBrainz Release Track Id): 3581059b-2f5e-385e-b040-0492cbc7d87c
```

Importing the file to the MPD database doesn't add this metadata, however. This can be determined by uncompressing the database file and manually inspecting. The basic id3 tag info is there (Artist, Track, etc.) but the MusicBrainz info is never scanned in.

Further, the response to `currentsong`, for example, will not include any MusicBrainz info.

I've not been able to get _any_ tracks' MusicBrainz info to be recognized by MPD. These are tracks whose metadata has been set by Picard, so this surprises me. However, a scan of past issues shows that this doesn't seem to have come up for anyone else.

MaxKellermann:
MPD's `mad` and `mpg123` plugins recognize your file just well!
Since you did not provide **any** information about how you configured MPD (and I don't even know which version you're using), I must assume that you're using the `ffmpeg` decoder plugin.
Your file uses tag names which are only specified for ID3 (https://picard.musicbrainz.org/docs/mappings/). However, in the `ffmpeg` decoder, MPD doesn't know it's looking at ID3, and the `ffmpeg` plugin supports only the generic MusicBrainz tag names.

matthewleon:
> Since you did not provide any information about how you configured MPD

My apologies. Version info below:

```
Music Player Daemon 0.20.9

Copyright (C) 2003-2007 Warren Dukes <warren.dukes@gmail.com>
Copyright 2008-2017 Max Kellermann <max.kellermann@gmail.com>
[...]

PR #83, opened 2017-07-19
Read MusicBrainz tags when using ffmpeg decoder.

matthewleon:
Addresses #82.

When using the ffmpeg decoder to read song metadata, MusicBrainz tags
were not being read.

Ffmpeg doesn't tell us the metadata container format, so we accept the
ID3v2 as well as the Vorbis/APEv2 keys from this table: https://picard.musicbrainz.org/docs/mappings/

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -936,7 +936,8 @@ libtag_a_SOURCES =\
 	src/tag/Generic.cxx src/tag/Generic.hxx \
 	src/tag/ApeLoader.cxx src/tag/ApeLoader.hxx \
 	src/tag/ApeReplayGain.cxx src/tag/ApeReplayGain.hxx \
-	src/tag/ApeTag.cxx src/tag/ApeTag.hxx
+	src/tag/ApeTag.cxx src/tag/ApeTag.hxx \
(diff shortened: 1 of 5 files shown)


matthewleon:
I've confirmed that this works through manual q/a.

matthewleon:
@MaxKellermann I have amended the commit according to your comments in https://github.com/matthewleon/MPD/commit/e3ac82b39c7d80a987532dd3238a684f2cf81982

Please review again.

MaxKellermann:
By moving the struct to a header, it is duplicated in every source that includes it, which bloats the binary. And it generates a compiler warning if a source includes the header, but does not use the table. It should be made `extern` instead, and implemented only once in a C++ source.
I suggest moving the table to a dedicated source file (e.g. `Id3MusicBrainz.cxx`) with one commit, and the second commit may then use that table in the `ffmpeg` plugin.

matthewleon:
@MaxKellermann I've followed your directions. Thanks for your patience. I'm a pretty inexperienced C++ programmer, so I won't be surprised if this benefits from further tweaks.

MaxKellermann:
Thanks, cherry-picked to the stable branch (v0.20.x).

PR #86, opened 2017-07-23
Albumart

tokyovigilante:
First rough cut. 

The most obvious deficiency is that each time a chunk is requested, the file is opened and closed. The file reader should probably be stored somewhere (with the client?) but I'm happy to take critiques.

### DIFF
--- a/doc/protocol.xml
+++ b/doc/protocol.xml
@@ -1573,6 +1573,42 @@ OK
       <title>The music database</title>
 
       <variablelist>
+        <varlistentry id="album_art">
+          <term>
+            <cmdsynopsis>
+              <command>albumart</command>
+              <arg choice="req"><replaceable>URI</replaceable></arg>
+              <arg choice="req"><replaceable>OFFSET</replaceable></arg>
+            </cmdsynopsis>
+          </term>
+          <listitem>
+            <para>
+              Searches the directory the file <varname>URI</varname>
+              resides in and attempts to return a chunk of an album
+              art image file at offset <varname>OFFSET</varname>.
+            </para>
+            <para>
+              Uses the filename "cover" with any of ".png, .jpg,
+              .tiff, .bmp".
+            </para>
+            <para>
+              Returns the file size and actual number
+              of bytes read at the requested offset, followed
+              by the chunk requested as raw bytes, then a
+              newline and the completion code.
+            </para>
+            <para>
+              Example:
+            </para>
+
+            <programlisting>albumart
+size: 1024768
+binary: 8192
+&lt;8192 bytes&gt;
+OK
+</programlisting>
+          </listitem>
+        </varlistentry>
 
         <varlistentry id="command_count">
           <term>
--- a/src/command/AllCommands.cxx
+++ b/src/command/AllCommands.cxx
@@ -84,6 +84,7 @@ static constexpr struct command commands[] = {
 	{ "add", PERMISSION_ADD, 1, 1, handle_add },
 	{ "addid", PERMISSION_ADD, 1, 2, handle_addid },
 	{ "addtagid", PERMISSION_ADD, 3, 3, handle_addtagid },
+	{ "albumart", PERMISSION_READ, 2, 2, handle_album_art },
 	{ "channels", PERMISSION_READ, 0, 0, handle_channels },
 	{ "clear", PERMISSION_CONTROL, 0, 0, handle_clear },
 	{ "clearerror", PERMISSION_CONTROL, 0, 0, handle_clearerror },
--- a/src/command/FileCommands.cxx
+++ b/src/command/FileCommands.cxx
@@ -36,8 +36,11 @@
 #include "fs/AllocatedPath.hxx"
 #include "fs/FileInfo.hxx"
 #include "fs/DirectoryReader.hxx"
+#include "input/InputStream.hxx"
 #include "LocateUri.hxx"
 #include "TimePrint.hxx"
+#include "thread/Mutex.hxx"
+#include "thread/Cond.hxx"
 
 #include <assert.h>
 #include <inttypes.h> /* for PRIu64 */
@@ -233,3 +236,114 @@ handle_read_comments(Client &client, Request args, Response &r)
 
 	gcc_unreachable();
 }
+
+/**
+ * Searches for the files listed in #artnames in the UTF8 folder
+ * URI #directory. This can be a local path or protocol-based
+ * URI that #InputStream supports. Returns the first successfully
+ * opened file or #nullptr on failure.
+ */
+static InputStreamPtr
+find_stream_art(const char *directory, Mutex &mutex, Cond &cond)
+{
+	static constexpr char const * art_names[] = {
+		"cover.png",
+		"cover.jpg",
+		"cover.tiff",
+		"cover.bmp"
+	};
+
+	for(const auto name: art_names) {
+		std::string art_file = PathTraitsUTF8::Build(directory, name);
+
+		try {
+			return InputStream::OpenReady(art_file.c_str(), mutex, cond);
+		} catch (const std::exception &e) {}
+	}
+	return nullptr;
+}
+
+static CommandResult
+read_stream_art(Response &r, const char *uri, size_t offset)
+{
+	const char *art_directory = PathTraitsUTF8::GetParent(uri).c_str();
+
+	Mutex mutex;
+	Cond cond;
+
+	InputStreamPtr is = find_stream_art(art_directory, mutex, cond);
+
+	if (is == nullptr) {
+		r.Error(ACK_ERROR_NO_EXIST, "No file exists");
+		return CommandResult::ERROR;
+	}
+	if (!is->KnownSize()) {
+		r.Error(ACK_ERROR_NO_EXIST, "Cannot get size for stream");
+		return CommandResult::ERROR;
+	}
+
+	const size_t art_file_size = is->GetSize();
+
+	constexpr size_t CHUNK_SIZE = 8192;
+	uint8_t buffer[CHUNK_SIZE];
+	size_t read_size;
+
+	is->Seek(offset);
+	read_size = is->Read(&buffer, CHUNK_SIZE);
+
+	r.Format("size: %" PRIu64 "\n"
+			 "binary: %u\n",
+			 art_file_size,
+			 read_size
+			 );
+
+	r.Write(buffer, read_size);
+	r.Write("\n");
+
+	return CommandResult::OK;
+}
+
+#ifdef ENABLE_DATABASE
+static CommandResult
+read_db_art(Client &client, Response &r, const char *uri, const uint64_t offset)
+{
+	const Storage *storage = client.GetStorage();
+	if (storage == nullptr) {
+		r.Error(ACK_ERROR_NO_EXIST, "No database");
+		return CommandResult::ERROR;
+	}
+	std::string uri2 = storage->MapUTF8(uri);
+	return read_stream_art(r, uri2.c_str(), offset);
+}
+#endif
+
+CommandResult
+handle_album_art(Client &client, Request args, Response &r)
+{
+	assert(args.size == 2);
+
+	const char *uri = args.front();
+	size_t offset = args.ParseUnsigned(1);
+
+	const auto located_uri = LocateUri(uri, &client
+#ifdef ENABLE_DATABASE
+					   , nullptr
+#endif
+					   );
+
+	switch (located_uri.type) {
+	case LocatedUri::Type::ABSOLUTE:
+	case LocatedUri::Type::PATH:
+		return read_stream_art(r, located_uri.canonical_uri, offset);
+	case LocatedUri::Type::RELATIVE:
+#ifdef ENABLE_DATABASE
+		return read_db_art(client, r, located_uri.canonical_uri, offset);
+#else
+		r.Error(ACK_ERROR_NO_EXIST, "Database disabled");
+		return CommandResult::ERROR;
+#endif
+	}
+	r.Error(ACK_ERROR_NO_EXIST, "No art file exists");
+	return CommandResult::ERROR;
+}
+
--- a/src/command/FileCommands.hxx
+++ b/src/command/FileCommands.hxx
@@ -33,4 +33,7 @@ handle_listfiles_local(Response &response, Path path_fs);
 CommandResult
 handle_read_comments(Client &client, Request request, Response &response);
 
+CommandResult
+handle_album_art(Client &client, Request request, Response &response);
+
 #endif


tokyovigilante:
Another obvious improvement is to allow a configurable art name search pattern.

coderkun:
I don’t think anyone uses BMP files for such a purpose these days. Instead I would add the file extension variants for JPG and TIF (jpg, jpeg, tif, tiff).

tokyovigilante:
> I don’t think anyone uses BMP files for such a purpose these days. Instead I would add the file extension variants for JPG and TIF (jpg, jpeg, tif, tiff).

Sure, that's really just an example. Mine are all PNGs, I'm going to hopefully make this configurable via the config file.

MaxKellermann:
I havn't looked at the code yet, but if I see "cleanup" patches at the end cleaning up stuff in your own submitted patches, then it's very cumbersome to review the code. Those cleanups should be folded into the original patches instead, or else I have to review known-bad changes.

src/command/FileCommands.cxx
```diff
@@ -234,9 +236,93 @@ handle_read_comments(Client &client, Request args, Response &r)
 	gcc_unreachable();
 }
 
+AllocatedPath
+find_art(AllocatedPath directory)
+{
+    std::vector<std::string> art_names = {
```
MaxKellermann: Big runtime overhead for `std::vector`,  but no gain.

src/command/FileCommands.cxx
```diff
@@ -234,9 +236,93 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+    const Storage *storage = client.GetStorage();
+#endif
+    const auto located_uri = LocateUri(uri, &client
+#ifdef ENABLE_DATABASE
+                                       , storage
+#endif
+                                       );
+    
+    const AllocatedPath path = storage->MapFS(located_uri.canonical_uri);
+    const AllocatedPath directory = path.GetDirectoryName();
```
MaxKellermann: `MapFS()` can return nullptr, and this line can crash if the `music_directory` is not a local path.

src/command/FileCommands.cxx
```diff
@@ -234,9 +236,93 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+    
+    AllocatedPath art_name = find_art(directory);
+    if (art_name.IsNull()) {
+        if (offset > 0) {
+            r.Error(ACK_ERROR_NO_EXIST, "No art found");
+            return CommandResult::ERROR;
+        }
+        r.Write("size: -1\n");
```
MaxKellermann: Huh, what's `-1`?

src/command/FileCommands.cxx
```diff
@@ -234,9 +236,93 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+        return CommandResult::OK;
+    }
+    FileReader *reader = new FileReader(art_name);
+
+    const char *name = PathTraitsFS::GetBase(art_name.c_str());
+    
+    const unsigned long chunk_size = 8192;
```
MaxKellermann: Why `unsigned long`? This is the worst possible choice, because it pretends to be large-file capable, but it's not.

src/command/FileCommands.cxx
```diff
@@ -234,9 +236,93 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+    AllocatedPath art_name = find_art(directory);
+    if (art_name.IsNull()) {
+        if (offset > 0) {
+            r.Error(ACK_ERROR_NO_EXIST, "No art found");
+            return CommandResult::ERROR;
+        }
+        r.Write("size: -1\n");
+        return CommandResult::OK;
+    }
+    FileReader *reader = new FileReader(art_name);
```
MaxKellermann: This allocation can leak memory. Why a dynamic (and unsafe) allocation, anyway?

src/command/FileCommands.cxx
```diff
@@ -234,9 +236,93 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+    const unsigned long chunk_size = 8192;
+
+    void *art_data = malloc(chunk_size);
+
+    reader->Seek(offset);
+    size_t read_size = reader->Read(art_data, chunk_size);
+    
+    r.Format("size: %" PRIu64 "\n"
+             "file: %s\n"
+             "offset: %" PRIu64 "\n"
```
MaxKellermann: `PRIu64` is the wrong format for `unsigned long`.

src/command/FileCommands.cxx
```diff
@@ -234,9 +236,93 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+        r.Write("size: -1\n");
+        return CommandResult::OK;
+    }
+    FileReader *reader = new FileReader(art_name);
+
+    const char *name = PathTraitsFS::GetBase(art_name.c_str());
+    
+    const unsigned long chunk_size = 8192;
+
+    void *art_data = malloc(chunk_size);
```
MaxKellermann: Again, a memory leak due to unsafe dynamic allocation.

src/command/FileCommands.cxx
tokyovigilante: Hi Max, thanks for the feedback. As I mentioned I have no real C++ experience, which presumably shows. What would be a more appropriate container for this string array?

src/command/FileCommands.cxx
tokyovigilante: This is intended to signal the client that the art does not exist. If the client has requested a chuck with an offset where a file does not exist, this is regarded as client error. Happy to change this.

src/command/FileCommands.cxx
tokyovigilante: This has been changed to a static allocation, but I do need to include closing the file reader in all exit points from the function.

src/command/FileCommands.cxx
MaxKellermann: But that was `ACK_ERROR_NO_EXIST`..

src/command/FileCommands.cxx
tokyovigilante: Noted. In theory I suppose the function should be able to return a large file, but in practice this is probably not a valid concern given the files we're expecting to work with.

src/command/FileCommands.cxx
tokyovigilante: Yup but only if an offset was requested. I would expect the first request would be without an offset, and so the command returns OK with a file size of -1. If the client requests an offset without checking the file exists first, I think that should be an error.

src/command/FileCommands.cxx
MaxKellermann: An array would be better:

    static constexpr const char art_names[] = {...

.. because it can be filled at compile time, and requires no run-time code at all. Additionally, your version allocates memory at runtime for both `std::vector` and each `std::string` instance (in addition to the compile-time memory allocated for the same data).

No big deal, that's a simple thing to change - maybe we should postpone small things like that until the big things are done.

src/command/FileCommands.cxx
tokyovigilante: Not sure how to get a buffer to read into then write to socket otherwise in C++. Is there a buffer type I should use?

src/command/FileCommands.cxx
tokyovigilante: Sure, will change it to a 32 bit unsigned as above.

src/command/FileCommands.cxx
tokyovigilante: OK makes sense, ideally we'd allow the user to specify these based on their library configuration too, but I haven't delved into the config files as yet.

src/command/FileCommands.cxx
tokyovigilante: Are there any other reasons offhand it isn't large file capable? Clearly we shouldn't have 64 bit chunks, but the offset is `unsigned long long` and presumably the `FileReader` is just passing through to Unix or Windows system calls.

src/command/FileCommands.cxx
MaxKellermann: If you declare `size: -1` to indicate an error, then I consider this badly designed. You can use magic values (in band signalling) for binary protocols (if they are documented well!), but for text protocols where we have good error reporting (`ACK_ERROR_...`), there is no need for such hacks. Giving a wrong offset is an obvious client error, because the client does know the size already from the first response. So this error condition could be translated to `ACK_ERROR_ARG`.

src/command/FileCommands.cxx
MaxKellermann: Why? Closing the object is automatic in C++.

src/command/FileCommands.cxx
MaxKellermann: We can always make that dynamic/configurable later; and once we do that, we'll have one global `std::vector` instead of allocating a new one for each function call.

src/command/FileCommands.cxx
tokyovigilante: Sorry, I'm not being very clear. I don't think asking for a file that doesn't exist is an error necessarily, as the client won't know until it asks. Therefore I'm not sure the function should return an error. So if we send `album art xyz.mp3` we get back
`size: -1\nOK\n` signalling the file doesn't exist. I don't consider this an error though.

However if the client then asks for `album art xyz.mp3 31268` I think this is an error, and then return `ACK_ERROR_NO_EXIST`. I can change to `ACK_ERROR_ARG` in this scenario if you think thats a better fit.

src/command/FileCommands.cxx
MaxKellermann: Example:

    std::unique_ptr<uint8_t[]> art_data(new uint8_t[chunk_size]);

But I wouldn't do that; we already have a reasonable upper bound on the chunk size, so let's just allocated that 8kB buffer on the stack. I'm always cautious with allocating large buffers on the stack, but 8kB is still ok.

src/command/FileCommands.cxx
tokyovigilante: OK, I can check for this condition. Is there any way to get the equivalent enclosing folder for non-local paths?

src/command/FileCommands.cxx
MaxKellermann: Oh yes, asking for a file that doesn't exist **is** an error! That one should be `ACK_ERROR_NO_EXIST`. It's the natural choice for "doesn't exist", because that's the very meaning of this ACK code!

src/command/FileCommands.cxx
tokyovigilante: Good to know! I take my hat off to C++. Orthogonal to this though is that to me it seems wasteful to create a new reader for each chunk, as the client presumably intends to request the whole file. I'm not sure of an elegant way to store this between requests though.

src/command/FileCommands.cxx
MaxKellermann: But you didn't use `unsigned long long`, but only `unsigned long`. We know that `unsigned long` is only 32 bit on some platforms (e.g. 32 bit Linux, and all versions of Windows, even 64 bit Windows). We don't know a platform where `unsigned long long` is smaller than 64 bit, but I don't think there's a guarantee on that.
If you want it large-file aware, use `off_t` or `uint64_t` (the latter is what MPD uses internally for file sizes, so let's use that).
It might sound ridiculous to support large files for album art - but that's a matter of "let's do file sizes consistently large-file-aware", a matter of principle. If you break with that rule, it may (and will) haunt you some day in the future.

src/command/FileCommands.cxx
tokyovigilante: Ok. I guess it's semantics, I'm guess what I'm really saying is that the question the client is asking on first request is "does the file exist?" rather than "just give me the file, I don't care if it exists or not." But if you want to treat it as an error I'm happy with that.

src/command/FileCommands.cxx
MaxKellermann: No, but we should have such a function, maybe located in `util/UriUtil.?xx`.

src/command/FileCommands.cxx
MaxKellermann: Don't care about this waste for now. It's only a small problem, and if it turns out to be wasteful, we can optimize later. Optimizing it would require creating lots of ugly code, and elegant code is more important than fast code.

src/command/FileCommands.cxx
tokyovigilante: Sure, I totally agree. No reason not to future-proof. I'm just used to using 64 bit platforms. I think the offset is already `unsigned long long` and the chunk size should probably be 32 bit anyway, but i'll change it to `uint64_t`.

src/command/FileCommands.cxx
MaxKellermann: It may be an expected error. The client doesn't need to feel bad for saying "gimme the file" and then get shouted at my MPD... it's ok!

src/command/FileCommands.cxx
tokyovigilante: Ok sounds good. I agree, it would get very messy very quickly.

> elegant code is more important than fast code.

Spoken like a true programmer ;)

src/command/FileCommands.cxx
tokyovigilante: Cool, so use `alloca`?

MaxKellermann: You like it complicated, do you? ;-)

    constexpr size_t CHUNK_SIZE = 8192;
    uint8_t buffer[CHUNK_SIZE];

Too simple?

src/command/FileCommands.cxx
tokyovigilante: No, just means I need to change my client library a bit ;)

src/command/FileCommands.cxx
tokyovigilante: Heh it did seem too easy...

MaxKellermann:
See https://github.com/MusicPlayerDaemon/MPD/pull/86#issuecomment-317654711 for why I havn't reviewed any of your new commits.

tokyovigilante:
> See #86 (comment) for why I havn't reviewed any of your new commits.

It's my usual practice to try to commit multiple changes with different purposes as separate commits. Each is functional, the CI build failures are due to the linux config not finding std::vector before it was removed. I'm happy to fold them into larger changes if it makes it easier to review though. 

I'm currently implementing support for the new command in my client so will relook at the server changes once that's done.

tokyovigilante:
I've implemented support for the command in my client, and an immediate issue is the speed of transfer. 

I'm getting about 56KB/s over a local connection. I'm testing with an 8.5MB image (I know it's big, but obviously robustness is key if this is the basis of a general binary transfer protocol) and it took around 1:35 to transfer.  

If I push the `chunk_size` up this improves, reducing down to a few seconds with a 640KB buffer. CPU utilisation is about 1% for the client and 40% for mpd, so clearly the current send function is slow. I suspect it's the file I/O, but will run mpd through Instruments to debug.

I wonder whether a solution is to allow the client to specify a chunk size, but then send multiple smaller chunks in the same call up to the client-specified size? i.e.
`albumart [FILE] [CHUNKSIZE] [OFFSET]`
`albumart "file.mp3" 512K 0`
Then response
`size: 1024768`
`transfer: 512K`
`file: art.jpg`
`sequence: 1 of 64 (64 chunks of 8192 bytes in 512K)`
`offset: 0`
`binary: 8192`
`<8KB binary>`
`OK`
`size: 1024768`
`transfer: 512K`
`file: art.jpg`
`sequence: 2 of 64`
`offset: 8192`
`binary: 8192`
`<8KB binary>`
`OK`
and so on. A bit more verbose but I think sending large files in 8K chunks with an open and close each time is relatively inefficient. I'll have a look at the function in the meantime and see if the I/O is indeed what's slow.

tokyovigilante:
OK, I take it all back, tried on my laptop and the 8MB file is transferring in about a second so the issue may be with my setup. The (slow) desktop is reading over a mounted SMB share so maybe this is the issue. Needs more debugging but the approach seems sound.

MaxKellermann:
It is good (and necessary) to split changes into multiple commits if different things are changed, for example if you need an internal API change for your new feature. One change should be one commit, multiple changes should be multiple commits. 
But for code review, it's bad to keep multiple iterations of the very same change in multiple commits. No separate fixup commits for unmerged commits; amend the existing unmerged commits instead.
Do you understand the difference?

tokyovigilante:
I've discovered that my SMB issues are due to the SMB client implementation in macOS being, to put it mildly, rubbish. I built Samba for macOS to test, and then mpd with libsmbclient support. This required a stream reader implementation of the album art command, which I've implemented. Pleasingly, transferring the 8MB test file now takes ~5sec rather than 1:30+.

Current commit folds all the changes into one. I've also enforced specifying an `offset`, as it only makes sense without an offset once during the retrieve (i.e. the first chunk).

I'm pretty happy with it other than ideally the art file names could be specified in the config, but otherwise seems to work pretty well.

doc/protocol.xml
```diff
@@ -2024,7 +2049,7 @@ OK
             </para>
 
             <programlisting>listmounts
-mount: 
+mount:
```
MaxKellermann: Whitespace change?

src/command/AllCommands.cxx
```diff
@@ -84,6 +84,7 @@ static constexpr struct command commands[] = {
 	{ "add", PERMISSION_ADD, 1, 1, handle_add },
 	{ "addid", PERMISSION_ADD, 1, 2, handle_addid },
 	{ "addtagid", PERMISSION_ADD, 3, 3, handle_addtagid },
+    { "albumart", PERMISSION_READ, 2, 2, handle_album_art },
```
MaxKellermann: Wrong indent

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
 	case LocatedUri::Type::ABSOLUTE:
 		return read_stream_comments(r, located_uri.canonical_uri);
 
-	case LocatedUri::Type::RELATIVE:
```
MaxKellermann: Why these whitespace changes?

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+        return read_file_comments(r, located_uri.path);
+    }
+
+    gcc_unreachable();
+}
+
+InputStreamPtr
+find_stream_art(const char *directory)
+{
+    static const std::string art_names[] = {
```
MaxKellermann: No `std::string`! And make it `constexpr`. Which is impossible with `std::string`...

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+    
+    for(const std::string& name: art_names) {
+        const std::string art_file = PathTraitsUTF8::Build(directory, name.c_str());
+        
+        try {
+            Mutex mutex;
+            Cond cond;
+            
+            auto is = InputStream::OpenReady(art_file.c_str(), mutex, cond);
+            return is;
```
MaxKellermann: This will crash, because `mutex` and `cond` are still referenced by the `InputStream`, but are running out of scope.

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+        } catch (const std::exception &e) {
+            continue;
```
MaxKellermann: Pointless `continue`?

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+{
+    std::string art_directory = PathTraitsUTF8::GetParent(uri);
+    InputStreamPtr is = find_stream_art(art_directory.c_str());
+    
+    if (is == nullptr) {
+        r.Error(ACK_ERROR_NO_EXIST, "No art file exists");
+        return CommandResult::ERROR;
+    }
+    
+    if (!is->IsSeekable()) {
```
MaxKellermann: Why is seeking unconditionally necessary at this point?

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+    const size_t art_file_size = is->GetSize();
+    
+    constexpr size_t CHUNK_SIZE = 8192;
+    uint8_t buffer[CHUNK_SIZE];
+    uint32_t read_size;
+    try {
+        is->Seek(offset);
+        read_size = is->Read(&buffer, CHUNK_SIZE);
+    } catch (const std::exception &e) {
+        PrintError(r, std::current_exception());
```
MaxKellermann: Why catch this error?

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+        return CommandResult::ERROR;
+    }
+    
+    r.Format("size: %" PRIu64 "\n"
+             "file: %s\n"
+             "offset: %" PRIu64 "\n"
+             "binary: %u\n",
```
MaxKellermann: Wrong format string for `uint32_t`.

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+    if (!is->KnownSize()) {
+        r.Error(ACK_ERROR_NO_EXIST, "Cannot get size for art file stream");
+        return CommandResult::ERROR;
+    }
+    
+    const size_t art_file_size = is->GetSize();
+    
+    constexpr size_t CHUNK_SIZE = 8192;
+    uint8_t buffer[CHUNK_SIZE];
+    uint32_t read_size;
```
MaxKellermann: Why `uint32_t`?

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+    try {
+        is->Seek(offset);
+        read_size = is->Read(&buffer, CHUNK_SIZE);
+    } catch (const std::exception &e) {
+        PrintError(r, std::current_exception());
+        return CommandResult::ERROR;
+    }
+    
+    r.Format("size: %" PRIu64 "\n"
+             "file: %s\n"
```
MaxKellermann: What's the point of telling the file name?

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+             "offset: %" PRIu64 "\n"
```
MaxKellermann: What's the point of telling the offset? The client knows it already, obviously. (I question my own suggestion, I know.)

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+        if (!FileExists(art)) {
+            continue;
+        }
+        return art;
+    }
+    return AllocatedPath::Null();
+}
+
+static CommandResult
+read_file_art(Response &r, const AllocatedPath uri, const uint64_t offset)
```
MaxKellermann: Lots and lots of duplicate code. Do we need two code paths, can't we merge them, do we need a special case for local files?

doc/protocol.xml
tokyovigilante: Noted.

src/command/AllCommands.cxx
tokyovigilante: Noted.

src/command/FileCommands.cxx
tokyovigilante: Copy and paste in Xcode with automatic reindent, will remove those.

src/command/FileCommands.cxx
tokyovigilante: Thanks, that slipped through in the rewrite.

src/command/FileCommands.cxx
tokyovigilante: Oh right, rookie mistake. So I'll need to declare them in the main function and pass by reference?

src/command/FileCommands.cxx
tokyovigilante: The intent is just to carry on with the next filename option if the `InputStream` open fails, assuming because the file isn't there. Is there a better way to handle the error? In Swift you can't have functions that throw outside a `try` block, even if that block is in a calling function.

src/command/FileCommands.cxx
tokyovigilante: Presumably if the file can't seek we can't return a chunk at an `offset`, unless you propose reading the whole file up to the offset and discarding the unneeded preceding bytes. This seems pretty inefficient.

src/command/FileCommands.cxx
tokyovigilante: We're only ever going to return 8K bytes, which will fit in a 32 bit int, (or 16 for that matter) but a `size_t` is returned from the read call so I'll change it to a 64 bit type.

src/command/FileCommands.cxx
tokyovigilante: Surely that's good practice? The file I/O could fail even if we successfully open the stream.

src/command/FileCommands.cxx
tokyovigilante: So the client can infer the file type. I personally don't need it as my client infers the image type from the header, but I do cache it as a local file, and this would seem useful information generally.

src/command/FileCommands.cxx
tokyovigilante: The client can then retrieve asynchronously without storing state. I personally am not at the moment, but you might want to bulk send a number of requests for chunks, then you can receive/process them in any order if you have the filename and offset. Obviously they are returned FIFO, but this make the client's job less complex.

src/command/FileCommands.cxx
tokyovigilante: Noted, will change all the types back to 64 bit.

src/command/FileCommands.cxx
tokyovigilante: True, it's largely because of all the conversions between std::string and C-strings. I can probably tidy this up. 

If you're happy to use an `InputStream` for local files, we can just have one set of functions, but I modelled the stream reader path on the `readcomments` functions above (hence the copy/paste errors!), assuming that's how you intended this sort of thing to work.

src/command/FileCommands.cxx
```diff
@@ -224,12 +230,212 @@ handle_read_comments(Client &client, Request args, Response &r)
...
 }
+
+CommandResult
+handle_album_art(Client &client, Request args, Response &r)
+{
+    assert(args.size == 2);
+    
+    const char *const uri = args.front();
+    uint64_t offset = 0;
+    if (args.size == 2) {
```
tokyovigilante: Note to myself, don't need this check.

src/command/FileCommands.cxx
MaxKellermann: Yes.

src/command/FileCommands.cxx
MaxKellermann: My comment was not about handling the error, but about the `continue` statement, which has no effect here.

src/command/FileCommands.cxx
MaxKellermann: Only with a positive `offset`, which is conditional. I asked for "unconditionally".

src/command/FileCommands.cxx
MaxKellermann: ... and now why 64 bit? `size_t` is the correct type for sizes in memory. Why not use that? Too easy?

src/command/FileCommands.cxx
MaxKellermann: I know it can fail, but why is it "good practice" to have a redundant exception handler? I think redundant code is almost never good practice.

src/command/FileCommands.cxx
MaxKellermann: So all you really wanted to tell the client was the file type?
But you told him the whole name, which on contemporary operating systems embeds a file type description. That design was bad from the start, but that's another story.
Anyway, remember my suggestion?

src/command/FileCommands.cxx
MaxKellermann: I don't want to support dumb clients which ask MPD for something, and instantly forget what they asked for.

src/command/FileCommands.cxx
MaxKellermann: Nooo! A file offset is 64 bit per definition in MPD, but a buffer size is `size_t`, as noted above.

src/command/FileCommands.cxx
MaxKellermann: `InputStream` is fine. It adds a tiny bit of overhead to local file operations, but that's justified if it reduces code duplication.
`readcomments` is different; it needs special cases for local files, because some tag readers cannot operate on `InputStream`. We could reduce code duplication, but we would also lose features, which is not acceptable.

src/command/FileCommands.cxx
tokyovigilante: I'm not quite sure what you mean here, are you suggesting all the stream types can seek forward? That would make this check redundant.

src/command/FileCommands.cxx
tokyovigilante: Fixed.

src/command/FileCommands.cxx
tokyovigilante: I'm not sure how this is redundant, if the seek or read can fail and throw an exception, and we don't handle the exception, won't the server crash?

src/command/FileCommands.cxx
tokyovigilante: If you mean sending the MIME type, then I'm not sure how you can send that without parsing the file, which seems a little bit out of scope. I think the file name is a good stand-in here, acknowledging your point about file extensions being bad practice, it is what it is.

src/command/FileCommands.cxx
MaxKellermann: No, think `offset==0`, which requires no seeking ability at all. And it is the most common value. MPD should make a best effort to fulfill the client's request.
Denying the most common request just because the `InputStream` is not seekable, when seeking is not necessary, is somewhat dull.

src/command/FileCommands.cxx
MaxKellermann: No, because https://github.com/MusicPlayerDaemon/MPD/blob/master/src/command/AllCommands.cxx#L421 already handles exceptions thrown from within client commands, and translates them to an error response.
Note that you didn't really answer my initial question about the `catch` block.

src/command/FileCommands.cxx
tokyovigilante: I'd disagree that this means the client is not well-designed, but I do store the request in mine and match it with the response, so removing this wouldn't effect my client. 

However I think it's reasonable for a binary protocol to fully describe what it's sending, so would favour keeping the file name and offset in the response. The server knows both of these to load the chunk, and the overhead for 8KB chunks is in the order of 0.5%.

src/command/FileCommands.cxx
tokyovigilante: Fixed.

src/command/FileCommands.cxx
tokyovigilante: Have rewritten the code to use `InputStream` for local and remote reads. I can see how a more generic function to read a chunk from an `InputStream` and write the response to the client would be useful, so am happy to refactor the code if you'd prefer that.

src/command/FileCommands.cxx
MaxKellermann: It is okay (for now) to guess the MIME type from the file name (`*.jpg` translates to `image/jpeg`), even more so because you have a tiny hard-coded list of file names.
Or just omit the MIME type for now. Detecting the file format of an image file is trivial enough. If necessary, we can add it later.
In any case, it is better to use MIME types than file name extensions (or worse: the full file name when only the suffix is interesting).

src/command/FileCommands.cxx
tokyovigilante: Oh right, I hadn't appreciated that. That can go then.

src/command/FileCommands.cxx
tokyovigilante: That's fair. OK I'll remove it for now.

src/command/FileCommands.cxx
MaxKellermann: I don't think that's reasonable. Which protocol does that? HTTP, for example, does not echo the request URI in a response, so it doesn't "fully describe what it's sending". HTTP does describe the `Content-Range`, but only because that may be an optional feature with implicit fallback to transferring the whole file - which is not the case for MPD. There's the `Content-Disposition` header which may contain a file name, but that was a kludge from a time when request URIs could not be mapped to scripts properly, so your browser won't save to local file `foo.php`.
"The server knows both of these to load the chunk" and the client must know it as well.

tokyovigilante: OK, I take your point. Removed as well.

src/command/FileCommands.cxx
tokyovigilante: OK. I assume most clients will intend to request the entire file, so we are just moving the failure to the following request for `offset 8192`, but you're right, we can satisfy `offset 0` without seeking.

tokyovigilante: Alternatively I guess we could just rely on the actual `Seek` call to throw an exception.

MaxKellermann: Agree. `IsSeekable()` exists only so (expensive) preparations can be skipped, but you can just be optimistic and then handle the error.

src/command/FileCommands.cxx
```diff
@@ -36,11 +36,17 @@
 #include "fs/AllocatedPath.hxx"
 #include "fs/FileInfo.hxx"
 #include "fs/DirectoryReader.hxx"
+#include "fs/FileSystem.hxx"
+#include "fs/io/FileReader.hxx"
```
MaxKellermann: Unused include

src/command/FileCommands.cxx
```diff
@@ -36,11 +36,17 @@
 #include "fs/AllocatedPath.hxx"
 #include "fs/FileInfo.hxx"
 #include "fs/DirectoryReader.hxx"
+#include "fs/FileSystem.hxx"
```
MaxKellermann: Unused include?

src/command/FileCommands.cxx
```diff
@@ -36,11 +36,17 @@
...
+#include "fs/io/FileReader.hxx"
+#include "input/InputStream.hxx"
 #include "LocateUri.hxx"
 #include "TimePrint.hxx"
 
 #include <assert.h>
 #include <inttypes.h> /* for PRIu64 */
+#include <string>
```
MaxKellermann: Unused include

src/command/FileCommands.cxx
```diff
@@ -36,11 +36,17 @@
...
+#include "thread/Mutex.hxx"
+#include "thread/Cond.hxx"
```
MaxKellermann: Please move these up to the other "internal" includes

src/command/FileCommands.cxx
```diff
@@ -233,3 +239,103 @@ handle_read_comments(Client &client, Request args, Response &r)
 
 	gcc_unreachable();
 }
+
+InputStreamPtr
+find_stream_art(const char *directory, Mutex &mutex, Cond &cond)
+{
+    static constexpr char const * art_names[] = {
```
MaxKellermann: Wrong indent - everything should be indented with tabs, not 4 spaces. See https://www.musicpd.org/doc/developer/code_style.html

doc/protocol.xml
```diff
@@ -1564,6 +1564,31 @@ OK
...
+              art image file at offset <varname>OFFSET</varname>.
+            </para>
+            <para>
+              Uses the filename "cover" with any of ".png, .jpg,
+              .tiff, .bmp".
+            </para>
+            <para>
+              Returns the file size and actual number
+              of bytes read at the requested offset before
+              transmitting the chunk as raw bytes.
```
MaxKellermann: Description missing what "raw bytes" means (the general binary protocol).

src/command/FileCommands.cxx
```diff
@@ -233,3 +239,103 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+        "cover.tiff",
+        "cover.bmp"
+    };
+    
+    for(const auto name: art_names) {
+        const char *art_file = PathTraitsUTF8::Build(directory, name).c_str();
+        
+        try {
+            auto is = InputStream::OpenReady(art_file, mutex, cond);
+            return is;
```
MaxKellermann: Make this a single `return` line without the `is` variable

src/command/FileCommands.cxx
```diff
@@ -233,3 +239,103 @@ handle_read_comments(Client &client, Request args, Response &r)
 
 	gcc_unreachable();
 }
+
+InputStreamPtr
+find_stream_art(const char *directory, Mutex &mutex, Cond &cond)
```
MaxKellermann: There should be API documentation explaining what `directory` means - i.e. it's an UTF-8 URI, not a local filesystem path in filesystem encoding.

src/command/FileCommands.cxx
```diff
@@ -233,3 +239,103 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+{
+    static constexpr char const * art_names[] = {
+        "cover.png",
+        "cover.jpg",
+        "cover.tiff",
+        "cover.bmp"
+    };
+
```
MaxKellermann: Line consisting only of trailing spaces (and there are LOTS more of that, my editor is all in red highlighting those)

src/command/FileCommands.cxx
```diff
@@ -233,3 +239,103 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+            return is;
+        } catch (const std::exception &e) {}
+    }
+    return nullptr;
+}
+
+static CommandResult
+read_stream_art(Response &r, const char *uri, size_t offset)
+{
+    const char *art_directory = PathTraitsUTF8::GetParent(uri).c_str();
```
MaxKellermann: This will crash because the returned `std::string` will be freed after the statement, and the `c_str()` result points to that freed memory.

src/command/FileCommands.cxx
```diff
@@ -233,3 +239,103 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+    r.Write(buffer, read_size);
+    r.Write("\n");
+    
+    return CommandResult::OK;
+}
+
+static CommandResult
+read_db_art(Client &client, Response &r, const char *uri, const uint64_t offset)
+{
+    const Storage *storage = client.GetStorage();
```
MaxKellermann: This will fail to compile if `ENABLE_DATABASE` is not defined.

src/command/FileCommands.cxx
```diff
@@ -233,3 +239,103 @@ handle_read_comments(Client &client, Request args, Response &r)
...
+    size_t offset = args.ParseUnsigned(1);
+    
+    const auto located_uri = LocateUri(uri, &client
+#ifdef ENABLE_DATABASE
+                                       , nullptr
+#endif
+                                       );
+    
+    switch (located_uri.type) {
+        case LocatedUri::Type::ABSOLUTE:
```
MaxKellermann: Wrong indent. Must be same indent level as the `switch`.

tokyovigilante: Fixed.

src/command/FileCommands.cxx
tokyovigilante: Fixed.

src/command/FileCommands.cxx
tokyovigilante: Fixed.

src/command/FileCommands.cxx
tokyovigilante: Fixed.

src/command/FileCommands.cxx
tokyovigilante: Fixed.

src/command/FileCommands.cxx
tokyovigilante: Fixed.

src/command/FileCommands.cxx
tokyovigilante: Fixed.

src/command/FileCommands.cxx
tokyovigilante: Fixed.

src/command/FileCommands.cxx
tokyovigilante: Fixed.

doc/protocol.xml
tokyovigilante: This seems self-explanatory to me.

src/command/FileCommands.cxx
tokyovigilante: This is an internal function, where would you like it documented?

MaxKellermann: Add Doxygen-style comments like the rest of the code does. "Obvious" things don't need to be documented (for some definition of "obvious"), but when it's about file names, you almost always need documentation, because the difference between filesystem charset and UTF-8 is important, and without explicit documentation, it's too easy to get wrong.
And if it's internal, then it should be `static`.

doc/protocol.xml
MaxKellermann: Not to me. For example, your code adds an extra newline after the binary data, but that part is not documented. It should be explicitly explained when binary transfer starts - that part is not even mentioned in your text.

src/command/FileCommands.cxx
tokyovigilante: OK that's reasonable, will do.

doc/protocol.xml
tokyovigilante: Fair point, I thought this would be inferred based on the [overview](https://www.musicpd.org/doc/protocol/syntax.html#idm5) but I'm happy to make it more explicit.

tokyovigilante: Added more detail and example.

src/command/FileCommands.cxx
tokyovigilante: Added comment and made static. Also fixed another `c_str()` access bug.

MaxKellermann:
Invalid XML:

    doc/protocol.xml:1601: parser error : Opening and ending tag mismatch: listitem line 1584 and para
            </para>

MaxKellermann:
Build failure with database support disabled:

    src/command/FileCommands.cxx: In function 'CommandResult handle_album_art(Client&, Request, Response&)':
    src/command/FileCommands.cxx:342:2: error: expected primary-expression before '}' token

MaxKellermann:
The rest looks ok now, so if you fix those two errors, I'll merge it.
Please rebase on lastest master branch, that makes merging easier for me.

tokyovigilante:
OK have made those changes and rebased. Thanks again for your help with this.

MaxKellermann:
Another XML error:

    doc/protocol.xml:1607: parser error : StartTag: invalid element name
    <8192 bytes>

MaxKellermann:
And more build failures:

    src/command/FileCommands.cxx: In function 'CommandResult handle_album_art(Client&, Request, Response&)':
    src/command/FileCommands.cxx:328:58: error: too many arguments to function 'LocatedUri LocateUri(const char*, const Client*)'
      const auto located_uri = LocateUri(uri, &client, nullptr);
                                                              ^
    In file included from src/command/FileCommands.cxx:40:0:
    src/LocateUri.hxx:87:1: note: declared here
     LocateUri(const char *uri, const Client *client
     ^~~~~~~~~
    src/command/FileCommands.cxx:330:9: error: enumeration value 'RELATIVE' not handled in switch [-Werror=switch]
      switch (located_uri.type) {

tokyovigilante:
Have updated documentation and verified it compiles. I get an error unrelated to my code when compiling without the database enabled.

```
g++ -std=c++14 -Qunused-arguments -pthread -g -O0 -fvisibility=hidden -fno-threadsafe-statics -fmerge-all-constants -ffast-math -ftree-vectorize -ffunction-sections -fdata-sections -Wall -Wextra -Wmissing-declarations -Wshadow -Wpointer-arith -Wcast-qual -Wwrite-strings -Wsign-compare   -o src/mpd src/Main.o libmpd.a -L/usr/local/samba/lib -Wl,-rpath,/usr/local/samba/lib -lsmbclient libneighbor.a    libplaylist_plugins.a  -L/usr/local/Cellar/flac/1.3.2/lib -lFLAC     libdecoder.a libxiph.a -L/usr/local/Cellar/libogg/1.3.2/lib -logg -L/usr/local/Cellar/libvorbis/1.3.5_1/lib -L/usr/local/Cellar/libogg/1.3.2/lib -lvorbis -logg  -L/usr/local/Cellar/flac/1.3.2/lib -lFLAC      -L/usr/local/Cellar/fluid-synth/1.1.6/lib -lfluidsynth      libffmpeg.a -L/usr/local/Cellar/ffmpeg/3.3.2/lib -lavformat -lavcodec -lavutil   -lfaad  libinput.a -lcurl -L/usr/local/samba/lib -Wl,-rpath,/usr/local/samba/lib -lsmbclient   libffmpeg.a -L/usr/local/Cellar/ffmpeg/3.3.2/lib -lavformat -lavcodec -lavutil   libarchive.a -lbz2   liboutput_plugins.a       -framework OpenAL     libtag.a -L/usr/local/opt/libid3tag/lib -lid3tag -lz libfilter_plugins.a libpcm.a  -L/usr/local/Cellar/libsamplerate/0.1.9/lib -lsamplerate libencoder_plugins.a libxiph.a -L/usr/local/Cellar/libogg/1.3.2/lib -logg -lmp3lame  -L/usr/local/Cellar/flac/1.3.2/lib -lFLAC   -L/usr/local/Cellar/libvorbis/1.3.5_1/lib -L/usr/local/Cellar/libogg/1.3.2/lib -lvorbisenc -lvorbis -logg libmixer_plugins.a   libconf.a libbasic.a libevent.a libthread.a libnet.a libfs.a -lz  libsystem.a libutil.a libicu.a    -framework AudioUnit -framework CoreAudio -framework CoreServices
Undefined symbols for architecture x86_64:
  "Song::Free()", referenced from:
      Song::LoadFromArchive(ArchiveFile&, char const*, Directory&) in libmpd.a(libmpd_a-SongUpdate.o)
  "Song::NewFile(char const*, Directory&)", referenced from:
      Song::LoadFromArchive(ArchiveFile&, char const*, Directory&) in libmpd.a(libmpd_a-SongUpdate.o)
  "Directory::GetName() const", referenced from:
      Song::UpdateFileInArchive(ArchiveFile&) in libmpd.a(libmpd_a-SongUpdate.o)
ld: symbol(s) not found for architecture x86_64
```

tokyovigilante:
Thanks!

MaxKellermann:
@raftopoulos is right - your latest commit which I merged still has this `std::string` use-after-free.

tokyovigilante:
Oh right, so it does. My bad. Shall I reupdate this patch or a new one on `master`?

tokyovigilante:
The `weird binary` is a feature though obviously, that's the first 8KB of the image data. The command is intended to be used by a client which can request the chunks sequentially, then display the reconstructed image file. Not so helpful from the command line, will be interpreted ASCII control codes everywhere.

MaxKellermann:
Once a commit has been merged, it must never again be amended. In that case, write a new commit.

tokyovigilante:
#97

PR #93, opened 2017-08-03
WIP HTSP

### LINKED ISSUE #92: [(large) feature] TVHeadend HTSP support
I just asked this on IRC; I'll just post it here too.

### About TVHeadend

[TVHeadend](https://tvheadend.org/) is a TV streaming server and recorder for Linux, FreeBSD and Android supporting DVB-S, DVB-S2, DVB-C, DVB-T, ATSC, ISDB-T, IPTV, SAT>IP and HDHomeRun as input sources.

On top of those input sources, anything that's streamable (without DRM, usually) TV or radio can get into TVHeadend.

Tvheadend offers the HTTP (VLC, MPlayer), HTSP (Kodi, Movian) and SAT>IP streaming.

Multiple EPG sources are supported (over-the-air DVB and ATSC including OpenTV DVB extensions, XMLTV, PyXML).

**TL;DR:** *TVHeadend streams radio, TV, from any source over the network.*

### What's currently possible with `mpd` and TVHeadend

I'm currently listening to "Radio 1" (Belgian national radio) using the `.m3u` playlist that TVHeadend generates (http[s]://[tvheadend]/playlist). This playlist contains a randomized session key, such that the player client does not have to authenticate with TVHeadend over HTTP AUTH.

An image as example (*note that the first four songs are from a CD*):

Example of the current setup in ncmpcpp

As you see, no channel name is available, nor the current programme name, although TVHeadend has this information. It is simply not transferred over this https stream.

# Actual feature request

*Note that I am willing to develop the feature myself; I just need confirmation that it would eventually be mainlined*

TVHeadend has a proprietary protocol called `htsp` (home theatre system streaming protocol, afaik), which can be queried for channels, EPG information, and can be used to stream the actual audio.

I suggest to have `mpd` have a configuration item, e.g.,

```
tvheadend {
  host "localhost:9982"
  user "mpd"
  pass "pass_mpd"
}
```

so it connects/tries to connect on MPD startup, and queries for channels, which it makes available for playback.

### Things to consider

- There will not be a clear mapping from (channel name, playing programme) -> (artist, album, genre, title) any more.
- More things I didn't think about.

rubdos:
Closes #92

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "vendor/pvr.hts"]
+	path = vendor/pvr.hts
+	url = https://github.com/kodi-pvr/pvr.hts.git
--- a/Makefile.am
+++ b/Makefile.am
@@ -260,6 +260,14 @@ ALSA_SOURCES = \
 	src/lib/alsa/Version.cxx src/lib/alsa/Version.hxx \
 	src/lib/alsa/NonBlock.cxx src/lib/alsa/NonBlock.hxx
 
+if ENABLE_HTSP
+libmpd_a_SOURCES += \
+	src/lib/libhts/htsbuf.c src/lib/libhts/htsmsg_binary.c \
+	src/lib/libhts/htsmsg.c src/lib/libhts/htsstr.c \
+	src/lib/libhts/net_posix.c src/lib/libhts/sha1.c
+endif
+# TODO: if Windows -> net_windows
+
 #
 # Android native library
 #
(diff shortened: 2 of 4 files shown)


rubdos:
For cross linking fun: https://gitlab.com/rubdos/htsp-rs/merge_requests/1

MaxKellermann:
As discussed on IRC, I'd like to avoid git submodules.

eutampieri:
Is implementing this possible? Are sub modules the only blocking issue?

MaxKellermann:
> Are submodules the only blocking issue?

I didn't look further, but looking at this again: there's no code in this PR. It just adds external sources, but they do nothing. They add nothing to MPD. There's no explanation in the commit message. I have no idea what this PR really does.

rubdos:
There's `WIP` in the title of this PR, which meant that it was a work in progress. I'm not planning to continue this. It's a lot of work, and I don't have a use for it any more.

PR #96, opened 2017-08-17
Fix EventPipe file descriptors

tokyovigilante:
Current MPD head crashes on starting playback on macOS Sierra. On investigation this seems to be due to a typo in `EventPipe.cxx` attempting to use the read file descriptor as both R/W.

This patch fixes playback on my system.


MaxKellermann:
D'oh!

PR #97, opened 2017-08-17
Fix std::string read after free in read_stream_art

tokyovigilante:


tokyovigilante:
And now with actual testing...

PR #100, opened 2017-08-21
OSX mixer

matthewleon:
I have based this off of v0.20.x. I'm not sure what the policy is on this. If you would prefer a version based off of `master`, I can do that as well.

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -20,6 +20,7 @@
 #include "config.h"
 #include "OSXOutputPlugin.hxx"
 #include "../OutputAPI.hxx"
+#include "mixer/MixerList.hxx"
 #include "util/ScopeExit.hxx"
 #include "util/RuntimeError.hxx"
 #include "util/Domain.hxx"
@@ -53,6 +54,9 @@ struct OSXOutput {
 	boost::lockfree::spsc_queue<uint8_t> *ring_buffer;
 
 	OSXOutput(const ConfigBlock &block);
+
+	int GetVolume();
+	void SetVolume(unsigned new_volume);
 };
 
 static constexpr Domain osx_output_domain("osx_output");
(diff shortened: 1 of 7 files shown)


src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -103,6 +107,37 @@ OSXOutput::OSXOutput(const ConfigBlock &block)
...
+OSXOutput::GetVolume()
+{
+	AudioUnitParameterValue dvolume;
+	char errormsg[1024];
+
+	OSStatus status = AudioUnitGetParameter(au, kHALOutputParam_Volume,
+			kAudioUnitScope_Global, 0, &dvolume);
+	if (status != noErr) {
+		osx_os_status_to_cstring(status, errormsg, sizeof(errormsg));
+		FormatWarning(osx_output_domain, "unable to get volume: %s", errormsg);
```
MaxKellermann: Throw an exception instead of logging a warning.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -103,6 +107,37 @@ OSXOutput::OSXOutput(const ConfigBlock &block)
...
+		return 100;
+	}
+	return static_cast<int>(dvolume * dvolume * 100.0);
```
MaxKellermann: This formula should be explained in a code comment so the next guy understands it.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -103,6 +107,37 @@ OSXOutput::OSXOutput(const ConfigBlock &block)
...
+OSXOutput::SetVolume(unsigned new_volume) {
+	char errormsg[1024];
+	AudioUnitParameterValue scaled_volume =
+		sqrt(static_cast<AudioUnitParameterValue>(new_volume) / 100.0);
+
+	OSStatus status = AudioUnitSetParameter(au, kHALOutputParam_Volume,
+			kAudioUnitScope_Global, 0, scaled_volume, 0);
+	if (status != noErr) {
+		osx_os_status_to_cstring(status, errormsg, sizeof(errormsg));
+		FormatWarning(osx_output_domain, "unable to set new volume %u: %s",
```
MaxKellermann: Throw exception, see above.

matthewleon:
> please submit a merge into master as soon as I've merged your (corrected) patch in v0.20.x, because I don't have a Mac and I can't test the build.

I am happy to do this. This should be ready for review again.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -103,6 +107,45 @@ OSXOutput::OSXOutput(const ConfigBlock &block)
...
+{
+	AudioUnitParameterValue dvolume;
+	char errormsg[1024];
+
+	OSStatus status = AudioUnitGetParameter(au, kHALOutputParam_Volume,
+			kAudioUnitScope_Global, 0, &dvolume);
+	if (status != noErr) {
+		osx_os_status_to_cstring(status, errormsg, sizeof(errormsg));
+		throw FormatRuntimeError("unable to get volume: %s", errormsg);
+		return 100;
```
MaxKellermann: Unreachable `return`.

matthewleon: Thanks. Fixed.

PR #106, opened 2017-09-01
Remove Content-Type check for webdav

sqozz:
This removes the Content-Type check if webdav is used as storage.
Several implementations of common WebDav servers (Synology File Stations, Nginx) do not correctly implement the RFC or in another way then expected from MPD. 
The RFC does not explicitly state that an client must check this header so therefore I completely removed the check.

This patch was tested against an password protected Nginx WebDav share and worked flawlessly. Reading and playing files from this share worked as expected.


MaxKellermann:
MPD does not and will not support servers which violate RFCs.

PR #109, opened 2017-09-12
Add sndio mixer, fix building on BSD

madroach:
Building on any unix with ``sin_size`` in ``sockaddr_in`` is broken.
I fix this not by using additional ``#ifdef``s, but by initializing the anonymous ``sockaddr_in`` with explicit field names. The ``sin_size`` field does not require initialization and is implicitly initialized to ``0``.

### DIFF
--- a/src/output/plugins/SndioOutputPlugin.cxx
+++ b/src/output/plugins/SndioOutputPlugin.cxx
@@ -19,6 +19,7 @@
 
 #include "config.h"
 #include "SndioOutputPlugin.hxx"
+#include "mixer/MixerList.hxx"
 #include "util/Domain.hxx"
 #include "Log.hxx"
 
@@ -47,8 +48,14 @@ SndioOutput::SndioOutput(const ConfigBlock &block)
 	:AudioOutput(0),
 	 device(block.GetBlockValue("device", SIO_DEVANY)),
 	 buffer_time(block.GetBlockValue("buffer_time",
-					 MPD_SNDIO_BUFFER_TIME_MS))
+					 MPD_SNDIO_BUFFER_TIME_MS)),
+	 raw_volume(SIO_MAXVOL)
 {
+}
+
+static void
+VolumeCallback(void *arg, unsigned int volume) {
+	((SndioOutput *)arg)->VolumeChanged(volume);
 }
 
 AudioOutput *
@@ -125,6 +132,15 @@ SndioOutput::Open(AudioFormat &audio_format)
 		throw std::runtime_error("Requested audio params cannot be satisfied");
 	}
 
+	// Set volume after opening fresh audio stream which does
+	// know nothing about previous audio streams.
+	sio_setvol(sio_hdl, raw_volume);
+	// sio_onvol returns 0 if no volume knob is available.
+	// This is the case on raw audio devices rather than
+	// the sndiod audio server.
+	if (sio_onvol(sio_hdl, VolumeCallback, this) == 0)
+		raw_volume = -1;
+
 	if (!sio_start(sio_hdl)) {
 		sio_close(sio_hdl);
 		throw std::runtime_error("Failed to start audio device");
@@ -148,9 +164,39 @@ SndioOutput::Play(const void *chunk, size_t size)
 	return n;
 }
 
+void
+SndioOutput::SetVolume(unsigned int volume) {
+	sio_setvol(sio_hdl, volume * SIO_MAXVOL / 100);
+}
+
+static inline unsigned int
+RawToPercent(int raw_volume) {
+	return raw_volume < 0 ? 100 : raw_volume * 100 / SIO_MAXVOL;
+}
+
+void
+SndioOutput::VolumeChanged(int _raw_volume) {
+	if (raw_volume >= 0 && listener != nullptr && mixer != nullptr) {
+		raw_volume = _raw_volume;
+		listener->OnMixerVolumeChanged(*mixer,
+		    RawToPercent(raw_volume));
+	}
+}
+
+unsigned int
+SndioOutput::GetVolume() {
+	return RawToPercent(raw_volume);
+}
+
+void
+SndioOutput::RegisterMixerListener(Mixer *_mixer, MixerListener *_listener) {
+	mixer = _mixer;
+	listener = _listener;
+}
+
 const struct AudioOutputPlugin sndio_output_plugin = {
 	"sndio",
 	sndio_test_default_device,
-	&SndioOutput::Create,
-	nullptr,
+	SndioOutput::Create,
+	&sndio_mixer_plugin,
 };
--- a/src/output/plugins/SndioOutputPlugin.hxx
+++ b/src/output/plugins/SndioOutputPlugin.hxx
@@ -18,6 +18,7 @@
  */
 
 #include "../OutputAPI.hxx"
+#include "mixer/Listener.hxx"
 
 #ifndef MPD_SNDIO_OUTPUT_PLUGIN_HXX
 #define MPD_SNDIO_OUTPUT_PLUGIN_HXX
@@ -25,15 +26,23 @@
 extern const struct AudioOutputPlugin sndio_output_plugin;
 
 class SndioOutput final : AudioOutput {
+	Mixer *mixer = nullptr;
+	MixerListener *listener = nullptr;
 	const char *const device;
 	const unsigned buffer_time; /* in ms */
 	struct sio_hdl *sio_hdl;
+	int raw_volume;
 
 public:
 	SndioOutput(const ConfigBlock &block);
 
 	static AudioOutput *Create(EventLoop &,
-				   const ConfigBlock &block);
+		const ConfigBlock &block);
+
+	void SetVolume(unsigned int _volume);
+	unsigned int GetVolume();
+	void VolumeChanged(int _volume);
+	void RegisterMixerListener(Mixer *_mixer, MixerListener *_listener);
 
 private:
 	void Open(AudioFormat &audio_format) override;
--- a/Makefile.am
+++ b/Makefile.am
@@ -1469,6 +1469,7 @@ if ENABLE_SNDIO
 liboutput_plugins_a_SOURCES += \
 	src/output/plugins/SndioOutputPlugin.cxx \
 	src/output/plugins/SndioOutputPlugin.hxx
+libmixer_plugins_a_SOURCES += src/mixer/plugins/SndioMixerPlugin.cxx
 endif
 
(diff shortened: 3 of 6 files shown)


src/net/IPv4Address.hxx
```diff
@@ -81,13 +81,9 @@ class IPv4Address {
...
 						      uint16_t port) {
 		return {
-#if defined(__APPLE__)
-			sizeof(struct sockaddr_in),
-#endif
-			AF_INET,
-			ToBE16(port),
-			address,
-			{},
+			.sin_family = AF_INET,
```
MaxKellermann: This is not portable. It is illegal C++ syntax, and works only because it uses a GCC extension.

src/output/plugins/SndioOutputPlugin.cxx
```diff
@@ -44,30 +45,21 @@ static constexpr unsigned MPD_SNDIO_BUFFER_TIME_MS = 250;
 
 static constexpr Domain sndio_output_domain("sndio_output");
 
-class SndioOutput final : AudioOutput {
```
MaxKellermann: This removal is obscure and hard to review, because it moves code and modifies it at the same time. Please make the move a separate commit.

src/output/plugins/SndioOutputPlugin.cxx
```diff
@@ -163,9 +158,34 @@ SndioOutput::Play(const void *chunk, size_t size)
...
+void
+SndioOutput::RegisterMixerListener(Mixer *_mixer, MixerListener *_listener) {
+	mixer = _mixer;
+	listener = _listener;
+}
+
 const struct AudioOutputPlugin sndio_output_plugin = {
 	"sndio",
 	sndio_test_default_device,
-	&SndioOutput::Create,
```
MaxKellermann: Why this change?

src/output/plugins/SndioOutputPlugin.cxx
```diff
@@ -163,9 +158,34 @@ SndioOutput::Play(const void *chunk, size_t size)
...
 
+void
+SndioOutput::SetVolume(unsigned int volume) {
+	sio_setvol(sio_hdl, volume * SIO_MAXVOL / 100);
+}
+
+void
+SndioOutput::VolumeChanged(unsigned int _raw_volume) {
+	raw_volume = _raw_volume;
+	if (listener != NULL && mixer != NULL) {
```
MaxKellermann: Use `nullptr` instead of the deprecated `NULL`.

src/output/plugins/SndioOutputPlugin.cxx
```diff
@@ -140,6 +132,9 @@ SndioOutput::Open(AudioFormat &audio_format)
 		throw std::runtime_error("Requested audio params cannot be satisfied");
 	}
 
+	sio_setvol(sio_hdl, raw_volume);
```
MaxKellermann: Why modify the volume while opening the output? No other plugin does that.

src/output/plugins/SndioOutputPlugin.cxx
```diff
@@ -44,30 +45,21 @@ static constexpr unsigned MPD_SNDIO_BUFFER_TIME_MS = 250;
...
-	struct sio_hdl *sio_hdl;
-
-public:
-	SndioOutput(const ConfigBlock &block);
-
-	static AudioOutput *Create(EventLoop &,
-				   const ConfigBlock &block) {
-		return new SndioOutput(block);
-	}
+void SndioMixerCallback(void *arg, unsigned int volume) {
```
MaxKellermann: Should be static.

src/output/plugins/SndioOutputPlugin.cxx
```diff
@@ -44,30 +45,21 @@ static constexpr unsigned MPD_SNDIO_BUFFER_TIME_MS = 250;
...
+	return new SndioOutput(block);
+}
 
 SndioOutput::SndioOutput(const ConfigBlock &block)
 	:AudioOutput(0),
 	 device(block.GetBlockValue("device", SIO_DEVANY)),
 	 buffer_time(block.GetBlockValue("buffer_time",
-					 MPD_SNDIO_BUFFER_TIME_MS))
+					 MPD_SNDIO_BUFFER_TIME_MS)),
+	 raw_volume(SIO_MAXVOL)
```
MaxKellermann: Use C++11 initializer.

src/output/plugins/SndioOutputPlugin.cxx
```diff
@@ -163,9 +158,34 @@ SndioOutput::Play(const void *chunk, size_t size)
...
+	raw_volume = _raw_volume;
+	if (listener != NULL && mixer != NULL) {
+		listener->OnMixerVolumeChanged(*mixer,
+		    _raw_volume * 100 / SIO_MAXVOL);
+	}
+}
+
+unsigned int
+SndioOutput::GetVolume() {
+	return (raw_volume * 100 / SIO_MAXVOL);
```
MaxKellermann: Duplicate formula; this code is already present in `VolumeChanged()`. Maybe better remember the MPD-style value instead of the raw value?

src/output/plugins/SndioOutputPlugin.cxx
MaxKellermann: Coding style violation.

src/output/plugins/SndioOutputPlugin.cxx
```diff
@@ -163,9 +158,34 @@ SndioOutput::Play(const void *chunk, size_t size)
...
+SndioOutput::VolumeChanged(unsigned int _raw_volume) {
+	raw_volume = _raw_volume;
+	if (listener != NULL && mixer != NULL) {
+		listener->OnMixerVolumeChanged(*mixer,
+		    _raw_volume * 100 / SIO_MAXVOL);
+	}
+}
+
+unsigned int
+SndioOutput::GetVolume() {
```
MaxKellermann: Another coding style violation.

src/output/plugins/SndioOutputPlugin.hxx
```diff
@@ -20,6 +20,38 @@
...
 
+#include "../OutputAPI.hxx"
+#include "mixer/Listener.hxx"
+
 extern const struct AudioOutputPlugin sndio_output_plugin;
 
+class PulseMixer;
+
+class SndioOutput final : AudioOutput {
+	friend class SndioMixer;
```
MaxKellermann: No `friend`, please, unless ultimately necessary! `friend` can access any private, and is obscure. Provide an API instead of friending other classes.

src/output/plugins/SndioOutputPlugin.hxx
```diff
@@ -20,6 +20,38 @@
...
+	friend void SndioMixerCallback(void *arg, unsigned int volume);
+
+	Mixer *mixer = NULL;
+	MixerListener *listener = NULL;
```
MaxKellermann: `nullptr` ...

src/net/IPv4Address.hxx
madroach: Oh, I'm sorry. I don't know much C++.
I didn't notice this feature was not yet standardized.

Then how would you fix this? Add a feature test macro?
I'd suggest removing the constexpr optimization altogether. This is no critical code path.

src/output/plugins/SndioOutputPlugin.cxx
madroach: Because continuing playback after stop will open a new sndiod stream which will have a new 100% volume assigned to it. Could be a loud surprise.

src/output/plugins/SndioOutputPlugin.cxx
madroach: I thought about this, but honestly don't remember the reason right now. I believe the code would be duplicated elsewhere when storing the mpd value, or rounding errors would appear when a third party changes the volume.

madroach:
Thanks for your review and feedback. Maybe I'll find some time to rework the diff on the weekend.

src/output/plugins/SndioOutputPlugin.cxx
MaxKellermann: So there is nothing inside sndio which remembers the volume? Then this line is ok, but add a code comment explaining it.

src/output/plugins/SndioOutputPlugin.cxx
MaxKellermann: OK, move the formula to a small function. Don't duplicate code. This is maintenance hell.

madroach:
I reworked the diff to address the requested changes.

MaxKellermann:
/home/max/git/mpd/src/mixer/plugins/SndioMixerPlugin.cxx: In constructor 'SndioMixer::SndioMixer(SndioOutput&, MixerListener&)':
/home/max/git/mpd/src/mixer/plugins/SndioMixerPlugin.cxx:31:3: error: declaration of 'listener' shadows a member of 'SndioMixer' [-Werror=shadow]
   :Mixer(sndio_mixer_plugin, listener), output(output)
   ^
In file included from /home/max/git/mpd/src/mixer/plugins/SndioMixerPlugin.cxx:20:0:
/home/max/git/mpd/src/mixer/MixerInternal.hxx:34:17: note: shadowed declaration is here
  MixerListener &listener;
                 ^~~~~~~~
/home/max/git/mpd/src/mixer/plugins/SndioMixerPlugin.cxx:31:3: error: declaration of 'output' shadows a member of 'SndioMixer' [-Werror=shadow]
   :Mixer(sndio_mixer_plugin, listener), output(output)
   ^
/home/max/git/mpd/src/mixer/plugins/SndioMixerPlugin.cxx:27:15: note: shadowed declaration is here
  SndioOutput &output;
               ^~~~~~

MaxKellermann:
I've merged the first two patches, thanks for that so far.

madroach:
renamed the shadowing parameter to ``_listener``.
clang doesn't know ``-Wshadow``. That's why I didn't get this error.

MaxKellermann:
The other shadow remains:

/home/max/git/mpd/src/mixer/plugins/SndioMixerPlugin.cxx: In constructor 'SndioMixer::SndioMixer(SndioOutput&, MixerListener&)':
/home/max/git/mpd/src/mixer/plugins/SndioMixerPlugin.cxx:31:3: error: declaration of 'output' shadows a member of 'SndioMixer' [-Werror=shadow]
   :Mixer(sndio_mixer_plugin, _listener), output(output)
   ^
/home/max/git/mpd/src/mixer/plugins/SndioMixerPlugin.cxx:27:15: note: shadowed declaration is here

madroach:
Thanks for your review and merging!

Issue #110, opened 2017-09-15
 Incorrect duration(length) on Matroska files


ChaoticConundrum:
MPD seems to have no issue playing and reading tags from Matroska files (.mka, .mkv), but the track length is always incorrect. Some clients display it as 0:00, and some 23:59:59. 

I am observing this on Arch Linux, with .mka files mixed along with other files in a large library. No other formats I have used exhibit this, and I have never seen a Matroska length displayed correctly. These Matroska files are encoded with ffmpeg, typically with Opus or Vorbis audio streams.

Other tools can correctly read the duration field. For example, `mkvinfo` shows the duration for one file as `Duration: 169.321s (00:02:49.321).`

I suspect because EBML fields are structured differently from other formats (e.g. global properties, track properties) the abstraction for tag parsing somewhere in MPD is not correctly identifying the Duration field, and simply leaving the value unset.

MaxKellermann:
I tried a few MKV files, and for all of these, MPD was unable to obtain a duration, because this is what FFmpeg says (FFmpeg is the only MPD plugin which can parse MKV). MPD communicates this problem to clients by saying "0". So if a client shows 0:00, it is just misinterpreting this "magic value" - not a MPD bug.
All clients which show "23:59:59" are severely bugged. This is also not a MPD bug.
Being unable to determine MKV duration is also not a MPD bug, but a missing feature in FFmpeg. If you wish to have that feature, ask the FFmpeg developers for it.

ChaoticConundrum:
Ok, so what I'm seeing is that before `FromFfmpegTimeChecked()` is called in `FfmpegDecode()`, `av_stream.duration` (populated by `avformat_open_input()`?) is `AV_NOPTS_VALUE` (0x8000000000000000), for which `FromFfmpegTimeChecked()` returns `SignedSongTime::Negative()`.

That -1 is passed to `client.Ready()`, and as far as I can see, nothing else is done to determine the duration.

PR #112, opened 2017-09-18
FFmpeg plugin: Read correct durations from some files that do not have per-stream durations.

ChaoticConundrum:
See issue #110. This appears to solve that problem for Matroska files.


MaxKellermann:
b253a6b71e4aaecfe8f2326b6e5b1c32d93fc83c

Cherry-picked to v0.20.x and fixed various coding style issues.

Issue #115, opened 2017-09-21
[Feature Request] GME Plugin: load sidecar M3U files


jprjr:
A lot of chiptunes are distributed with an m3u file that contains extra metadata like song titles, track length, etc.

It looks like libgme has a function for parsing these m3u files: `gme_load_m3u( Music_Emu*, const char path [] );`, it's meant to be called after `gme_open_file`

It would be great if the GME plugin could check for the existence of an m3u file in the same directory as the chiptune file and call `gme_load_m3u` if it exists. It could be limited to m3u files with a matching name, ie, `chiptune.gbs`, when loaded, would check for `chiptune.m3u`

jprjr:
Closed with #116

PR #116, opened 2017-09-22
GME Plugin: try loading m3u sidecar files

jprjr:
This is for issue #115  I'll be honest, my C++ skills are pretty lacking, I hope this is up-to-snuff! I saw there were some Path classes in the MPD codebase, but I wasn't sure how to properly use them, so I'm just using `std::string`.

The `gme_load_m3u` function seems to be really flexible/tolerant with what you throw at it, I tried loading several m3u files, cat-ing /dev/urandom into m3u files (some a few bytes, some a few hundred megabytes), and missing m3u files. GME will report an error, but otherwise continue loading/playing songs.

I wasn't sure if MPD has any classes/mechanisms for finding a file while ignoring case, or if that's not a particularly large concern?

### DIFF
--- a/src/decoder/plugins/GmeDecoderPlugin.cxx
+++ b/src/decoder/plugins/GmeDecoderPlugin.cxx
@@ -105,16 +105,39 @@ ParseContainerPath(Path path_fs)
 	return { path_fs.GetDirectoryName(), track - 1 };
 }
 
+static Music_Emu*
+LoadGmeAndM3u(GmeContainerPath container) {
+
+	const char *path = container.path.c_str();
+	const char *suffix = uri_get_suffix(path);
+
+	Music_Emu *emu;
+	const char *gme_err =
+		gme_open_file(path, &emu, GME_SAMPLE_RATE);
+	if (gme_err != nullptr) {
+		LogWarning(gme_domain, gme_err);
+		return nullptr;
+	}
+
+	if(suffix == nullptr) {
+		return emu;
+	}
+
+	std::string m3u_path(path,suffix);
+	m3u_path += "m3u";
+
+	gme_load_m3u(emu,m3u_path.c_str());
+	return emu;
+}
+
+
 static void
 gme_file_decode(DecoderClient &client, Path path_fs)
 {
 	const auto container = ParseContainerPath(path_fs);
 
-	Music_Emu *emu;
-	const char *gme_err =
-		gme_open_file(container.path.c_str(), &emu, GME_SAMPLE_RATE);
-	if (gme_err != nullptr) {
-		LogWarning(gme_domain, gme_err);
+	Music_Emu *emu = LoadGmeAndM3u(container);
+	if(emu == nullptr) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


src/decoder/plugins/GmeDecoderPlugin.cxx
```diff
@@ -105,16 +105,43 @@ ParseContainerPath(Path path_fs)
 	return { path_fs.GetDirectoryName(), track - 1 };
 }
 
-static void
-gme_file_decode(DecoderClient &client, Path path_fs)
-{
-	const auto container = ParseContainerPath(path_fs);
+static Music_Emu
```
MaxKellermann: Coding style violation: the `*` belongs after the `Music_Emu`. This is unreadable!

src/decoder/plugins/GmeDecoderPlugin.cxx
```diff
@@ -105,16 +105,43 @@ ParseContainerPath(Path path_fs)
...
 	if (gme_err != nullptr) {
 		LogWarning(gme_domain, gme_err);
+		return nullptr;
+	}
+
+	if(suffix == nullptr) {
+		return emu;
+	}
+
+	std::string m3u_path(path,suffix - path);
```
MaxKellermann: Just `std::string m3u_path(path, suffix)` does the same.

src/decoder/plugins/GmeDecoderPlugin.cxx
```diff
@@ -105,16 +105,43 @@ ParseContainerPath(Path path_fs)
...
+	m3u_path += "m3u";
```
MaxKellermann: Does this work? I ask because this looks like it removes the dot, but never readds it.

src/decoder/plugins/GmeDecoderPlugin.cxx
```diff
@@ -105,16 +105,43 @@ ParseContainerPath(Path path_fs)
...
+
+
+	gme_err = gme_load_m3u(emu,m3u_path.c_str());
+	if(gme_err != nullptr) {
+		LogWarning(gme_domain,gme_err);
```
MaxKellermann: This is not worth a warning. Such a warning in a normal situation (there's no m3u file!) would confuse users.

src/decoder/plugins/GmeDecoderPlugin.cxx
jprjr: It seems to, `suffix` is just `gbs`, `nsf` (it doesn't start at the dot)

jprjr:
I've made corrections - do you prefer for this branch to be squashed into a single commit, or to keep the changes as a separate commit?

src/decoder/plugins/GmeDecoderPlugin.cxx
MaxKellermann: True, I forgot that `GetSuffix()` returns a pointer to after the dot.

MaxKellermann:
Until it's merged, corrections squashed please.

PR #117, opened 2017-09-23
GME Plugin: fix track numbering, only load m3u when it exists

jprjr:
GME starts all track indexes at zero, but the subtune prefixes start at one. This caused an off-by-one error with track titles.

It also turns out, if you load an NSFE file (which has embedded track titles), then attempt to load an M3U file, it causes GME to lose all information found in the NSFE file. This adds a check that the M3U file exists before attempting to load.


MaxKellermann:
Don't mix two independent changes in one commit. And if your pull request needs additional text, it is clearly a sign that your commit message is bad.
The FileExists change deserves a code comment explaining why this needs to be done.

MaxKellermann:
Cherry-picked the fix to v0.20.x: ea80587ddb1905851623200b4cf5ef260f8c9631

PR #137, opened 2017-10-28
decoder/vorbis: scale and clip tremor-decoded samples to 15 bits

marcin1j:
Tremor decoder is unusable since commit 2ee43c4. Sound is distorted to
the point where it's nothing but noise.

The data from vorbis_synthesis_pcmout() needs to be scaled and
clipped for 16 bit sample size. For reference see
http://lists.xiph.org/pipermail/tremor/2010-April/001642.html and
http://lists.xiph.org/pipermail/vorbis/2006-October/026513.html.


MaxKellermann:
Cherry-picked to v0.20.x: 3b23cf0258e4225d094e10d16fcdc95b10d49d84

Issue #139, opened 2017-10-31
hang on certain FLAC files w/ http/opus output & soxr resampler


cathugger:
Problem: mpd hangs (with ~163% CPU usage) when attempting to play some of my FLAC files.
Problem happens when mpd automatically switches to such file, it stays at 0% and wastes CPU time.
Manually trying to play track/seek usually helps, and track starts playing.
Used gmpc as client, but shouldn't matter.

testcase1:
mpd version: 0.20.11
system: archlinux x86_64
testcase2:
mpd version: 0.20.9
system: debian aarch64

I packed relevant files and configs for testing:
https://a.pomf.cat/xuykyd.tar.xz - whole testing folder with 1 flac which doesn't cause problem and 1 which does
https://gist.github.com/cathugger/1074b3c1f530b47fdfb4fff942184a32 - mpd.conf alone
https://gist.github.com/cathugger/5c5fb739e21e2027cc28371ed949f126 - mpd.log alone (doesn't show anything significant imo)

Steps to reproduce:
download testing folder, unpack (will create folder named `ttt`), adjust paths in ttt/mpd.conf, run `mpd --no-daemon ttt/mpd.conf`.
connect gmpc (or other client of choice), enable repeat (if not already enabled), add both tracks (if not already added),
play `03. Violet Emerald - I Really Don't Love You.flac`, it's supposed to play fine.
connect some client to http port.
`03. Violet Emerald - I Really Don't Love You.flac` is supposed to play fine so just seek to the end and wait till mpd automatically switches to other track.
when trying to play `white ghosts - far away from here - 03 livin' dead.flac` it will hang.

I was able to reproduce this both on my espressobin (debian aarch64) and laptop (archlinux amd64).

cathugger:
I tried doing some more changing things around and noticed that when opus output is disabled, it doesn't hang up. So it's probably related either to http output or opus encoder.

cathugger:
main difference I see between 2 files in ffprobe is that one which works well uses s16 (16 bit) while one which fails uses s32 (24 bit).

cathugger:
this is where it's spending most of its time:
![2017-11-01-150334_693x362_scrot](https://user-images.githubusercontent.com/26472432/32276000-4d88e4ce-bf05-11e7-817a-6f18b4c8e274.png)

cathugger:
[...]

PR #140, opened 2017-11-01
tolerate empty data packets from resampler. #139

cathugger:


cathugger:
bad commit user, delet

PR #141, opened 2017-11-01
tolerate empty data packets from resampler. #139

cathugger:


MaxKellermann:
Thanks for debugging and figuring this all out, that's excellent work, BUT:
This PR is an inferior approach at solving the problem, and the patch doesn't even do what the description says. Because MPD already "tolerates empty packets", and your patch does nothing to change this.
What it actually does is flush unused tags even in the presence of empty data. This bug only occurs with the `tags "no"` configuration, where `AudioOutputSource::ReadTag()` never gets called, and thus `pending_tag` never gets cleared, and therefore `AudioOutputSource::Fill()` never calls `pipe.Consume()`.
But the way it flushes those unused tags isn't good. It calls `AudioOutputSource::ConsumeData(0)`, saying "flush no data at all", and an undocumented side effect of that method is that it marks an empty chunk consumed. This is fragile and not obvious.
It would be better to explicitly flush the tag by always calling `ReadTag()`.

cathugger:
I though AudioOutputSource::ConsumeData() was main place where current_chunk is supposed to be null'd but I overlooked what AudioOutputSource::Fill() does, sorry.
In either case it's relying on behavior which isn't obvious, but your proposed method is more correct. So I'll use that and leave comment.

MaxKellermann:
Folded & cherry-picked to v0.20.x

PR #159, opened 2017-11-26
Problem with GCC 7.2 and mounts

FlashSystems:
GCC 7.2 does not generate a correct assignment for mnt->mounted_database = db. I'm not sure why but using a function fixes this.

Without this function the mounted_database property is not set and the mountpoint does not work as expected.

### DIFF
--- a/src/db/plugins/simple/Directory.cxx
+++ b/src/db/plugins/simple/Directory.cxx
@@ -85,17 +85,6 @@ Directory::CreateChild(const char *name_utf8)
 	return child;
 }
 
-Directory*
-Directory::CreateChild(const char *name_utf8, Database* db)
-{
-	assert(db != nullptr);
-
-	Directory *child = CreateChild(name_utf8);
-	child->mounted_database = db;
-
-	return child;
-}
-
 const Directory *
 Directory::FindChild(const char *name) const noexcept
 {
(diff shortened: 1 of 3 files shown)


MaxKellermann:
This looks fishy. If you don't know why your change works, then it's not a good change.
Does removing the `gcc_malloc` fix the issue as well?

FlashSystems:
Removing `gcc_malloc` from `Directory::CreateChild` did the trick. I was not aware that the attribute of the function returning the pointer could be the problem here. Thank you for pointing this out.

MaxKellermann:
Neither did I, that was just a guess. This means that this is a GCC bug.

FlashSystems:
Hm, than adding
```
#if GCC_CHECK_VERSION(7,2)
#  define gcc_malloc __attribute__((malloc))
#else
#  define gcc_malloc
#endif
```
to `Compiler.hxx` would be a better solution. Maybe there are more problems lingering...

MaxKellermann:
12085038881a4aba94c74b7104c5090bc777001e
Indeed this `malloc` attribute was wrong and that's not a GCC bug, because aliasing can also mean that somewhere else there lurks a pointer to the same object, even if it's not directly reachable.
I checked the other `malloc` attributes left in the MPD sources, and the others appear to be correct.

PR #160, opened 2017-11-26
Save and restore mountpoints

FlashSystems:
The mount command is really useful to add smb or nfs shares. It's a pitty the mounts are not saved within the state file and restored on restart. For me this means most of my music collection is gone until I remount the shares.
This pull request stores the mounts within the state file and restores them on startup.

Maybe it would be a nice addition to extend the mount command to have a "temporary" switch that allow mounts to be marked as "do not save".

### DIFF
--- a/src/StateFile.cxx
+++ b/src/StateFile.cxx
@@ -24,6 +24,7 @@
 #include "fs/io/TextFile.hxx"
 #include "fs/io/FileOutputStream.hxx"
 #include "fs/io/BufferedOutputStream.hxx"
+#include "storage/StorageState.hxx"
 #include "Partition.hxx"
 #include "Instance.hxx"
 #include "mixer/Volume.hxx"
@@ -56,6 +57,7 @@ StateFile::RememberVersions() noexcept
 	prev_output_version = audio_output_state_get_version();
 	prev_playlist_version = playlist_state_get_hash(partition.playlist,
 							partition.pc);
+	prev_storage_version = storage_state_get_hash(partition.instance);
 }
 
 bool
@@ -64,7 +66,8 @@ StateFile::IsModified() const noexcept
 	return prev_volume_version != sw_volume_state_get_hash() ||
 		prev_output_version != audio_output_state_get_version() ||
 		prev_playlist_version != playlist_state_get_hash(partition.playlist,
-								 partition.pc);
+								 partition.pc) ||
+		prev_storage_version != storage_state_get_hash(partition.instance);
 }
 
 inline void
@@ -72,6 +75,11 @@ StateFile::Write(BufferedOutputStream &os)
 {
 	save_sw_volume_state(os);
 	audio_output_state_save(os, partition.outputs);
+
+#ifdef ENABLE_DATABASE
+	storage_state_save(os, partition.instance);
+#endif
+
 	playlist_state_save(os, partition.playlist, partition.pc);
 }
 
@@ -123,6 +131,10 @@ try {
 			playlist_state_restore(line, file, song_loader,
 					       partition.playlist,
 					       partition.pc);
+#ifdef ENABLE_DATABASE
+		success = success || storage_state_restore(line, file, partition.instance);
+#endif
+
 		if (!success)
 			FormatError(state_file_domain,
 				    "Unrecognized line in state file: %s",
--- a/src/storage/StorageState.cxx
+++ b/src/storage/StorageState.cxx
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2003-2017 The Music Player Daemon Project
+ * http://www.musicpd.org
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+/*
+ * Save and load mounts of the compound storage to/from the state file.
+ *
+ */
+
+#include "config.h"
+#include "StorageState.hxx"
+#include "fs/io/TextFile.hxx"
+#include "fs/io/BufferedOutputStream.hxx"
+#include "storage/Registry.hxx"
+#include "storage/CompositeStorage.hxx"
+#include "db/plugins/simple/SimpleDatabasePlugin.hxx"
+#include "util/StringCompare.hxx"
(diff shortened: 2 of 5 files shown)


src/StateFile.cxx
```diff
@@ -118,11 +124,16 @@ try {
 
 	const char *line;
 	while ((line = file.ReadLine()) != nullptr) {
-		success = read_sw_volume_state(line, partition.outputs) ||
-			audio_output_state_read(line, partition.outputs) ||
-			playlist_state_restore(line, file, song_loader,
+		success =            read_sw_volume_state(line, partition.outputs);
+		success = success || audio_output_state_read(line, partition.outputs);
+		success = success || playlist_state_restore(line, file, song_loader,
```
MaxKellermann: Why did you refactor that, and hid it in this commit?

src/StateFile.cxx
```diff
@@ -118,11 +124,16 @@ try {
...
 					       partition.playlist,
 					       partition.pc);
+#ifdef ENABLE_DATABASE
+		success = success || ((CompositeStorage *)partition.instance.storage)->MountStateRestore(
```
MaxKellermann: This needs a short-circuiting `||` operator.

src/storage/CompositeStorage.cxx
```diff
@@ -21,13 +21,23 @@
 #include "CompositeStorage.hxx"
 #include "FileInfo.hxx"
 #include "fs/AllocatedPath.hxx"
+#include "fs/io/BufferedOutputStream.hxx"
+#include "storage/Registry.hxx"
```
MaxKellermann: I'd like to keep this out of this class, because the way you implemented it, the generic `CompositeStorage` class now has a dependency on all storage plugins. `CompositeStorage` should stay generic and independent.

src/storage/CompositeStorage.cxx
```diff
@@ -232,6 +242,63 @@ CompositeStorage::Unmount(const char *uri)
...
+		try {
+			((SimpleDatabase *)db)->Mount(uri.c_str(), url.c_str());
+		} catch (...) {
+			throw;
+		}
+	}
+#endif
+
+	Mount(uri.c_str(), storage);
+	//instance.EmitIdle(IDLE_MOUNT);
```
MaxKellermann: What does this comment mean?

src/StateFile.cxx
FlashSystems: I refactored it to make the calls to the different save functions consistent. If you prefer I can keep the original calls.

src/StateFile.cxx
FlashSystems: There is a short-circuiting `||` operator. Am I missing something?

src/storage/CompositeStorage.cxx
FlashSystems: I see. But to restore the mountpoints this dependency is neccessary. Having a `onRestoreMount` callback to move it out of this class only moves the dependency into `StateFile.cxx` which is even worse I think. I don't see any better solution right now. Could you give me a hint where to put it?

src/StateFile.cxx
MaxKellermann: Sorry, you're right.

src/StateFile.cxx
MaxKellermann: I see no added consistency here.

src/storage/CompositeStorage.cxx
MaxKellermann: Why is that worse? `StateFile.cxx` is glue code with a shitload of dependencies - to everything which can be referenced in the state file. That's a natural dependency. But I wouldn't have it exactly in `StateFile.cxx`, rather in a separate source like `StorageState.cxx` or so. It gets called from `StateFile.cxx` and calls into a `CompositeStorage` instance to save and restore its state.

FlashSystems:
I hope that my refactored code fits better into MPD.

MaxKellermann:
64d141f71e690a4258ba9ee8712140bc9d961883

MaxKellermann:
Note that your code introduced a major crash bug when used with database plugins other than "simple". I'm currently trying to repair the damage.

src/storage/StorageState.cxx
```diff
@@ -0,0 +1,137 @@
...
+	FormatDebug(storage_domain, "Restoring mount %s => %s", uri.c_str(), url.c_str());
+
+	auto &event_loop = instance.io_thread.GetEventLoop();
+	Storage *storage = CreateStorageURI(event_loop, url.c_str());
+	if (storage == nullptr) {
+		FormatError(storage_domain, "Unrecognized storage URI: %s", url.c_str());
+		return true;
+	}
+
+#ifdef ENABLE_DATABASE
```
MaxKellermann: Unnecessary `#ifdef`

src/storage/StorageState.cxx
```diff
@@ -0,0 +1,137 @@
...
+
+	if (url.empty() || uri.empty()) {
+		LogError(storage_domain, "Missing value in mountpoint state.");	
+		return true;
+	}
+
+	FormatDebug(storage_domain, "Restoring mount %s => %s", uri.c_str(), url.c_str());
+
+	auto &event_loop = instance.io_thread.GetEventLoop();
+	Storage *storage = CreateStorageURI(event_loop, url.c_str());
```
MaxKellermann: Memory leak

src/storage/StorageState.cxx
```diff
@@ -0,0 +1,137 @@
...
+			return;
+
+		os.Format(
+			MOUNT_STATE_BEGIN "\n"
+			MOUNT_STATE_STORAGE_URI "%s\n"
+			MOUNT_STATE_MOUNTED_URL "%s\n"
+			MOUNT_STATE_END "\n", mount_uri, uri.c_str());
+	};
+
+	((CompositeStorage*)instance.storage)->VisitMounts(visitor);
```
MaxKellermann: This cast crashes if this is not a `CompositeStorage`

src/storage/StorageState.cxx
```diff
@@ -0,0 +1,137 @@
...
+	if (db != nullptr && db->IsPlugin(simple_db_plugin)) {
+		try {
+			((SimpleDatabase *)db)->Mount(uri.c_str(), url.c_str());
+		} catch (...) {
+			throw;
+		}
+	}
+#endif
+
+	((CompositeStorage*)instance.storage)->Mount(uri.c_str(), storage);
```
MaxKellermann: This cast crashes as well if this is not a `CompositeStorage`

src/storage/StorageState.cxx
```diff
@@ -0,0 +1,137 @@
...
+		return true;
+	}
+
+#ifdef ENABLE_DATABASE
+	Database *db = instance.database;
+	if (db != nullptr && db->IsPlugin(simple_db_plugin)) {
+		try {
+			((SimpleDatabase *)db)->Mount(uri.c_str(), url.c_str());
+		} catch (...) {
+			throw;
```
MaxKellermann: Useless `try`/`catch`/`throw`, but no error handling. What's supposed to happen if there is an exception? How will the user be able to recover/repair this?

src/storage/StorageState.cxx
FlashSystems: Sorry for breaking MPD and causing you extra work. I didn't realize that there isn't always a composite storage instance available.

PR #161, opened 2017-11-26
Mount-Points are purged from database on update/rescan

FlashSystems:
Mount-Points are removed from the database on update/rescan because they are detected as empty. This pull request keeps the mount points intact on rescan of the base database.


MaxKellermann:
Cherry-picked to v0.20.x: 63fc98591d9be2ee4f1ef11965f7eb0717689d68

PR #166, opened 2017-12-03
Correctly set the mixer plugin for osx output

Wang-Yue:


MaxKellermann:
7ba7ce3af7f013de10f458540e2cd9e7ff3c212f (Your commit had an indent bug)

PR #167, opened 2017-12-03
OSXOutputPlugin: Make the ring buffer in the larger. 

Wang-Yue:
This would allow MPD having sufficient space to write on.

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -686,7 +686,9 @@ OSXOutput::Open(AudioFormat &audio_format)
 					 errormsg);
 	}
 
-	ring_buffer = new boost::lockfree::spsc_queue<uint8_t>(buffer_frame_size);
+	// Make the ring buffer 4 times as large as the hardware buffer size.
+	// This would allow MPD having sufficient space to write on.
+	ring_buffer = new boost::lockfree::spsc_queue<uint8_t>(buffer_frame_size * 4);
 
 	status = AudioOutputUnitStart(au);
 	if (status != 0) {


MaxKellermann:
Why was the previous buffer insufficient?
Is that a problem only with certain chips which have a very small hardware ring buffer?
Larger buffers come at a price - they increase latency. Such a change should be explained/justified.
If this is only about certain hardware buffer sizes, then maybe we need a different metric; for example, your formula quadruples the buffer size even for chips which have a very large ring buffer already, with already more than enough buffer time. That may solve a problem for some users, but gives disadvantages to others who never had a problem.

MaxKellermann:
Would this be a candidate for the stable branch?

Wang-Yue:
You are right.

For certain format (hi-res files) and normal buffer size hardware, The hardware may at once consume most of the buffers. However, since in Delay() function, MPD is supposed to wait for 25 ms after the next try. Then it will create a hiccup. The negative impact is much major than increasing the latency.

I understand larger buffers come at a price. That's why in my earlier commit I significantly reduce the However, the buffer size in CoreAudio is set according to the hardware, which is super small latency. For instance, the system audio of 2015 generation of macbook pro has buffer size of 2^15 samples, which is just 0.07s for 44.1k framerate. compare to the 0.5 sec latency alsa plugin has, even if we quadruple it, it's still super tiny. 

In my test, double it already works fine. Do you mind if I change the ratio to 2?

Wang-Yue:
You are right.

For certain format (hi-res files) and normal buffer size hardware, The hardware may at once consume most of the buffers. However, since in Delay() function, MPD is supposed to wait for 25 ms after the next try. Then it will create a hiccup. The negative impact is much major than increasing the latency.

I understand larger buffers come at a price. That's why in my earlier commit last year I significantly reduced it. However, the buffer size in CoreAudio is set according to the hardware, which is super small latency. For instance, the system audio of 2015 generation of macbook pro has maximum buffer size of 4096 samples, which is just 0.09s for 44.1k framerate, or 0.04 for 96k frames --- . compare to the 0.5 sec latency alsa plugin has, even if we quadruple it, it's still super tiny. 

In my test, doubling it already works fine. Do you mind if I change the ratio to 2? If you have better options please let me know

MaxKellermann:
That `Delay()` implementation is bad; there should not be a fixed-length delay. The ALSA plugin uses another approach: it waits on a condition variable. The downside there is that the output thread now blocks in `Play()` and cannot respond to commands.
So if your problem is that the ring buffer is not large enough to feed these 25ms, then your patch still doesn't guarantee that. Let's say your current hardware has a buffer large enough for 15ms, and that will obviously xrun within these 25ms; so your approach is to double (30ms) or quadruple it (60ms). That will fix the xruns.
But the next guy buys a chip with just 5ms of hardware buffer. Not even quadrupling it will solve the xrun.
Your approach is conceptually flawed. You say "let's make it somewhat larger", but that "somewhat" doesn't even try to get an understanding of the problem.
If you want to get over these 25ms, choose a buffer size that covers at least those 25ms instead of "four times the hardware buffer size". Or do away with this fixed-length delay and do something more clever. Like a condition variable.

Wang-Yue:
Again, you are right. 

Technically the delay should be set according to the hardware buffer size calculated (I set it to be the time of 1024 frames, which is 1/4 of the maximum size). However, even if I did that, there're still hiccups. I assume there're other delays in MPD, which is larger than the calculated delay, which cause the output to delay further. I don't have time to look where the issue is. 

I could however, follow alsa's implementation, and just use 0.1s buffer time. Here's the patch:

```
diff --git a/src/output/plugins/OSXOutputPlugin.cxx b/src/output/plugins/OSXOutputPlugin.cxx
index e6b039b52..deadab271 100644
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -36,6 +36,8 @@
 
 #include <memory>
 
+static constexpr unsigned MPD_OSX_BUFFER_TIME_MS = 100;
+
 struct OSXOutput final : AudioOutput {
        /* configuration settings */
        OSType component_subtype;
@@ -686,7 +688,8 @@ OSXOutput::Open(AudioFormat &audio_format)
                                         errormsg);
        }
 
-       ring_buffer = new boost::lockfree::spsc_queue<uint8_t>(buffer_frame_size);
+       size_t ring_buffer_size = MPD_OSX_BUFFER_TIME_MS * audio_format.GetFrameSize() * audio_format.sample_rate / 1000;
+       ring_buffer = new boost::lockfree::spsc_queue<uint8_t>(ring_buffer_size);
 
        status = AudioOutputUnitStart(au);
        if (status != 0) {
@@ -708,7 +711,7 @@ OSXOutput::Delay() const noexcept
 {
        return ring_buffer->write_available()
                ? std::chrono::steady_clock::duration::zero()
-               : std::chrono::milliseconds(25);
+               : std::chrono::milliseconds(MPD_OSX_BUFFER_TIME_MS / 4);
 }
 
 int
```

You can accept either patch. I'm fine with either approach.  But I don't like the idea  --- it  should be set according to the hardware buffer size to maximize the performance. 

FYI, cmus's CoreAudio plugin, which share the identical logic (since MPD's OS X plugin implementation was just ported from there by me), doesn't need the buffer size to be quadrupled or doubled, as there's no other significant delays.

Wang-Yue:
Alternative patch sent as #168 168

PR #168, opened 2017-12-04
OSXOutputPlugin: set the buffer time to be 100ms

Wang-Yue:


MaxKellermann:
Please add an explanation to the commit message. Always consider that people who havn't participated in our discussion should be able to understand what's going on.

Wang-Yue:
I think the commit message "OSXOutputPlugin: set the buffer time to be 100ms" is clear enough.

MaxKellermann:
And I think the guy who wrote code is always the worst person to judge whether his one-sentence non-explaining commit message is enough.

MaxKellermann:
Sigh. I fixed your commit and cherry-picked it to v0.20.x: f1ef9f9d31c71857db021029326715028a217906

PR #170, opened 2017-12-07
lib/upnp: use include path without upnp/ prefix and honor pkg-config …

ukleinek:
…CFLAGS

If libupnp is installed in a non-standard location we must rely on the
include path provided by $(pkg-config --cflags libupnp). Relative to the
path given from that command no prefix must be used to find the respective
files.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -792,6 +792,7 @@ libneighbor_a_SOURCES += \
 NEIGHBOR_LIBS += \
 	$(EXPAT_LIBS) \
 	$(UPNP_LIBS)
+libneighbor_a_CXXFLAGS = $(UPNP_CFLAGS)
 endif
 
 endif
@@ -843,6 +844,7 @@ libdb_plugins_a_SOURCES += \
 DB_LIBS += \
 	$(EXPAT_LIBS) \
 	$(UPNP_LIBS)
+libdb_plugins_a_CXXFLAGS = $(UPNP_CFLAGS)
 endif
 
 endif
--- a/src/lib/upnp/Action.hxx
+++ b/src/lib/upnp/Action.hxx
@@ -22,7 +22,7 @@
 
 #include "Compiler.h"
 
-#include <upnp/upnptools.h>
+#include <upnptools.h>
 
 static inline constexpr unsigned
 CountNameValuePairs() noexcept
--- a/src/lib/upnp/Callback.hxx
+++ b/src/lib/upnp/Callback.hxx
@@ -20,7 +20,7 @@
 #ifndef MPD_UPNP_CALLBACK_HXX
 #define MPD_UPNP_CALLBACK_HXX
 
-#include <upnp/upnp.h>
+#include <upnp.h>
 
 /**
  * A class that is supposed to be used for libupnp asynchronous
--- a/src/lib/upnp/ClientInit.cxx
+++ b/src/lib/upnp/ClientInit.cxx
@@ -24,7 +24,7 @@
 #include "thread/Mutex.hxx"
 #include "util/RuntimeError.hxx"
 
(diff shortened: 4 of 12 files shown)


Makefile.am
```diff
@@ -792,6 +792,7 @@ libneighbor_a_SOURCES += \
 NEIGHBOR_LIBS += \
 	$(EXPAT_LIBS) \
 	$(UPNP_LIBS)
+libneighbor_a_CXXFLAGS = $(UPNP_CFLAGS)
```
MaxKellermann: This must be `_CPPFLAGS` because `UPNP_CFLAGS` contains preprocessor flags, not compiler flags.

ukleinek: The output of `pkg-config --cflags ...` contains both pre-processor and compile flags and so both `CPPFLAGS` and `CXXFLAGS` are not optimal but IMHO `CFLAGS` is less wrong. Also the naming of the pkg-config stuff (`--cflags`, `PKG_CHECK_MODULES` setting `[$1]_CFLAGS` and not `[$1]_CPPFLAGS`) is a hint that CXXFLAGS is the better. In my universe using CFLAGS is also the more common.

But I'm not religious about that and if this is the only concern, I can rework.

MaxKellermann: Oh, right, libupnp abuses this to inject `-pthread` into the using application...

MaxKellermann: Note that `-pthread` is not a compile-time option, but a link-time option. It is wrong to add it to the `--cflags` output. This is a libupnp bug!

ukleinek: I don't know if this is to be considered abuse or not, but looking at 
https://codesearch.debian.net/search?q=Cflags%3A.*-pthread&perpkg=1 and https://codesearch.debian.net/search?q=Libs%3A.*-pthread&perpkg=1 it doesn't look unusual at least. I will ask pkg-config upstream about the two things (-pthread handling and `CFLAGS` vs. `CPPFLAGS`)

MaxKellermann: You don't need to ask pkg-config upstream, because this is not a pkg-config issue. And this is not a question about "CFLAGS vs CPPFLAGS" - because as I said, `-pthread` is not a compiler option at all! If anything, it belongs to `pkg-config --libs`.

MaxKellermann: Clarification to avoid misunderstandings (because reality is more complex than my simplified explanation): yes, `-pthread` be can be useful at compile time and GCC documentation explains this, but **only** if the application to be compiled uses the thread API. For a library like libupnp, this is not applicable. libupnp itself should be compiled with `-pthread`, yes, but applications linking with libupnp do not need that flag.
Note that libupnp does ship with some header files which include `pthread.h` (of which MPD uses none), but these look like internal headers which are not part of the public API. I guess installing these as public headers was another mistake of the libupnp project.

ukleinek: according to gcc docs `-pthread` is a preprocessor option and a linker option. Quoting `info gcc-7`:
> '-pthread'
>     Define additional macros required for using the POSIX threads
>     library.  You should use this option consistently for both
>     compilation and linking.

MaxKellermann: I knew you would cite this, that's why I posted the clarification (at the same time).

ukleinek: thanks for your insights, so for libupnp.pc the right thing to use is:
> Libs: -L${libdir} -lupnp -lthreadutil -lixml -pthread
> Cflags: -I${includedir}/upnp
(i.e. drop `-pthread` from `Cflags` and (slightly unrelated) move it to the end of `Libs` to get it into the right link order), right?

MaxKellermann:
Cherry-picked to v0.20.x and fixed remaining problems: 5a11e03725ce31f0486495c5cd747c33138cd775

PR #171, opened 2017-12-07
Calling visit on mounted databases creates wrong uris

FlashSystems:
If `SimpleDatabase::Visit` is called on a database that contains a mounted directry the URIs of the elements passed to the callbacks are not prefixed by the mountpoint path. This leads to `lsinfo` and `add` not working because they use the wrong URI.

### DIFF
--- a/src/db/plugins/simple/Directory.cxx
+++ b/src/db/plugins/simple/Directory.cxx
@@ -227,7 +227,7 @@ Directory::Walk(bool recursive, const SongFilter *filter,
 		   call will lock it again */
 		const ScopeDatabaseUnlock unlock;
 		WalkMount(GetPath(), *mounted_database,
-			  recursive, filter,
+			  "", recursive, filter,
 			  visit_directory, visit_song,
 			  visit_playlist);
 		return;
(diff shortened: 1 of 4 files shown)


MaxKellermann:
I havn't checked your code yet, but I wonder why your PR contains lots and lots of text, which is missing in your commit message.
A good pull request requires no further explanation at all! The commit messages should be enough!

FlashSystems:
I tried to explain _why_  this change is necessary. I don't think this belongs into the commit message.
The commit message states _what_  this commit fixes/changes and I think the changes are simple enough to need no further explaination.

MaxKellermann:
If your changes need no further explanation, then why did you write further explanation in this PR?

Yes, I do think this belongs in the commit message.
I like that you explained your changes; few people do, and I'm glad for every contributor who takes the time to explain. That helps a lot, and from a global perspective, really saves time.
But if you leave your text only in the GitHub PR, it will be lost, locked in a proprietary database, and nobody will ever see it again.

FlashSystems:
Sure, your project, your rules. I'll shorten the PR a little and modify the commit message this weekend.
I would appreciate it if your could take a look at the changes and tell me if further modifications are necessary.

FlashSystems:
Ok, I've extended my commit message with information from the pull request and shorted the PR message.
While changing the commit message I also removed a superfluous comment in `SimpleDatabasePlugin.cxx`.

MaxKellermann:
c488d3123f33fe28525c2b8b229aaae2f9d4dc06

PR #183, opened 2018-01-05
output/haiku: return when as at least one full frame has been copied

sfanxiang:


MaxKellermann:
via `AudioFormat::GetFrameSize()`, which was passed to method `Open()`.

But looking again at the plugin code written by @mmuman it appears I misunderstood his buffering and synchronization concept (but I think I found one or two bugs in it).

I think your change is correct after all, because it does wait for a new buffer by `acquire_sem(new_buffer)`, which should guarantee that the buffer is not yet full. And I think it needs not take care for partial frames, because the buffer passed to `fill_buffer()` should always align to the frame size on `BSoundPlayer` (doesn't it?). So it *should* be impossible to generate partial frames. @mmuman is my assumption correcT?

@mmuman and please review this PR; I may be wrong, because I've never used semaphores.

mmuman:
I don't remember much but it seems I didn't make any assumption on the size of the incoming data, but probably the code can be simplified if we assume at most one buffer.

mmuman:
@sfanxiang did you actually try your changes? Because currently MPD fails to build here… let's fix this first.

mmuman:
Ok, partial build fix for Haiku with #186 …

mmuman:
Ok, #187 fixes the build.

mmuman:
Now it builds but it asserts:
```
mpd: src/player/Control.hxx:277:void PlayerControl::Wait(): thread.IsInside()
```

mmuman:
Ok, looks like a race condition:
pthread_create() starts the thread, which asserts before thread::handle is assigned.
Either it's a bug in Haiku's pthread, or it's an broad assumption about the order of things happening inside pthread_create, not sure what POSIX says.

mmuman:
http://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_create.html : 

> There is no requirement on the implementation that the ID of the created thread be available before the newly created thread starts executing.

Indeed, POSIX doesn't specify if the ID is set before or after the thread starts, so the asserts are based on the Linux behavior…

mmuman:
Ok it's 5am, I really need to sleep.
@sfanxiang : hrev51732 should soon be available in the repository, then you can update, cherry-pick from my two PRs if they haven't been merged yet, and rebase your patch over.

MaxKellermann:
It's been 3 months. Has everybody lost interest in MPD/Haiku support?

MaxKellermann:
Closing due to lack of feedback after more than 3 months.

mmuman:
Hi, sorry my health dragged me away from keyboard for a while…

PR #205, opened 2018-01-26
opus stream chaining

cathugger:
support for chaining streams to enable changing stream' metadata on the fly.
currently support is opt-in because lots of clients can't handle this properly yet.

### DIFF
--- a/src/encoder/plugins/OpusEncoderPlugin.cxx
+++ b/src/encoder/plugins/OpusEncoderPlugin.cxx
@@ -24,6 +24,7 @@
 #include "config/ConfigError.hxx"
 #include "util/Alloc.hxx"
 #include "system/ByteOrder.hxx"
+#include "util/StringUtil.hxx"
 
 #include <opus.h>
 #include <ogg/ogg.h>
@@ -55,27 +56,30 @@ class OpusEncoder final : public OggEncoder {
 	ogg_int64_t granulepos = 0;
 
 public:
-	OpusEncoder(AudioFormat &_audio_format, ::OpusEncoder *_enc);
+	OpusEncoder(AudioFormat &_audio_format, ::OpusEncoder *_enc, bool _chaining);
 	~OpusEncoder() override;
 
 	/* virtual methods from class Encoder */
 	void End() override;
 	void Write(const void *data, size_t length) override;
 
-	size_t Read(void *dest, size_t length) override;
+	void PreTag() override;
+	void SendTag(const Tag &tag) override;
 
 private:
 	void DoEncode(bool eos);
 	void WriteSilence(unsigned fill_frames);
 
+	void GenerateHeaders(const Tag *tag);
 	void GenerateHead();
-	void GenerateTags();
+	void GenerateTags(const Tag *tag);
 };
 
 class PreparedOpusEncoder final : public PreparedEncoder {
 	opus_int32 bitrate;
 	int complexity;
 	int signal;
+	const bool chaining;
 
 public:
 	PreparedOpusEncoder(const ConfigBlock &block);
@@ -89,6 +93,7 @@ public:
 };
 
 PreparedOpusEncoder::PreparedOpusEncoder(const ConfigBlock &block)
+	:chaining(block.GetBlockValue("opustags", false))
 {
 	const char *value = block.GetBlockValue("bitrate", "auto");
 	if (strcmp(value, "auto") == 0)
@@ -124,8 +129,8 @@ opus_encoder_init(const ConfigBlock &block)
 	return new PreparedOpusEncoder(block);
 }
 
-OpusEncoder::OpusEncoder(AudioFormat &_audio_format, ::OpusEncoder *_enc)
-	:OggEncoder(false),
+OpusEncoder::OpusEncoder(AudioFormat &_audio_format, ::OpusEncoder *_enc, bool _chaining)
+	:OggEncoder(_chaining),
 	 audio_format(_audio_format),
 	 frame_size(_audio_format.GetFrameSize()),
 	 buffer_frames(_audio_format.sample_rate / 50),
@@ -134,6 +139,7 @@ OpusEncoder::OpusEncoder(AudioFormat &_audio_format, ::OpusEncoder *_enc)
 	 enc(_enc)
 {
 	opus_encoder_ctl(enc, OPUS_GET_LOOKAHEAD(&lookahead));
+	GenerateHeaders(nullptr);
 }
 
 Encoder *
@@ -171,7 +177,7 @@ PreparedOpusEncoder::Open(AudioFormat &audio_format)
 	opus_encoder_ctl(enc, OPUS_SET_COMPLEXITY(complexity));
 	opus_encoder_ctl(enc, OPUS_SET_SIGNAL(signal));
 
-	return new OpusEncoder(audio_format, enc);
+	return new OpusEncoder(audio_format, enc, chaining);
 }
 
 OpusEncoder::~OpusEncoder()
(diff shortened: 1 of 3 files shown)


MaxKellermann:
Without looking at the code: if your PR contains text, this means that your commit message is bad.

cathugger:
should I change it?

MaxKellermann:
Please fold the two commits (the second one is just fixup for the first one, even though it's not documented that way) and add a proper commit message.
Your first patch is hard to read because you moved original code around and renamed a method; that appears in the diff, even though nothing (but the method name) was changed. Please move that back so I can read your patch properly.

cathugger:
@MaxKellermann done. it's hard to read because back when I changed it, it was kind of rewrite.
there are some changes on how it inserts tags into stream (old version overrides Read() to inject them, my version pushes them when tags appear because otherwise I'd need to hold tags somewhere waiting for Read() and doing that didn't make much sense to me).
also some unrelated changes like changing mime type from "audio/ogg" to "audio/ogg; codecs=opus" because that made more sense to me (I can get rid of this one if you want, I didn't observed any changes in clients from it).

cathugger:
@MaxKellermann I attempted to make it easier to review by moving functions to be closer to their original locations

src/encoder/plugins/OpusEncoderPlugin.cxx
```diff
@@ -183,24 +190,25 @@ OpusEncoder::~OpusEncoder()
 void
 OpusEncoder::DoEncode(bool eos)
 {
-	assert(buffer_position == buffer_size);
+	if (!eos)
+		assert(buffer_position == buffer_size);
```
MaxKellermann: You shouldn't call a macro after an `if` statement without curly braces, because this is unsafe; and it doesn't make sense to check the `eos` flag in the non-debug build. It's better to move that check into the `assert()`.

src/encoder/plugins/OpusEncoderPlugin.cxx
```diff
@@ -84,7 +88,7 @@ class PreparedOpusEncoder final : public PreparedEncoder {
 	Encoder *Open(AudioFormat &audio_format) override;
 
 	const char *GetMimeType() const override {
-		return "audio/ogg";
+		return "audio/ogg; codecs=opus";
```
MaxKellermann: This change has nothing to do with chaining feature, does it? Should be a separate commit; I can cherry-pick such a tiny commit easily, even if there are issues left with the "large" commit.

src/encoder/plugins/OpusEncoderPlugin.cxx
```diff
@@ -52,30 +53,33 @@ class OpusEncoder final : public OggEncoder {
...
 	void GenerateHead();
-	void GenerateTags();
+	void GenerateTags(const Tag *tag);
 };
 
 class PreparedOpusEncoder final : public PreparedEncoder {
 	opus_int32 bitrate;
 	int complexity;
 	int signal;
+	bool chaining;
```
MaxKellermann: Make this `const` and use a C++ initializer in the constructor.

src/encoder/plugins/OpusEncoderPlugin.cxx
```diff
@@ -183,24 +190,25 @@ OpusEncoder::~OpusEncoder()
...
-				    sizeof(buffer2));
+		                    (const float *)buffer,
+		                    buffer_frames,
+		                    buffer2,
+		                    sizeof(buffer2));
 	if (result < 0)
 		throw std::runtime_error("Opus encoder error");
 
-	granulepos += buffer_frames;
+	granulepos += buffer_position / frame_size;
```
MaxKellermann: What is this change about?

src/encoder/plugins/OpusEncoderPlugin.cxx
```diff
@@ -52,30 +53,33 @@ class OpusEncoder final : public OggEncoder {
 
 	ogg_int64_t packetno = 0;
 
-	ogg_int64_t granulepos;
+	ogg_int64_t granulepos = 0;
```
MaxKellermann: Why does this need to be initialized now? Was this a bug previously, or did your other code changes require it?

MaxKellermann:
Thanks, much better now. And each time before you push, please rebase on my master branch.

src/encoder/plugins/OpusEncoderPlugin.cxx
cathugger: will do

src/encoder/plugins/OpusEncoderPlugin.cxx
cathugger: it has nothing to do with chaining feature, yes

src/encoder/plugins/OpusEncoderPlugin.cxx
cathugger: on final stream packet (one with eos set), we are allowed to have granulepos less than packet actually contains. decoder should take smaller granulepos into account and discard unused data. this is for stream cutting - we need to push final packet to inject tags, but we don't have enough data for whole packet. therefore we encode full packet, but specify lesser granulepos.

src/encoder/plugins/OpusEncoderPlugin.cxx
cathugger: >Was this a bug previously
probably. it's not initialised anywhere else. technically granulepos in stream can start from any value, so decoders probably didn't complain and it was unnoticed.

MaxKellermann: You're right, so this change has nothing to do with this patch. Separate patch, please, so I can cherry-pick this into the stable branch v0.20.x

src/encoder/plugins/OpusEncoderPlugin.cxx
cathugger: rest of things are parsed and initialized inside PreparedEncoder::PreparedEncoder()'s body.
I just did things in consistent way. Should I really change this?

cathugger: though they all had reason to be in body, because they couldn't be done with just initializer so I guess I will

MaxKellermann: That's because the other attributes have more complex parsers. But I welcome patches which move the parsers into separate (throwing) functions and make the other attributes `const` as well. The old code was from a time when MPD had C++ exceptions disabled (`-fno-exceptions`) and error handling was different (more complicated).

cathugger:
@MaxKellermann did changes you requested. didn't fold yet to not disturb previous review

src/encoder/plugins/OpusEncoderPlugin.cxx
cathugger: @MaxKellermann https://github.com/MusicPlayerDaemon/MPD/pull/222

src/encoder/plugins/OpusEncoderPlugin.cxx
cathugger: From RFC 7845:
   An "Ogg Opus file" consists of one or more sequentially multiplexed
   segments, each containing exactly one Ogg Opus stream.  The
   RECOMMENDED mime-type for Ogg Opus files is "audio/ogg".

   If more specificity is desired, one MAY indicate the presence of Opus
   streams using the codecs parameter defined in [RFC6381] and
   [RFC5334], e.g.,

                            audio/ogg; codecs=opus

   for an Ogg Opus file.

so I guess I should leave it as audio/ogg, as it's recommended

cathugger:
squashed

MaxKellermann:
Thanks, this looks good now. Now what's missing is a patch for `doc/user.xml` with documentation and maybe a `NEWS` line telling people about it, then I'll merge it.

cathugger:
@MaxKellermann documented.
hopefully I got docs syntax right.
NEWS line isn't exactly same as commit message, I hope that's ok.

cathugger:
uh, documentation isn't quite correct, it uses ICY headers only in case of http output plugin. need to fix

cathugger:
fixed documentation. now should be all good.

cathugger:
pushed some minor syntax fixes

PR #214, opened 2018-02-06
Hi fi

epicfrequency:
add some hi fi tweak


epicfrequency:
sorry did this by mistake

PR #215, opened 2018-02-07
playlist/Queue: add one-shot to single mode 

hahiserw:
Extends single mode.

If you enable one-shot single mode the music will be paused after current song and the single mode will be disabled. Useful if you don't like leaving playlist paused in the middle of a song and want to be able to resume it afterwards without disabling single mode manually.

I guess you could say this saves you one (or rather two) keystroke every time you want to leave playlist paused not in the middle of a song.

### DIFF
--- a/src/SingleMode.cxx
+++ b/src/SingleMode.cxx
@@ -28,14 +28,14 @@ const char *
 SingleToString(SingleMode mode) noexcept
 {
 	switch (mode) {
-	case SingleMode::ONE_SHOT:
-		return "oneshot";
-
 	case SingleMode::OFF:
-		return "off";
+		return "0";
 
 	case SingleMode::ON:
-		return "on";
+		return "1";
+
+	case SingleMode::ONE_SHOT:
+		return "oneshot";
 	}
 
 	assert(false);
@@ -47,17 +47,12 @@ SingleFromString(const char *s)
 {
 	assert(s != nullptr);
 
-	if (strcmp(s, "off") == 0)
+	if (strcmp(s, "0") == 0)
 		return SingleMode::OFF;
-	else if (strcmp(s, "on") == 0)
+	else if (strcmp(s, "1") == 0)
 		return SingleMode::ON;
 	else if (strcmp(s, "oneshot") == 0)
 		return SingleMode::ONE_SHOT;
-	// backward compatibility?
-	else if (strcmp(s, "1") == 0)
-		return SingleMode::ON;
-	else if (strcmp(s, "0") == 0)
-		return SingleMode::OFF;
 	else
-		throw std::invalid_argument("Unrecognized single mode");
+		throw std::invalid_argument("Unrecognized single mode, expected 0, 1, or oneshot");
 }
--- a/src/SingleMode.cxx
+++ b/src/SingleMode.cxx
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2003-2017 The Music Player Daemon Project
+ * http://www.musicpd.org
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "SingleMode.hxx"
+
+#include <stdexcept>
+
+#include <assert.h>
+#include <string.h>
+
+const char *
+SingleToString(SingleMode mode) noexcept
+{
+	switch (mode) {
+	case SingleMode::ONE_SHOT:
+		return "oneshot";
+
+	case SingleMode::OFF:
+		return "off";
+
+	case SingleMode::ON:
+		return "on";
+	}
+
+	assert(false);
+	gcc_unreachable();
+}
+
+SingleMode
+SingleFromString(const char *s)
+{
+	assert(s != nullptr);
+
+	if (strcmp(s, "off") == 0)
+		return SingleMode::OFF;
+	else if (strcmp(s, "on") == 0)
+		return SingleMode::ON;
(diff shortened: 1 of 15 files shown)


hahiserw:
Sorry for so many commits. I've never used stgit, and forgot about a few things.

src/queue/PlaylistState.cxx
```diff
@@ -152,7 +154,7 @@ playlist_state_restore(const char *line, TextFile &file,
 		} else if ((p = StringAfterPrefix(line, PLAYLIST_STATE_FILE_REPEAT))) {
 			playlist.SetRepeat(pc, StringIsEqual(p, "1"));
 		} else if ((p = StringAfterPrefix(line, PLAYLIST_STATE_FILE_SINGLE))) {
-			playlist.SetSingle(pc, StringIsEqual(p, "1"));
+			playlist.SetSingle(pc, (SingleMode)atoi(p));
```
MaxKellermann: This cast can cause invalid values to be set, e.g. if a state file from a future MPD version happens to get loaded. The state file loader should be robust against such a thing.

src/SingleMode.cxx
```diff
@@ -0,0 +1,63 @@
...
+
+#include "SingleMode.hxx"
+
+#include <stdexcept>
+
+#include <assert.h>
+#include <string.h>
+
+const char *
+SingleToString(SingleMode mode) noexcept
```
MaxKellermann: This function is never used. Why implement it?

src/SingleMode.cxx
```diff
@@ -0,0 +1,63 @@
...
+{
+	assert(s != nullptr);
+
+	if (strcmp(s, "off") == 0)
+		return SingleMode::OFF;
+	else if (strcmp(s, "on") == 0)
+		return SingleMode::ON;
+	else if (strcmp(s, "oneshot") == 0)
+		return SingleMode::ONE_SHOT;
+	// backward compatibility?
```
MaxKellermann: Backward compatibility is important, so it's very good you considered it. I fear that the strings `off` and `on` you introduced are somewhat inconsistent. They are only used client->server, but never server->client. This asymmetry is not so good. I suggest allowing only `1` or `0` or `oneshot`; that's not perfect, but consistent (with what MPD has been doing previously).

src/SingleMode.hxx
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Return the string representation of a #SingleMode.
+ */
+gcc_pure
+const char *
+SingleToString(SingleMode mode) noexcept;
+
+/**
+ * Parse a string to a #SingleMode.  Throws std::runtime_error on error.
```
MaxKellermann: The function throws `std::invalid_argument`, but that's not derived from `std::runtime_error`, see http://en.cppreference.com/w/cpp/error/invalid_argument
(Many MPD API docs refer to `std::runtime_error`, but this is too narrow for many use cases. I've come to the conclusion that callers shall catch all exceptions instead of just `std::runtime_error`, and I'd like to change this in all of MPD.)

src/SingleMode.hxx
```diff
@@ -0,0 +1,48 @@
...
+ */
+SingleMode
+SingleFromString(const char *s);
+
+// #define SingleToString ToString
+// #define SingleFromString FromString
```
MaxKellermann: What's that?

src/command/PlayerCommands.cxx
```diff
@@ -142,7 +143,7 @@ handle_status(Client &client, gcc_unused Request args, Response &r)
 		 COMMAND_STATUS_STATE ": %s\n",
 		 playlist.GetRepeat(),
 		 playlist.GetRandom(),
-		 playlist.GetSingle(),
+		 (int)playlist.GetSingle(),
```
MaxKellermann: This `int` cast is suspicious. Are you aware what this will do for the `oneshot` case? Is this what you wanted it to do?
Your patches don't contain protocol documentation updates, but that would have been interesting to read, how you thought the protocol would be extended here.

src/queue/Playlist.cxx
```diff
@@ -257,7 +258,9 @@ playlist::SetRepeat(PlayerControl &pc, bool status)
 
 	queue.repeat = status;
 
-	pc.LockSetBorderPause(queue.single && !queue.repeat);
+	pc.LockSetBorderPause((queue.single == SingleMode::ON
+			       || queue.single == SingleMode::ONE_SHOT)
```
MaxKellermann: Maybe `single!=SingleMode::OFF`? Easier, more robust and faster.

src/queue/Playlist.cxx
```diff
@@ -277,14 +280,17 @@ playlist_order(playlist &playlist)
...
 {
 	if (status == queue.single)
 		return;
 
 	queue.single = status;
 
-	pc.LockSetBorderPause(queue.single && !queue.repeat);
+
+	pc.LockSetBorderPause((queue.single == SingleMode::ON
+			       || queue.single == SingleMode::ONE_SHOT)
```
MaxKellermann: Same here.

src/queue/Playlist.cxx
```diff
@@ -353,7 +359,9 @@ playlist::GetNextPosition() const noexcept
 	if (current < 0)
 		return -1;
 
-	if (queue.single && queue.repeat)
+	if ((queue.single == SingleMode::ON
+	     || queue.single == SingleMode::ONE_SHOT)
```
MaxKellermann: Same here.

src/queue/PlaylistState.cxx
```diff
@@ -234,7 +236,7 @@ playlist_state_get_hash(const playlist &playlist,
 		(unsigned(player_status.state) << 24) ^
 		(playlist.queue.random << 27) ^
 		(playlist.queue.repeat << 28) ^
-		(playlist.queue.single << 29) ^
+		((playlist.queue.single != SingleMode::OFF) << 29) ^
```
MaxKellermann: This will not detect when the state changes from `ON` to `ONESHOT`.

src/queue/Queue.cxx
```diff
@@ -19,6 +19,7 @@
 
 #include "config.h"
 #include "Queue.hxx"
+#include "SingleMode.hxx"
```
MaxKellermann: Unnecessary `#include` because its header includes it already.

hahiserw:
I've corrected mentioned lines and updated protocol documentation.

src/SingleMode.cxx
hahiserw: It's now used in handle_status in command/PlayerCommands

MaxKellermann:
1628d801f9534ab33a711d337cdd691dca2c5710
Squashed into one commit.

PR #222, opened 2018-02-17
encoder/opus: initialize granulepos to 0

cathugger:
it was uninitialized before


MaxKellermann:
c43ea74b3005b97b9e494ba62d86d0e02ef39d55 (v0.20.x)

PR #223, opened 2018-02-17
youtube-dl plugin

arcnmx:
Usage with the input plugin:
```
mpc add https://www.youtube.com/watch?v=Iab-eTykmBs
mpc add ytdl://https://www.youtube.com/watch?v=Iab-eTykmBs
```

Usage with the playlist plugin:
```
mpc load 'https://www.youtube.com/watch?v=Iab-eTykmBs&list=PLFF158063346349F8'
mpc load ytdl://https://www.youtube.com/playlist?list=PLFF158063346349F8 # equiv to the above
```

Rough items still needing work:
- [ ] code formatting
  - [ ] fix warnings
  - [ ] accessors instead of public fields
  - [ ] I have no idea how to wrap long lines
- [x] rebase onto master
- [x] support `RemoteTagScanner`
- [x] async/EventLoop
- [ ] use a `ProxyInputStream`
- [ ] fix fork/exec
  - [x] `O_CLOEXEC`
  - [ ] signal mask around fork
- [ ] fix `mpc add playlisturl`
- [ ] split out playlist plugin? or at least tone down the metadata assignments for now
- [x] how to initiate `RemoteTagScanner` on a playlist load (fixed in 0.21.7)
- [ ] build properly on windows (either disable plugin, or alternative exec approach)
- [ ] api docs
- [ ] user docs


src/LocateUri.cxx
```diff
@@ -62,7 +63,7 @@ LocateAbsoluteUri(const char *uri
 		  )
 {
 	if (!uri_supported_scheme(uri))
-		throw std::runtime_error("Unsupported URI scheme");
+		throw ProtocolError(ACK_ERROR_NO_EXIST, "Unsupported URI scheme");
```
MaxKellermann: The problem here is not that this URI doesn't exist (it does exist, after all!), but is an invalid URI. This sounds like `ACK_ERROR_ARG` is the correct code.

src/input/plugins/CurlInputPlugin.cxx
```diff
@@ -83,7 +83,7 @@ struct CurlInputStream final : public AsyncInputStream, CurlResponseHandler {
 	CurlInputStream(const CurlInputStream &) = delete;
 	CurlInputStream &operator=(const CurlInputStream &) = delete;
 
-	static InputStream *Open(const char *url, Mutex &mutex, Cond &cond);
+	static InputStream *Open(const char *url, Mutex &mutex, Cond &cond, Tag *tag);
```
MaxKellermann: What's this new parameter? Add API documentation. Why is it writable? Who owns the pointer?

src/input/plugins/CurlInputPlugin.cxx
```diff
@@ -218,7 +218,6 @@ CurlInputStream::OnHeaders(unsigned status,
 	if (i != headers.end()) {
 		TagBuilder tag_builder;
 		tag_builder.AddItem(TAG_NAME, i->second.c_str());
-
```
MaxKellermann: Why remove this line?

src/input/plugins/YtdlInputPlugin.cxx
```diff
@@ -0,0 +1,87 @@
...
+
+#include <cinttypes>
+
+#include <assert.h>
+#include <string.h>
+
+static constexpr Domain ytdl_domain("youtube-dl");
+
+static void
+input_ytdl_init(const ConfigBlock &block)
```
MaxKellermann: What's the point of implementing this?

src/input/plugins/YtdlInputPlugin.cxx
```diff
@@ -0,0 +1,87 @@
...
+{
+	// TODO: a configuration should allow overriding the curl input plugin
+	// this plugin would then accept http/https URIs as well
+}
+
+static void
+input_ytdl_finish() noexcept
```
MaxKellermann: What's the point of implementing this?

src/input/plugins/YtdlInputPlugin.cxx
```diff
@@ -0,0 +1,87 @@
...
+{
+}
+
+struct YtdlUriFlags {
+	bool notags;
+
+	YtdlUriFlags() : notags(false) { }
+};
+
+static std::string ytdl_uri(const char *uri, YtdlUriFlags* flags) {
```
MaxKellermann: Bad C++: must use reference instead of pointer for obligatory pointer parameter; and since this is purely a return value, it should be declared as a return value and not as a parameter.

src/input/plugins/YtdlInputPlugin.cxx
```diff
@@ -0,0 +1,87 @@
...
+	std::string url;
```
MaxKellermann: Useless `std::string` allocation.

src/input/plugins/YtdlInputPlugin.cxx
```diff
@@ -0,0 +1,87 @@
...
+	if (!strncmp(uri, "youtube-dl://", 13)) {
+		uri += 13;
+		if (!strncmp(uri, "notag/", 6)) {
```
MaxKellermann: Use `StringAfterPrefix()` instead.

src/input/plugins/YtdlInputPlugin.cxx
```diff
@@ -0,0 +1,87 @@
...
+			flags->notags = true;
```
MaxKellermann: The URI path is called `notag` but the variable is called `notags`..?

src/input/plugins/YtdlInputPlugin.cxx
```diff
@@ -0,0 +1,87 @@
...
+	if (!strncmp(uri, "youtube-dl://", 13)) {
+		YtdlParseContext context;
+		YtdlUriFlags flags;
+		std::string url = ytdl_uri(uri, &flags);
+		if (!url.empty() && YtdlParseJson(&context, url.c_str(), YTDL_PLAYLIST_MODE_SINGLE)) {
+			if (context.extractor.empty()) {
+				context.extractor = "youtube-dl";
+			}
+			if (!context.builder->HasType(TAG_ALBUM)) {
+				context.builder->AddItem(TAG_ALBUM, context.extractor.c_str());
```
MaxKellermann: album==extractor??

src/input/plugins/YtdlInputPlugin.cxx
```diff
@@ -0,0 +1,87 @@
...
+			}
+			context.builder->AddItem(TAG_COMMENT, context.webpage_url.c_str());
```
MaxKellermann: comment==url?

src/input/plugins/YtdlInputPlugin.cxx
```diff
@@ -0,0 +1,87 @@
...
+		url = uri;
+	}
+
+	return url;
+}
+
+
+static InputStream *input_ytdl_open(const char *uri, Mutex &mutex, Cond &cond)
+{
+	InputStream *res = nullptr;
```
MaxKellermann: On error, must throw exception instead of returning `nullptr`.

src/lib/ytdl/YtdlParser.hxx
```diff
@@ -0,0 +1,61 @@
+#ifndef MPD_LIB_YTDL_PARSER_HXX
+#define MPD_LIB_YTDL_PARSER_HXX
+
+#include <stddef.h>
+#include <string>
+#include <vector>
+#include <forward_list>
+#include <memory>
+#include <utility>
+#include <map>
```
MaxKellermann: Clean up this `#include` mess.

src/lib/ytdl/YtdlParser.hxx
```diff
@@ -0,0 +1,61 @@
...
+	static const void* YajlCallbacks();
+};
+
+enum YTDL_PLAYLIST_MODE {
+	YTDL_PLAYLIST_MODE_SINGLE,
+	YTDL_PLAYLIST_MODE_FLAT,
+	YTDL_PLAYLIST_MODE_FULL,
+};
+
+struct YtdlPlaylistSort {
```
MaxKellermann: Not used here.

src/LocateUri.cxx
arcnmx: Oh right, I just had this in my fork when testing because it made my life easier, as it's impossible to get ncmpcpp to execute a playlist load without throwing this error code back at it. I'll do up a separate PR for that change. That "try add first then load" behaviour is also part of the reason why the input plugin and the playlist plugin use two different URI schemes - not sure if that should be the case, I'll add that as a discussion point.

src/lib/ytdl/YtdlParser.hxx
```diff
@@ -0,0 +1,61 @@
...
+	std::string extractor;
+	int playlist_index;
+
+	YtdlParseContext();
+	~YtdlParseContext();
+
+	static const void* YajlCallbacks();
+};
+
+enum YTDL_PLAYLIST_MODE {
```
MaxKellermann: Strictly-typed enum.

src/lib/ytdl/YtdlParser.hxx
```diff
@@ -0,0 +1,61 @@
...
+};
+
+struct YtdlPlaylistSort {
+	inline bool operator() (const YtdlParseContext& con0, const YtdlParseContext& con1)
+	{
+		return con0.playlist_index < con1.playlist_index;
+	}
+};
+
+bool YtdlParseJson(YtdlParseContext *context, const char *url, YTDL_PLAYLIST_MODE playlist_mode);
```
MaxKellermann: API documentation!

src/lib/ytdl/YtdlParser.cxx
```diff
@@ -0,0 +1,325 @@
...
+	uint8_t buffer[0x80];
+
+	if (pipe(pipefd) < 0) {
+		goto fail;
+	}
+
+	pid = fork();
+
+	if (pid < 0) {
+		goto fail;
```
MaxKellermann: No `goto`!

MaxKellermann:
And just in case you didn't notice the GitHub warning: "This branch has conflicts that must be resolved"

Your branch is based on tag v0.20.17, but you want to merge it into master. That can't work.

src/lib/ytdl/YtdlParser.cxx
```diff
@@ -0,0 +1,325 @@
...
+	bool result = false;
+	int ret;
+	yajl_status status;
+	uint8_t buffer[0x80];
+
+	if (pipe(pipefd) < 0) {
+		goto fail;
+	}
+
+	pid = fork();
```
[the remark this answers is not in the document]
arcnmx: Hm, what is a portable alternative here, is pread available? Or an explicit platform ifdef and use windows APIs?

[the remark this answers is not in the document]
MaxKellermann: You mean `popen()`? (`pread()` is a different beast)
Yes, that is portable, and in use by the known-bad `pipe` output plugin, but is unsafe, because it doesn't restore the signal block mask and the signal handlers. (Your code doesn't do that either.)
Spawning processes is something I like to avoid completely, because it is extremely difficult to get right.
For your plugin, I think it would be acceptable to disable it (for now) on operating systems which don't support `fork()` (i.e. Windows). But in any case, your code must be proper - i.e. get the thing right, be exception-safe, not break the build and not introduce resource leaks (e.g. fd leaks).

doc/user.xml
```diff
@@ -4607,6 +4620,35 @@ run</programlisting>
...
+
+        <para>
+          Downloads a song or playlist with <ulink
+          url="https://rg3.github.io/youtube-dl/">youtube-dl</ulink> from <ulink
+          url="https://rg3.github.io/youtube-dl/supportedsites.html">hundreds of
+          supported audio and video hosting sites</ulink>. <filename>youtube-dl
+          </filename> must be installed and available in the system <varname>
+          PATH</varname>, and the <link linkend="youtube_dl_playlist">youtube-dl
+          input plugin</link> must also be enabled. Add web URLs prefixed with
+          <filename>ytdl://</filename> to use it.
```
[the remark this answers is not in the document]
arcnmx: The playlist and input plugin are distinct and use two different schemes, see that discussion point (partially because of add-then-load fallback behaviour, and partially because the input URI is considered an implementation detail and file add is inferior to the playlist load from a UX perspective even for single file additions).

If I can figure out what RemoteTagScanner is used for and how it works that may make adds with only the input plugin more user friendly if it does metadata loading in the background? I'm not really sure what it does.

src/LocateUri.cxx
MaxKellermann: So this is implemented as a playlist plugin only because you figured this allows you to attach metadata? No, that's not acceptable.
If you want to add one file, `mpc add` must be used, not `mpc load`. Playlists are for playlists.
This should be an input plugin, not a playlist plugin at all. If you want metadata, then yes you can implement a `RemoteTagScanner`.
Note that your code should be non-blocking to make it cancellable. Without this, MPD may block indefinitely when the server is not reachable. (This is a serious problem, and in my ongoing fight against existing blocking code, I don't want to merge any more new blocking code.)
For examples on how to do non-blocking input plugins with HTTP and JSON, look at the Tidal/Qobuz plugins.

src/lib/ytdl/YtdlParser.cxx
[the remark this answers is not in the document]
arcnmx: Er sorry yeah - I had a popen implementation there initially but swapped it out for manual fork/exec because it bothered me. What you've said makes sense to me.

src/LocateUri.cxx
arcnmx: Well, it's a playlist plugin because it adds playlists too. It just happens that adding a single file playlist `load` pulls in the metadata immediately, which is a bit better UX than `add`ing the URL and not having it populate until the file is actually played. Will implementing a `RemoteTagScanner` resolve that?

src/input/plugins/CurlInputPlugin.cxx
arcnmx: This was a hack to apply metadata to a `CurlInputStream` - I don't expect it to stay.

arcnmx:
Yup, rebasing onto master is on the task list. I didn't think it was appropriate to try and get it merged into 0.20.x, but that's what I've been testing with so far for no particular reason. I will probably do that sooner rather than later to get `RemoteTagScanner` available for use.

doc/user.xml
```diff
@@ -2388,6 +2388,19 @@ run</programlisting>
...
+
+      <section id="youtube_dl_input">
+        <title><varname>youtube-dl</varname></title>
+
+        <para>
+          Plays songs using <application>youtube-dl</application>. Use the <link
+          linkend="youtube_dl_playlist">youtube-dl playlist plugin</link>
+          instead of this directly, as it is able to populate more accurate
+          metadata. This plugin depends on curl and uses the <filename>
+          youtube-dl://</filename> URI scheme.
```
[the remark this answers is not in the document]
arcnmx: Invented. Other applications that use `youtube-dl` such as mpv simply pass all http/etc URLs to it instead of using a specific URI scheme. If it fails to load, they move on to try other plugins like curl (or try curl first depending on user configuration). That approach is more user-friendly, and I'd like if it worked that way, but I'm not sure how well it fits into mpd's existing plugin design. `ytdl://` has my vote as it's slightly less annoying to type for a user.

arcnmx:
So there are a lot of comments regarding metadata/tag assignments, I'll just respond here. The following information is (generally, depending on the website) available to us:
- `uploader`
- `upload_date`
- `creator` (often not available, but sometimes youtube links an artist to a video uploaded by someone else for advertising purposes)
- `playlist_webpage_url`
- `playlist_title`
- `webpage_url`
- `title`
- `description`
- `extractor`

There are some obvious mappings to tags:
- `TAG_TITLE` = `title`
- `TAG_ARTIST` = `creator`
  - Rarely available, but when `creator` exists it's a perfect fit.
- `TAG_COMMENT` = `description`
- `TAG_DATE` = `date`
  - I need to format this properly

There are a few mappings to tags that aren't as straightforward:
- `TAG_ARTIST` = `uploader`
  - This is true on websites like soundcloud or bandcamp. On youtube it's often true, but obviously not always.
  - I'd apply this as a fallback if `creator` is not available.
- `TAG_ALBUM` = `playlist_title`
  - This is true on bandcamp and soundcloud. On youtube that is not necessarily the case, but it is still often true when adding an album playlist that contains tracks.
  - Considering that real album information is never otherwise available, I'd like to apply this as well.

There are a few mappings that aren't really necessary but could be considered:
- `TAG_ALBUM` = `extractor`
  - Even if we do apply playlist title as the album, single files added won't have that.
  - Having a song say it's `Song Name - YouTube` seems somewhat relevant and makes it easy to identify remote songs in your play queue.
- `TAG_COMMENT` += `webpage_url`
  - It's very convenient to have a link to the video so you can link it to others, or visit it to view user comments, etc.
- `TAG_COMMENT` += `playlist_webpage_url`
  - Sort of the same reasoning as `webpage_url`, but probably a bit less often useful.

My inclination is to fill in as much metadata fields as we can, but I'm open to toning it down, or perhaps making some of the not-always-accurate mappings into configuration options. Also note that tags from the decoder once the file is played should overwrite these if available.

src/LocateUri.cxx
MaxKellermann: "because it adds playlists too" that's a different feature which requires implementing a different plugin. I think the most important feature is playing one single file by its id. You can add a playlist feature later, but that's not the primary goal.
With "populate", you mean provide metadata? Yes, `RemoteTagScanner` can do that, but it's an input plugin feature.

doc/user.xml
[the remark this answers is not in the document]
MaxKellermann: Can't we just make the plugin intercept all plain http:// YouTube URLs? That's what the user already has.

MaxKellermann:
I don't think filling as many metadata fields as possible is a good goal, not when the tags do not match the contents. Only fill those which match, and if you feel others are important, we can discuss later how to integrate them. For now, to get this merged, let's get to a minimal commit; one that is stable and robust.

doc/user.xml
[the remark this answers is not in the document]
arcnmx: We certainly can, and I think that will work best! My concern was that it may interfere with the curl plugin, since I don't see any explicit support for falling back to another plugin, but that could instead be done with `OpenCurlInputStream(uri)` if running the `youtube-dl` fails.

src/LocateUri.cxx
MaxKellermann: I will never add kludges to MPD just because some client decided to make funny assumptions on something that's not explicitly documented in the protocol spec (and yes, the protocol spec isn't very specific on some things, and needs to improve).
I'm not sure if I should call this ncmpcpp behavior "bug", but I wouldn't have done that. I maintain ncmpc (of which ncmpcpp is a clone with a very unfortunate name), and ncmpc doesn't do it. Andd I maintain mpc, which doesn't do that either.

doc/user.xml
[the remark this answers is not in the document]
MaxKellermann: It doesn't interfere because CURL wouldn't be able to get meaningful data from a `http://youtube.com/...` URL.

There is a "fallback" - the MPD core iterates over all plugins and the first one which doesn't return `nullptr` wins: https://github.com/MusicPlayerDaemon/MPD/blob/master/src/input/Open.cxx#L40

Yes, a whitelist would probably be good. Can it be auto-generated at runtime somehow from calling youtube-dl once?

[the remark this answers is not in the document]
arcnmx: See [How can I detect whether a given URL is supported by youtube-dl?](https://github.com/rg3/youtube-dl/#how-can-i-detect-whether-a-given-url-is-supported-by-youtube-dl) It doesn't seem like we could generate it, but I think a whitelist also makes sense even if we just populate it with some high-profile websites. Personally I'd be inclined to just set it to `.*` as long as the fallback behaviour is working, sounds like a config option to me.

The issue with the fallback is that curl plugin will return a stream for any valid URL, before it makes the request and validates that there's playable media at the end of it. Otherwise I'm not sure we would need a whitelist - but I think a whitelist will help, and if we do eventually fail, we can fall back to the original provided URI anyway.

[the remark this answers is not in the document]
MaxKellermann: No catch-all!
The fallback is only used when the plugin decides it does not support the URI, not when the plugin fails to open the URI.
And don't consider interpreting `youtube-dl`'s failure to open something as "not supported"; it is an error.
There must not be overhead for "regular" HTTP streams just because your plugin spawns `youtube-dl` each time.

arcnmx:
Sounds good. The second category of "not as straightforward" mappings are pretty important to me personally, because they do still contain very relevant information on other platforms and usually youtube too. The third category is very meh, and easy to leave out.

I think if we're going minimal I may just include the playlist plugin in a supplementary PR - which simplifies the metadata decisions as well.

src/LocateUri.cxx
arcnmx: Mm, not so much a bug as a missing feature - it's unable to load playlists, but just happens to be able to if the server returns the right error when adding a file!

On an unrelated note perhaps I should look into whether I can use ncmpc for personal use.

doc/user.xml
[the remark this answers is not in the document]
arcnmx: > There must not be overhead for "regular" HTTP streams just because your plugin spawns youtube-dl each time.

Agreed, which is why the whitelist must exist.

> And don't consider interpreting youtube-dl's failure to open something as "not supported"; it is an error.

Hm, I actually object to this, because as a user I know that every URL I use is supported by youtube-dl, and am okay with the rare said overhead for "regular" streams if I opt into it. Although I don't think it needs to be the default behaviour, it's reasonable behaviour that works like other tools I'm used to such as mpv, and not having the option to do that would be a bit off-putting.

MaxKellermann:
What will happen to this PR? Is this being worked on, or is this obsolete?

arcnmx:
It's still being worked on (temporarily in the [ytdl-async](https://github.com/arcnmx/MPD/tree/ytdl-async) branch), though slowly because some other life things have gotten in the way recently.

viisauksena:
i would love to test and if i can elaborate on this .. 
can you briefly explain how to use it after build?

roizcorp:
I found this by mistake, what was the end of it?

viisauksena:
basicly you simply need to parse some "special" url into youtube-dl script(different program) to get valid streaming link - while we made some experiments in CCCFr with this, this is on hold by us, until we find time and joy to do it.  so far from our point

roizcorp:
Yes I am familiar with youtube-dl. I think there is a great potential to use it with mpd

moretocome:
Well it may be to late for this project, but I still just want to mention [NewPipe Extractor](https://github.com/TeamNewPipe/NewPipeExtractor) as an alternative to youtube-dl.

It is the open source library that enable the Android apps NewPipe and MusicPiped to stream audio from Youtube without displaying the video. Soundcloud streaming is also supported.

MaxKellermann:
@moretocome this code is useless for MPD because it's Java. Certainly, MPD won't load a full JavaVM into the process only to use YouTube ...

moretocome:
@MaxKellermann Okay excuse my ignorance, my coding skills are limited to light scripting, so I did not know that, but it makes perfect sense to me and my low-powered but smooth running raspberry pi based mpd-server :)

roizcorp:
Cool!

MaxKellermann:
Oh, it's been two years already... I'm going to close this for now; let's reopen this when you are finished and want a new review.

arcnmx:
That's fair, my life fell apart in the meantime and I'm still trying to find time to complete this. It's really just the organization+formatting+docs I need to get to, but realistically I won't have a day to spare until probably April/May at this rate.

Tajnymag:
@arcnmx Would you mind looking into this patch again, please?

No stress, no worries, even if you didn't want to complete the work. I just felt were all interested in some kind of update on what has happened to you personally and the patch.

arcnmx:
It's still on my radar, I've mostly finished a move now but work has just kept me a *lot* busier than it has any right to... Things are hopefully slowing down over the next few weeks, but I said the same thing a few weeks ago and then more problems came up ._.

FWIW the [existing patch](https://github.com/MusicPlayerDaemon/MPD/compare/v0.21.16...arcnmx:ytdl-0.21.16.diff) should still be working and apply to the latest 0.21.25 release, feel free to try it and report how it goes. I need to start using it again myself, I just still haven't had a chance to set up my machine yet after all these months...

Emilot:
Can i ask how can we test this plugin? What are the requirements?

- Download the source code from 0.21.5 branch at your git, compile it and install it

- I have enable input plugin at mpd.conf

- I tried with mpc add ytdl://https://www.youtube.com/watch?v=Iab-eTykmBs and mpc add https://www.youtube.com/watch?v=Iab-eTykmBs

First return "Unsupported URI scheme"
Second "cannot decode stream"

I have the light version of ffmpeg. Does it play any role?

Thank you for your work at this. It really misses in MPD.

arcnmx:
> Can i ask how can we test this plugin? What are the requirements?

Beyond compiling with the patch, you will need the youtube-dl binary installed (and available in $PATH when running the service).

> 0.21.5

I assume this is a typo, but yeah 0.21.25 branch is what I'm using and should be working fine.

> I have these at mpd -V
> I tried with mpc add ytdl://

These do work here... This might be a silly question, but have you made sure you restarted your mpd (systemd?) service since installing the patched mpd?

> I have the light version of ffmpeg. Does it play any role?

It shouldn't matter, adding a ytdl:// would succeed, then possibly playing it would fail after if it were a problem.

Emilot:
Oh sorry forgot to mention that i had installed the youtube-dl binary, before the compiling. I restart the service after the compile was done.

Sorry about the typo. 

Must be a mistake in my configuration, then. I wanted to know that all i want is:

youtube-dl
your version of mpd
input plugin at mpd.conf

Nothing else.

Thank you and we are waiting for the final PR at Max's git! 👍

Rio6:
I found out there's a pr for youtube url already after I made my own...

Anyways, my implementation is at [https://github.com/Rio6/MPD/tree/youtube-input](https://github.com/Rio6/MPD/tree/youtube-input), people can try it out too. It takes `https://` url that youtube-dl supports.

Not sure what I'll do with it after this gets merged in master :P

Emilot:
Hi RIO6! I had seen your work but i haven't a way to communicate with you about it.

After implementation you can make a PR youtube plugin for myMPD project. I do not know if you are aware about it.

Thank you for your work too! Nice if MAX implement any of these two projects.

Rio6:
Replying to https://github.com/MusicPlayerDaemon/MPD/pull/223#issuecomment-366427063

I think using uploader and playlist title as fall back to artist and album is reasonable, as that's often the case on Youtube.

Not sure how long TAG_COMMENT supposed to be, but I know in many videos the description is very long. Video url is probably nicer as comment.

mnzaki:
Hey folks, thank you everyone involved in getting [youtube-dl integrated with MPD](# "it's the last bit I need for my Amazingly Decentralized Media Management and Consumption System")!
There are a several issues/PRs and I just want to confirm that the separate efforts were merged and the latest work on this is at https://github.com/arcnmx/MPD on the [ytdl branch](https://github.com/arcnmx/MPD/tree/ytdl "which is the default anyway but for historical propriety")
Also curious about the state of this being merged into master, I have some time to help.

For me and future searchers who are looking for easily saving youtube vimeo soundcloud instagram facebook videos and music through youtube-dl directly into their music/media collection in their favorite [MPD](https://musicpd.org) [user interface or client](https://www.musicpd.org/clients/) like [ncmpcpp](https://rybczak.net/ncmpcpp/) or [Persephone](https://persephone.fm/) or [myMPD](https://jcorporation.github.io/myMPD/)

Emilot:
Latest development is at:

https://github.com/arcnmx/MPD/tree/ytdl-0.23.8

@arcnmx is doing a wonderful job to bring this into life!!!

@mnzaki if you have time then please take a look at this project and help @arcnmx, that will be awesome :)

thank you everyone for keep this alive!

ItsTerm1n4l:
Has there been any progress on this?

PR #229, opened 2018-02-21
Fix curl storage plugin failure if the authentication method is different than basic.

tpoeiras:


MaxKellermann:
3f3e0739c4df2454118ca5bd114525d031b28840

PR #237, opened 2018-02-26
Add initial DSD over PCM support to macOS output

coroner21:
This just copies the necessary bits and pieces from the ALSA plugin and applies them to OSXOutput based on dop config setting. It changes only the OSXOutput plugin as needed for DoP (further changes to support additionally e.g. integer mode or setting the physical device mode require rather a complete rewrite of the output plugin).

Fortunately the Core Audio API is by default bit perfect and supports DoP with minimal changes (setting the sampling rate accordingly after ensuring that the physical mode supports at least 24 bits per channel seems to be enough). Tested on macOS High Sierra using an Amanero Combo384 device hooked up to an ES9018 DAC.

USAGE (try only on DACs that support DoP):
- Add dop "yes" option to mpdconf
- Be sure to set at least 24bits per channel before playing some DSD file (using Audio-MIDI-Setup)
- Based on the dop setting, MPD will change the sample rate as required and output DoP signal to the DAC
- Hog mode is recommended to ensure that no other program will try to mix some output with the DoP stream (resulting in bad noise)
- Alternatively set the default output device to another device (e.g. the built-in output) to avoid having other audio interfere with DSD playback

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -24,6 +24,7 @@ ver 0.21 (not yet released)
   - ao: fix crash bug due to partial frames
   - shout: support the Shine encoder plugin
   - sndio: remove support for the broken RoarAudio sndio emulation
+  - osx: initial support for DSD over PCM
 * mixer
   - sndio: new mixer plugin
 * encoder
--- a/doc/user.xml
+++ b/doc/user.xml
@@ -4332,6 +4332,26 @@ run</programlisting>
                   select the best possible for each file.
                 </entry>
               </row>
+			  <row>
+                <entry>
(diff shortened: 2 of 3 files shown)


MaxKellermann:
Please fix your coding style (indent with tabs) and add documentation to `doc/user.xml` (and optionally a line to `NEWS`).

coroner21:
Thanks, updated!

MaxKellermann:
e89c421313b53ed98b249cccb9aee3a0d0344a3d

PR #248, opened 2018-03-07
doc/protocol.xml: Add examples to the description of 'count'

DamienCassou:


DamienCassou:
I would have liked to see how it looks but the Makefile doesn't seem to contain anything to generate html files.

MaxKellermann:
There are the rules: https://github.com/MusicPlayerDaemon/MPD/blob/master/Makefile.am#L2454-L2471

DamienCassou:
I had a look at my Makefile but important things were missing regarding documentation. With your link, I learned about the option `--enable-documentation` which makes the Makefile much more reasonable :-).

Thank you.

Issue #249, opened 2018-03-07
ALSA output noise, Centos 7 compilation


msmucr:
Hi,
during my tests of current development branch, I had one weird issue.
When playing audio via ALSA plugin, there is full blast noise. At some hardware configurations, there can be heard some momentary pieces of intended sound (so it doesn't look like complete garbage with swapped byteorder or likes), but still 95% of the time, there is the noise.

- v0.20.x branch is fine
- master branch compiled and used at Fedora 27 and Ubuntu 16.04 is fine (just two different Linux distributions, which I had by hand or in VM)

- AO library, as well as caputred audio output to wav file via file recorder is fine, so it affects just ALSA output plugin
- it happens with different audio hardware (few audio interfaces, either via direct hw access, plughw.. it doesn't matter) or different computers
- I haven't found any xruns during playback
- there's likely no issue with autodetection of period, buffer size, as it's exactly the same like with working MPD version. Also verified those parameters via /proc.
- i've managed to compile minimal mpd binary at Centos and tested it at Fedora - the issue persisted, so there's likely something in the compiled binary and not at runtime libraries.
- i tried to upgrade ALSA library to current version 1.1.5 (just vanilla version) and compile mpd against that. No changes.
- Centos 7 itself have too old gcc to fulfill build requirements for C++14 capable compiler, so I generally use "official" SCL, https://www.softwarecollections.org/en/scls/rhscl/devtoolset-7 with gcc7 and corresponding C++ standard library. Normally I haven't experienced any issues with that.
To test another compiler, I've tried also llvm/clang compilation.. first with GNU stdlibc++, second time with current libc++ and libc++abi from LLVM project.
- briefly checked few commits affecting ALSA plugin.. and tried lets say five different compilations with rolled back source tree changes to some previous commits.. (the earliest tested was from last June) 
Honestly those were more like random picks among commits, to find out some "good" version at least. That also wasn't succesfull. Only way to compile working version at Centos 7 was to switch to v0.20.x branch.

After two evenings, I'm bit clueless..  :) There has to be something, which is causing that problem. Do you have any tip, what else to try..?

Do you think, that for example particular Boost library (1.53) at Centos could be problem..? And is it worth of testing its different version?

All the best,

Michal

msmucr:
O.K. I've probably got that.
Boost library was the culprit.

[...]

PR #251, opened 2018-03-08
Bump minimum required version of Boost to 1.54

msmucr:
lockfree library used by ALSA output plugin is part of Boost from version 1.53,
so this can be theoretically the lowest required version, however
there are issues which are resolved from 1.54 onward.

corresponding issue: #249

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -465,7 +465,7 @@ dnl ---------------------------------------------------------------------------
 dnl Mandatory Libraries
 dnl ---------------------------------------------------------------------------
 
-AX_BOOST_BASE([1.46],, [AC_MSG_ERROR([Boost not found])])
+AX_BOOST_BASE([1.54],, [AC_MSG_ERROR([Boost not found])])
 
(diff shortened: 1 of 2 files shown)


MaxKellermann:
Cherry-picked to v0.20.x: c745e14f4704ebfb0294e8b22a7ae503fc05c2a4

msmucr:
Thanks, but v0.20.x branch was fine even with older versions, because there isn't used queue from boost/lockfree.. so the issue was with current master branch, unless version => 1.54 was used during build. 
Maybe I wrote that in confusing way at issue description, sorry for that.

It likely doesn't harm to have 1.54 requirement in v0.20.x, besides it would cut out direct compatibility with boost packages at older Linux distributions (Wheezy, RHEL/Centos 7). So maybe unless you intend to merge the alsa plugin with lockfree also to v0.20.x, it would make more sense to keep that as is and leave that to master and upcoming 0.21.x.

All the best,
Michal

Issue #265, opened 2018-04-18
Boost 1.67.0 build failure


ilovezfs:
```
rm -f libmixer_plugins.a
ar cru libmixer_plugins.a src/mixer/plugins/libmixer_plugins_a-NullMixerPlugin.o src/mixer/plugins/libmixer_plugins_a-SoftwareMixerPlugin.o     src/mixer/plugins/libmixer_plugins_a-OSXMixerPlugin.o   
ranlib libmixer_plugins.a
clang++ -std=c++14 -DHAVE_CONFIG_H -I.  -DNDEBUG -I./src -I/usr/local/include -DSYSTEM_CONFIG_FILE_LOCATION='"/usr/local/etc/mpd.conf"'           -Qunused-arguments -pthread -g -O2 -fvisibility=hidden -fno-threadsafe-statics -fmerge-all-constants -ffast-math -ftree-vectorize -ffunction-sections -fdata-sections -Wall -Wextra -Wmissing-declarations -Wshadow -Wpointer-arith -Wcast-qual -Wwrite-strings -Wsign-compare -Wno-noexcept-type -c -o src/output/plugins/liboutput_plugins_a-OSXOutputPlugin.o `test -f 'src/output/plugins/OSXOutputPlugin.cxx' || echo './'`src/output/plugins/OSXOutputPlugin.cxx
In file included from src/output/plugins/OSXOutputPlugin.cxx:35:
/usr/local/include/boost/lockfree/spsc_queue.hpp:140:43: error: no member named 'next' in namespace 'boost'
        const ConstIterator last = boost::next(begin, input_count);
                                   ~~~~~~~^
/usr/local/include/boost/lockfree/spsc_queue.hpp:145:51: error: no member named 'next' in namespace 'boost'
            const ConstIterator midpoint = boost::next(begin, count0);
                                           ~~~~~~~^
2 errors generated.
make[1]: *** [src/output/plugins/liboutput_plugins_a-OSXOutputPlugin.o] Error 1
make: *** [all] Error 2
```
https://gist.github.com/ilovezfs/67bba5f253530569ff7efd4fc5fbf392

MaxKellermann:
Sorry, that is a Boost bug.

ilovezfs:
@MaxKellermann OK we'll patch it: https://github.com/Homebrew/homebrew-core/pull/26784

PR #266, opened 2018-04-18
Fix build with Boost 1.67.0

ilovezfs:
Fixes https://github.com/MusicPlayerDaemon/MPD/issues/265.

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -35,6 +35,7 @@
 #include <CoreAudio/CoreAudio.h>
 #include <AudioUnit/AudioUnit.h>
 #include <CoreServices/CoreServices.h>
+#include <boost/next_prior.hpp>
 #include <boost/lockfree/spsc_queue.hpp>
 
 #include <memory>


MaxKellermann:
Sorry, that is a Boost bug. Wrong place to fix this.

ilovezfs:
@MaxKellermann regardless of whether you consider it a "Boost bug" it seems undesirable for mpd to be unable to build with Boost 1.67.

ilovezfs:
@MaxKellermann are you suggesting that every downstream distribution of Boost needs to backport https://github.com/boostorg/lockfree/commit/12726cda009a855073b9bedbdce57b6ce7763da2 ?

MaxKellermann:
That looks like the fix, so yes, vanilla 1.67.0 should not be used by anybody. Either keep using 1.66 or patch 1.67. That's better than patching each and every application using Boost.

But WTF, this patch is 4 months old, and hasn't found its way into the Boost version released just a few days ago. There appears to be something wrong with Boost's patch management.

ilovezfs:
@MaxKellermann yeah ... tell me about it. This happens like every time :(

MaxKellermann:
Ha, that's the dirty real world!
Shit happens, but then you clean up, but don't cover it.

PR #275, opened 2018-05-04
decoder/opus: add support for R128_ALBUM_GAIN tag

autrimpo:
Opus for some reason doesn't read album gain tags, this adds support for it.


MaxKellermann:
Cherry-picked to v0.20.x because it looked rather trivial.

PR #282, opened 2018-05-24
Added always_on to streaming example in mpdconf.example

michamilz:
always_on yes|no 	
If set to yes, then MPD attempts to keep this audio output always open. This may be useful for streaming servers, when you don't want to disconnect all listeners even when playback is accidentally stopped.


MaxKellermann:
I don't think it is useful to add each and every setting to that file. It is way too complex already. Advanced users shall consult the documentation for a list of supported settings.

PR #286, opened 2018-06-02
Fix pause for OSX output plugin (delay one second)

coroner21:
While working on a new Mac OS output plugin I found that Pause() was not reflected in Delay() function causing high CPU load. Apologies for the oversight in the first place...


MaxKellermann:
Rule of thumb: if your PR needs text, then your commit message is bad.

Issue #271, opened 2018-04-24
osx sync_sample_rate does not sync bit depth


cavemandaveman:
macOS 10.13.4
mpd 0.20.18

The `osx` `audio_output` option `sync_sample_rate` sets the correct sample rate on the audio device, but does not take into account the bit depth.

For example, if I play a 88.2 kHz / 24-bit file, my audio output in "Audio MIDI Setup" switches to the 88.2 kHz / 16-bit Format (I'm assuming because that's the first 88.2 kHz option in the list).

Ideally mpd would check the bit depth as well, and update the output accordingly. It would be great to achieve true bit-perfect playback.

MaxKellermann:
https://www.musicpd.org/help/

cavemandaveman:
I don't understand why you linked to the help page..? Also, not sure why I am listed as an assignee now?

I would consider this a bug if `sync_sample_rate` is meant to sync bit depth as well as sample rate. If not, then this would be a feature request to sync output bit depth.

MaxKellermann:
`auto_update` is only implemented on Linux, and `id3v1_encoding` has been removed a while ago.

But let me judge whether the log is relevant. If you see nothing relevant, then either your judgement is wrong, or you're looking at the wrong file.

coroner21:
Just to clear this up: This is a feature request. The Mac output is implemented in a way that it only changes samplerate. In my view this should still be fine for bit perfect playback (just set the highest bitrate the DAC / soundcard supports)...

Still, I think it would be desirable to change bitrate automatically and also support the macOS integer mode (bypassing the CoreAudio float32 processing).

MaxKellermann:
I know it's a feature request, but the thing is that this feature is already implemented.
The submitter said his MPD plays 24 bit files at 16 bit. If that happens, it's a bug, because MPD should switch to 32 bit mode and not 16 bit (according to `OSXOutput::Open()`).
I'd like to see why that happens, but the submitter refuses to give me the required data, which I think is pretty annoying.

cavemandaveman:
@MaxKellermann I am not refusing to give you data. I want to help as much as I can, that's why I opened the issue in the first place. Forgive me, at first I did not have logging set to verbose.

Here is the logfile during playback. Something definitely looks wrong with all the mismatch lines.

```
Apr 25 14:41 : playlist: play 0:"HiRes.flac"
Apr 25 14:41 : client: [0] command returned 0
[...]

PR #292, opened 2018-06-10
New MacOS output plugin using CoreAudio (rewrite)

coroner21:
Thanks for the feedback. Will adjust and work on this accordingly. Being new to C++ and also not a programmer I can always use some guidance.

To comment on this further, this provides following improvements:
- This is a stab at #271 adding all the API calls required to change bit depth and sample rate
- It has the additional feature of integer mode: Passing directly SInt samples to the device which is a desired feature by audiophiles
- It directly accesses the devices buffers in the callback so no overhead from AudioUnit framework

Since there are also drawbacks (e.g. currently no callback for device changes implemented and therefore change to sample rate or change in output device cannot be handled) I thought it might be best to keep the existing available as fallback. Complete drop-in replacement would however also be an option. You may have noticed that the plugin and mixer code is not really a complete rewrite since the structure / options etc. have been taken over. The way CoreAudio API is handled is completely different though: AudioUnit vs. raw device access.


MaxKellermann:
Please keep only the rebased commit in this PR - not the old one, and not the merge commit.
Is it useful to keep the old plugin? Ideally, I'd like a new one to be a drop-in replacement, and users should not be forced to change their configuration; everything works just like before, or better.
Also the commit message should point out why this rewrite was done. What problems does it solve? How is it better? Why did you do a rewrite from scratch instead of improving the old one? Complete rewrites are almost always a bad idea.

src/lib/coreaudio/CoreAudioDevice.hxx
```diff
@@ -0,0 +1,119 @@
...
+#include "CoreAudioStream.hxx"
+#include "AudioFormat.hxx"
+#include <CoreAudio/CoreAudio.h>
+
+class CoreAudioDevice
+{
+public:
+	CoreAudioDevice();
+	explicit CoreAudioDevice(AudioDeviceID dev_id);
+	virtual ~CoreAudioDevice();
```
MaxKellermann: Why is this `virtual`?

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+#include "util/StringBuffer.hxx"
+#include <AudioToolbox/AudioToolbox.h>
+#include "system/ByteOrder.hxx"
+#include "util/Domain.hxx"
+#include "Log.hxx"
+
+static constexpr Domain macos_output_domain("macos_output");
+
+CoreAudioDevice::CoreAudioDevice()  :
+	started					(false	),
```
MaxKellermann: Bad coding style (whitespace!). Also use C++11 initializers in this case.

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+	output_format.mFormatID = 0;
+}
+
+CoreAudioDevice::~CoreAudioDevice()
+{
+	Close();
+}
+
+bool
+CoreAudioDevice::Open(const std::string &device_name)
```
MaxKellermann: Why use `std::string` here?

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+		return false;
+	
+	if(!Enumerate())
+	{
+		FormatError(macos_output_domain, "Failed to get stream information from output device.");
+		return false;
+	}
+
+	AudioObjectPropertyAddress aopa;
+	aopa.mScope		= kAudioDevicePropertyScopeOutput;
```
MaxKellermann: Again bad whitespace.

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+bool
+CoreAudioDevice::Enumerate()
+{
+	AudioStreamIdList stream_list;
+	bool ret = false;
+	stream_infos.clear();
+	is_planar = true;
+	
+	if (GetStreams(&stream_list))
+	{
```
MaxKellermann: Coding style: curly brace on previous line.

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+		return;
+
+	// Stop the device if it was started
+	Stop();
+
+	// Unregister the IOProc if available
+	RemoveIOProc();
+
+	// Make sure to release hog state
+	SetHogStatus(false);
```
MaxKellermann: What if this function throws? Will this leak memory? (If `SetHotStatus()` cannot throw, it should be declared that way.)
This "problem" probably affects much of this patch.

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+
+	if (err != noErr)
+	{
+		FormatError(macos_output_domain, "Unable to get device name - id: 0x%04x. Error = %s", (uint)device_id, GetError(err).c_str());
+	}
+	else
+	{
+		// The +1 is for having space for the string to be NUL terminated
+		CFIndex bufferSize = CFStringGetLength(deviceName) + 1;
+		char buffer[bufferSize];
```
MaxKellermann: Variable-length array is not C++. Only C99.

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+
+	std::string name;
+	CFStringRef deviceName = nullptr;
+	UInt32 propertySize = sizeof(deviceName);
+
+	OSStatus err = AudioObjectGetPropertyData(device_id, &aopa, 0, NULL, &propertySize, &deviceName);
+
+	if (err != noErr)
+	{
+		FormatError(macos_output_domain, "Unable to get device name - id: 0x%04x. Error = %s", (uint)device_id, GetError(err).c_str());
```
MaxKellermann: Should this better throw an exception? I mean, what does it return now, an empty string? Do all callers check this condition?

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+	{
+		// The +1 is for having space for the string to be NUL terminated
+		CFIndex bufferSize = CFStringGetLength(deviceName) + 1;
+		char buffer[bufferSize];
+		
+		if (CFStringGetCString(deviceName, buffer, bufferSize, kCFStringEncodingUTF8))
+		{
+			name.assign(buffer);
+		}
+		CFRelease(deviceName);
```
MaxKellermann: This call is omitted (and causes memory leak) if an exception is thrown before. Please review this whole patch for exception-safety!

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+	{
+		for(UInt32 buffer = 0; buffer < buffer_list->mNumberBuffers; ++buffer)
+			channels += buffer_list->mBuffers[buffer].mNumberChannels;
+	}
+	else
+	{
+		FormatError(macos_output_domain, "Unable to get total device output channels - id: 0x%04x. Error = %s", (uint)device_id, GetError(err).c_str());
+	}
+
+	free(buffer_list);
```
MaxKellermann: Not exception-safe, again!

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+	aopa.mScope    = kAudioDevicePropertyScopeOutput;
+	aopa.mElement  = 0;
+	aopa.mSelector = kAudioDevicePropertyStreamConfiguration;
+
+	UInt32 size = 0;
+	OSStatus err = AudioObjectGetPropertyDataSize(device_id, &aopa, 0, NULL, &size);
+	if (err != noErr)
+		return channels;
+
+	AudioBufferList* buffer_list = (AudioBufferList*)malloc(size);
```
MaxKellermann: Use `new[]` instead of `malloc()` in C++!

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+	aopa.mScope    = kAudioDevicePropertyScopeOutput;
+	aopa.mElement  = 0;
+	aopa.mSelector = kAudioDevicePropertyStreamConfiguration;
+
+	UInt32 size = 0;
+	OSStatus err = AudioObjectGetPropertyDataSize(device_id, &aopa, 0, NULL, &size);
+	if (err != noErr)
+		return channels;
+  
+	AudioBufferList* buffer_list = (AudioBufferList*)malloc(size);
```
MaxKellermann: No `malloc()`... Is this allocation size even correct?

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+	err = AudioObjectGetPropertyData(device_id, &aopa, 0, NULL, &size, buffer_list);
+	if (err == noErr)
+	{
+		if (stream_idx < buffer_list->mNumberBuffers)
+			channels = buffer_list->mBuffers[stream_idx].mNumberChannels;
+	}
+	else
+	{
+	  FormatError(macos_output_domain, "Unable to get number of stream output channels - id: 0x%04x. Error = %s.",
```
MaxKellermann: Throw exception?

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+	AudioStreamID* streams_list = new AudioStreamID[stream_count];
+	
+	err = AudioObjectGetPropertyData(device_id, &aopa, 0, NULL, &size, streams_list);
+	if (err == noErr)
+	{
+		for (UInt32 stream = 0; stream < stream_count; stream++)
+			stream_id_list->push_back(streams_list[stream]);
+	}
+	
+	delete[] streams_list;
```
MaxKellermann: Not exception-safe!

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+		return false;
+
+	UInt32 stream_count = size / sizeof(AudioStreamID);
+	AudioStreamID* streams_list = new AudioStreamID[stream_count];
+	
+	err = AudioObjectGetPropertyData(device_id, &aopa, 0, NULL, &size, streams_list);
+	if (err == noErr)
+	{
+		for (UInt32 stream = 0; stream < stream_count; stream++)
+			stream_id_list->push_back(streams_list[stream]);
```
MaxKellermann: Return value instead of passing an output-parameter!

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+			 had success on getting hog status
+			 We do this only when AudioObjectSetPropertyData didn't set hog_pid because
+			 it seems that in the other cases the GetHogStatus could return -1
+			 which would overwrite our valid hog_pid again */
+			if (hog_pid == -1)
+				hog_pid = GetHogStatus();
+
+			if (err || hog_pid != getpid())
+			{
+				FormatError(macos_output_domain, "Unable to set hog mode. Error = %s", GetError(err).c_str());
```
MaxKellermann: Throw exception?

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+		return false;
+
+	AudioObjectPropertyAddress  aopa;
+	aopa.mScope    = kAudioDevicePropertyScopeOutput;
+	aopa.mElement  = 0;
+	aopa.mSelector = kAudioDevicePropertyHogMode;
+
+	pid_t hogPid = -1;
+	UInt32 size = sizeof(hogPid);
+	AudioObjectGetPropertyData(device_id, &aopa, 0, NULL, &size, &hogPid);
```
MaxKellermann: Error check missing?

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+
+	AudioObjectPropertyAddress  aopa;
+	aopa.mScope    = kAudioDevicePropertyScopeOutput;
+	aopa.mElement  = 0;
+	aopa.mSelector = kAudioHardwareServiceDeviceProperty_VirtualMasterVolume;
+
+	OSStatus err = AudioObjectSetPropertyData(device_id, &aopa, 0, NULL, sizeof(Float32), &vol);
+	if (err != noErr)
+	{
+		FormatError(macos_output_domain, "Unable to set output device volume. Error = %s", GetError(err).c_str());
```
MaxKellermann: Throw exception?

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+	float output_score  = 0;
+	
+	// loop over all streams to find one with a suitable format
+	for(UInt32 stream_idx = 0; stream_idx < stream_infos.size(); stream_idx++)
+	{
+		
+		// Probe physical formats
+		const StreamFormatList *formats = &stream_infos[stream_idx].format_list;
+		
+		for (StreamFormatList::const_iterator j = formats->begin(); j != formats->end(); ++j)
```
MaxKellermann: Use range-based `for`

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,676 @@
...
+bool CoreAudioDevice::SetFormat(const AudioFormat &audio_format, bool prefer_unmixable)
+{
+	FormatDebug(macos_output_domain, "Finding CoreAudio stream for format %s.", ToString(audio_format).c_str());
+	
+	bool format_found  = false;
+	memset(&output_format, 0, sizeof(output_format));
+	float output_score  = 0;
+	
+	// loop over all streams to find one with a suitable format
+	for(UInt32 stream_idx = 0; stream_idx < stream_infos.size(); stream_idx++)
```
MaxKellermann: Use range-based `for`

src/lib/coreaudio/CoreAudioDevice.hxx
MaxKellermann: Why `virtual`?

src/lib/coreaudio/CoreAudioDevice.hxx
```diff
@@ -0,0 +1,119 @@
...
+	AudioDeviceID device_id;
+	UInt32 output_stream_idx;
+	AudioStreamBasicDescription output_format;
+	AudioDeviceIOProc io_proc;
+	CoreAudioStream output_stream;
+	bool has_volume;
+	pid_t hog_pid;
+	UInt32 buffer_size_restore;
+	
+	// Collect information from the device
```
MaxKellermann: What does this mean?

src/lib/coreaudio/CoreAudioDevice.hxx
```diff
@@ -0,0 +1,119 @@
...
+	// To identify devices where streams are single channels
+	bool is_planar;
+	AudioDeviceID device_id;
+	UInt32 output_stream_idx;
+	AudioStreamBasicDescription output_format;
+	AudioDeviceIOProc io_proc;
+	CoreAudioStream output_stream;
+	bool has_volume;
+	pid_t hog_pid;
+	UInt32 buffer_size_restore;
```
MaxKellermann: This name sounds rather important, and I guess this variable should better be documented.

src/lib/coreaudio/CoreAudioDevice.hxx
```diff
@@ -0,0 +1,119 @@
...
+	/** Scores a format based on:
+	 *	1. Matching sample rate (or integer fraction / multiple)
+	 *	2. Matching bits per channel (or higher).
+	 *	3. Matching number of channels (or higher).
+	 *
+	 *	The ASBD which should be scored against the
+	 *	MPD AudioFormat format which should be matched
+	 *	to be passed.
+	 */
+	float ScoreFormat(const AudioStreamBasicDescription &format_desc, const AudioFormat &format) const;
```
MaxKellermann: What does the return value mean?

src/lib/coreaudio/CoreAudioDevice.hxx
```diff
@@ -0,0 +1,119 @@
...
+	
+	/** Scores a samplerate based on:
+	 * 	1. Prefer exact match
+	 *	2. Prefer exact multiple of source samplerate
+	 *	The destination samplerate to score is compared against
+	 *	the requested source rate of the audio format.
+	 */
+	float ScoreSampleRate(Float64 destination_rate, unsigned int source_rate) const;
```
MaxKellermann: What does the return value mean?

src/lib/coreaudio/CoreAudioDevice.hxx
```diff
@@ -0,0 +1,119 @@
...
+	
+	typedef struct
```
MaxKellermann: No `typedef`!

src/lib/coreaudio/CoreAudioDevice.hxx
```diff
@@ -0,0 +1,119 @@
...
+	{
+		AudioStreamID stream_id;
+		StreamFormatList format_list;
+		UInt32 num_channels;
+	} stream_info;
```
MaxKellermann: Coding style violation: CamelCase for struct names.

src/lib/coreaudio/CoreAudioHardware.cxx
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+#include "CoreAudioHardware.hxx"
+#include "CoreAudioHelpers.hxx"
+#include "util/Domain.hxx"
+#include "Log.hxx"
+
+static constexpr Domain macos_output_domain("macos_output");
+
+AudioDeviceID CCoreAudioHardware::FindAudioDevice(const std::string &search_name)
```
MaxKellermann: Why `std::string`?

src/lib/coreaudio/CoreAudioHardware.cxx
```diff
@@ -0,0 +1,107 @@
...
+{
+	AudioDeviceID device_id = 0;
+
+	if (!search_name.length())
+		return device_id;
+
+	std::string search_name_lower = search_name;
+	std::transform(search_name_lower.begin(), search_name_lower.end(), search_name_lower.begin(), ::tolower );
```
MaxKellermann: `tolower` is not UTF-8-safe! Never use it!

src/lib/coreaudio/CoreAudioHardware.cxx
```diff
@@ -0,0 +1,107 @@
...
+	if (search_name_lower.compare("default") == 0)
```
MaxKellermann: `search_name_lower == "default"`

src/lib/coreaudio/CoreAudioHardware.cxx
```diff
@@ -0,0 +1,107 @@
...
+		return 0;
+	}
+
+	// Attempt to locate the requested device
+	std::string device_name;
+	for (size_t dev = 0; dev < device_count; dev++)
+	{
+		CoreAudioDevice device(device_list[dev]);
+		device_name = device.GetName();
+		std::transform(device_name.begin(), device_name.end(), device_name.begin(), ::tolower );
```
MaxKellermann: No `tolower`!

src/lib/coreaudio/CoreAudioHardware.cxx
```diff
@@ -0,0 +1,107 @@
...
+		if (search_name_lower.compare(device_name) == 0)
+			device_id = device_list[dev];
+		if (device_id)
+			break;
+	}
+	delete[] device_list;
```
MaxKellermann: Not exception-safe.

src/lib/coreaudio/CoreAudioHardware.hxx
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "CoreAudioDevice.hxx"
+
+// There is only one AudioSystemObject instance system-side.
+// Therefore, all CCoreAudioHardware methods are static
+class CCoreAudioHardware
```
MaxKellermann: Not a class. Just a collection of `static` methods. Don't declare it as a class if there will never be an instance.

src/lib/coreaudio/CoreAudioHardware.hxx
```diff
@@ -0,0 +1,29 @@
...
+{
+public:
+	static AudioDeviceID FindAudioDevice(const std::string &device_name);
```
MaxKellermann: Why `std::string`?

src/lib/coreaudio/CoreAudioHelpers.cxx
```diff
@@ -0,0 +1,261 @@
...
+
+static constexpr Domain macos_output_domain("macos_output");
+
+// Helper Functions
+std::string
+GetError(OSStatus error)
+{
+	char buffer[128];
+ 
+	*(UInt32 *)(buffer + 1) = CFSwapInt32HostToBig(error);
```
MaxKellermann: Alignment error.

src/lib/coreaudio/CoreAudioHelpers.cxx
```diff
@@ -0,0 +1,261 @@
...
+{
+	char fourCC[5] = {
+		(char)((desc.mFormatID >> 24) & 0xFF),
+		(char)((desc.mFormatID >> 16) & 0xFF),
+		(char)((desc.mFormatID >>  8) & 0xFF),
+		(char) (desc.mFormatID        & 0xFF),
+		0
+	};
+
+	std::stringstream sstr;
```
MaxKellermann: Please please don't use iostreams. They add huge amounts of overhead - a hugely bloated C++ misfeature.

src/lib/coreaudio/CoreAudioHelpers.cxx
```diff
@@ -0,0 +1,261 @@
...
+	return str.c_str();
+}
+
+AudioBufferList *
+AllocateABL(const AudioStreamBasicDescription asbd, const UInt32 capacity_frames)
+{
+	AudioBufferList *buffer_list = nullptr;
+	UInt32 num_buffers = (asbd.mFormatFlags & kAudioFormatFlagIsNonInterleaved) ? asbd.mChannelsPerFrame : 1;
+	
+	if((buffer_list = static_cast<AudioBufferList *>(calloc(1, offsetof(AudioBufferList, mBuffers) + (sizeof(AudioBuffer) * num_buffers)))) == nullptr)
```
MaxKellermann: No `calloc`!

src/lib/coreaudio/CoreAudioHelpers.cxx
```diff
@@ -0,0 +1,261 @@
...
+		throw std::runtime_error("Unable to allocate memory for AudioBufferList.");
```
MaxKellermann: `std::bad_alloc`? Now if you were using C++ instead of C, then you'd get this exception automatically.

src/lib/coreaudio/CoreAudioHelpers.cxx
```diff
@@ -0,0 +1,261 @@
...
+	}
+	return buffer_list;
+}
+
+void
+DeallocateABL(AudioBufferList *buffer_list)
+{
+	if(buffer_list != nullptr)
+	{
+		for(UInt32 buffer_index = 0; buffer_index < buffer_list->mNumberBuffers; ++buffer_index) {
```
MaxKellermann: Why `UInt32`? Forcing the compiler into one register size may reduce performance.

src/lib/coreaudio/CoreAudioHelpers.cxx
```diff
@@ -0,0 +1,261 @@
...
+			free(buffer_list->mBuffers[buffer_index].mData);
```
MaxKellermann: Don't use `free` in C++.

src/lib/coreaudio/CoreAudioHelpers.cxx
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+	free(buffer_list);
+}
+
+void
+AllocateAudioBuffer(AudioBuffer &buffer, const AudioStreamBasicDescription asbd, const UInt32 capacity_frames)
+{
+	if((buffer.mData = static_cast<void *>(calloc(capacity_frames, asbd.mBytesPerFrame))) == nullptr)
```
MaxKellermann: No `calloc` ...

src/lib/coreaudio/CoreAudioHelpers.cxx
```diff
@@ -0,0 +1,261 @@
...
+		case SampleFormat::S32:
+			out_format.mBitsPerChannel = 32;
+			out_format.mFormatFlags = kAudioFormatFlagIsSignedInteger;
+			break;
+		case SampleFormat::FLOAT:
+			out_format.mBitsPerChannel = 32;
+			out_format.mFormatFlags = kAudioFormatFlagIsFloat;
+			break;
+		default:
+			break;
```
MaxKellermann: Uninitialized memory in this case.

src/lib/coreaudio/CoreAudioStream.cxx
```diff
@@ -0,0 +1,293 @@
...
+
+#include "CoreAudioStream.hxx"
+#include "CoreAudioHelpers.hxx"
+#include "util/Domain.hxx"
+#include "Log.hxx"
+
+static constexpr Domain macos_output_domain("macos_output");
+
+CoreAudioStream::CoreAudioStream() :
+	stream_id  (0    )
```
MaxKellermann: Bad coding style

src/lib/coreaudio/CoreAudioStream.cxx
```diff
@@ -0,0 +1,293 @@
...
+
+bool CoreAudioStream::Open(AudioStreamID id)
+{
+	stream_id = id;
+	FormatDebug(macos_output_domain, "Opened stream 0x%04x.", (uint)stream_id);
+	
+	// Get original stream formats
+	if (!GetVirtualFormat(&original_virtual_fmt))
+	{
+		FormatError(macos_output_domain, "Unable to retrieve current virtual format for stream 0x%04x.", (uint)stream_id);
```
MaxKellermann: Throw exception on error!

src/lib/coreaudio/CoreAudioStream.cxx
```diff
@@ -0,0 +1,293 @@
...
+
+void CoreAudioStream::Close(bool restore)
+{
+	std::string format_string;
+	if (!stream_id)
+		return;
+
+	// remove the physical/virtual property listeners
+	AudioObjectPropertyAddress aopa;
+	aopa.mScope    = kAudioObjectPropertyScopeGlobal;
```
MaxKellermann: Bad coding style - whitespace art!

src/lib/coreaudio/CoreAudioStream.cxx
```diff
@@ -0,0 +1,293 @@
...
+	if (AudioObjectAddPropertyListener(stream_id, &aopa, HardwareStreamListener, this) != noErr)
+		FormatError(macos_output_domain, "Couldn't set up a virtual property listener for Core Audio stream.");
+
+	return true;
+}
+
+
+void CoreAudioStream::Close(bool restore)
+{
+	std::string format_string;
```
MaxKellermann: What is this? Why is it here?

src/lib/coreaudio/CoreAudioStream.cxx
```diff
@@ -0,0 +1,293 @@
...
+	AudioObjectPropertyAddress aopa;
+	aopa.mScope    = kAudioObjectPropertyScopeGlobal;
+	aopa.mElement  = kAudioObjectPropertyElementMaster;
+	aopa.mSelector = kAudioStreamPropertyVirtualFormat;
+
+	UInt32 property_size = sizeof(AudioStreamBasicDescription);
+	OSStatus ret = AudioObjectSetPropertyData(stream_id, &aopa, 0, NULL, property_size, desc);
+	if (ret)
+	{
+		FormatError(macos_output_domain, "Unable to set virtual format for stream 0x%04x. Error = %s",
```
MaxKellermann: Throw exception!!

src/lib/coreaudio/CoreAudioStream.hxx
```diff
@@ -0,0 +1,57 @@
...
+#include <vector>
+
+typedef std::vector<AudioStreamID> AudioStreamIdList;
+typedef std::vector<AudioStreamRangedDescription> StreamFormatList;
+
+class CoreAudioStream
+{
+public:
+	CoreAudioStream();
+	virtual ~CoreAudioStream();
```
MaxKellermann: Why `virtual`?

src/output/plugins/MacOSOutputPlugin.cxx
```diff
@@ -0,0 +1,553 @@
...
+#include "util/StringBuffer.hxx"
+#include "pcm/PcmExport.hxx"
+#include <AudioToolbox/AudioConverter.h>
+#include "Log.hxx"
+#include "lib/coreaudio/CoreAudioDevice.hxx"
+#include "lib/coreaudio/CoreAudioHelpers.hxx"
+#include <boost/lockfree/spsc_queue.hpp>
+
+static constexpr Domain macos_output_domain("macos_output");
+// Set output frame buffer to double of the 512 default value
```
MaxKellermann: What kind of default value is this? And why double?

src/output/plugins/MacOSOutputPlugin.cxx
```diff
@@ -0,0 +1,553 @@
...
+// Ring buffer of at least 100ms
+static constexpr unsigned BUFFER_TIME_MS = 100;
+
+struct MacOSOutput final : AudioOutput {
+	
+	// Layer around CoreAudio
+	CoreAudioDevice device;
+
+	// Plugin settings as fetched from config
+	std::string device_name;
```
MaxKellermann: Why `std::string`?

src/output/plugins/MacOSOutputPlugin.cxx
```diff
@@ -0,0 +1,553 @@
...
+#endif
+	
+	pcm_export->Open(audio_format.format, audio_format.channels, params);
+	/** Setup the ring_buffer to hold
+	 *	BUFFER_TIME_MS or four times
+	 *	the device frame buffer.
+	 */
+	UInt32 dev_frame_buffer = device.GetBufferSize();
+	if(dev_frame_buffer == 0)
+		throw std::runtime_error("Cannot get valid frame buffer size from device.");
```
MaxKellermann: Leaks memory

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: I assume you would like to have char * instead, right? Can change all the instances accordingly.

MaxKellermann: Yes, I prefer plain C strings unless there is a good reason to use `std::string` with all its overhead. So I asked for your motivation.

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Will change everywhere

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: No motivation in particular other than convenience.

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Adjusted

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Adjusted

src/lib/coreaudio/CoreAudioDevice.hxx
coroner21: No point. Will remove virtual

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Okay, using new instead.

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Done

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Added try / catch blocks.

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: added

src/lib/coreaudio/CoreAudioDevice.hxx
coroner21: already resolved

src/lib/coreaudio/CoreAudioDevice.hxx
coroner21: Added more verbose comment

src/lib/coreaudio/CoreAudioDevice.hxx
coroner21: Added some more comments.

src/lib/coreaudio/CoreAudioDevice.hxx
coroner21: Added comment on return value.

src/lib/coreaudio/CoreAudioDevice.hxx
coroner21: Added comment.

src/lib/coreaudio/CoreAudioDevice.hxx
coroner21: Removed typedef

src/lib/coreaudio/CoreAudioDevice.hxx
coroner21: adjusted

src/lib/coreaudio/CoreAudioHardware.cxx
coroner21: changed to char*

src/lib/coreaudio/CoreAudioHardware.cxx
coroner21: Removed, just made device names case-sensitive

src/lib/coreaudio/CoreAudioHardware.cxx
coroner21: working with C string now, therefore strcmp

src/lib/coreaudio/CoreAudioHardware.cxx
coroner21: removed

src/lib/coreaudio/CoreAudioHardware.hxx
coroner21: Removed this completely (instead copying the methods to CoreAudioHelpers)

src/lib/coreaudio/CoreAudioHardware.hxx
coroner21: changed to char*

src/lib/coreaudio/CoreAudioHelpers.cxx
coroner21: This already considers all possible formats (except UNKNOWN and DSD which will be caught by the assertion). Added gcc_unreachable here.

src/lib/coreaudio/CoreAudioStream.hxx
coroner21: removed

src/output/plugins/MacOSOutputPlugin.cxx
coroner21: changed to char *

src/output/plugins/MacOSOutputPlugin.cxx
coroner21: The default value is 512 frames which means that mpd will be called for new input every 512 frames. The double 1024 default value uses a lot less CPU as larger chunks are passed for each callback.

src/lib/coreaudio/CoreAudioHelpers.cxx
coroner21: Changed this according to https://developer.apple.com/library/archive/samplecode/CoreAudioUtilityClasses/Listings/CoreAudio_PublicUtility_CAXException_h.html

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Corrected

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: This method now throws an exception whenever no valid name can be returned

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Added try / catch block

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Yes, in general I changed a lot of the code to just throw instead of only printing errors

src/lib/coreaudio/CoreAudioHardware.cxx
coroner21: fixed

src/lib/coreaudio/CoreAudioHelpers.cxx
coroner21: Replaced with operator new

src/lib/coreaudio/CoreAudioHelpers.cxx
coroner21: fixed

src/lib/coreaudio/CoreAudioHelpers.cxx
coroner21: for loops changed to unsigned int

src/lib/coreaudio/CoreAudioHelpers.cxx
coroner21: replaced with delete

src/lib/coreaudio/CoreAudioHelpers.cxx
coroner21: new used instead

src/output/plugins/MacOSOutputPlugin.cxx
coroner21: How so? Apart from memory allocated within the device class that should be freed when the output is closed, nothing was allocated.

MaxKellermann: `PcmExport::Open()` may have allocated memory.

src/lib/coreaudio/CoreAudioStream.cxx
coroner21: Changed this to void function throwing exception on error

src/lib/coreaudio/CoreAudioStream.cxx
coroner21: corrected

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Adjusted

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Done

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Done

src/lib/coreaudio/CoreAudioHelpers.cxx
coroner21: Adjusted, using the StringBuffer / StringFormat methods that are also used to print AudioFormat

src/lib/coreaudio/CoreAudioStream.cxx
coroner21: adjusted

src/lib/coreaudio/CoreAudioStream.cxx
coroner21: Done

src/lib/coreaudio/CoreAudioStream.cxx
coroner21: Removed. Was used to pass into FormatToString conversion method, not needed with StringBuffer approach apparently.

src/output/plugins/MacOSOutputPlugin.cxx
coroner21: OK, moved `PcmExport::Open()` to end of the block.

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Done

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Need to stick with traditional for loop as I need an index (count of the element in the vector)

src/lib/coreaudio/CoreAudioHelpers.cxx
coroner21: OK, added the two additional elements of the SampleFormat enum

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: OK, if hog mode cannot be set it means that some other program already owns hog mode which is in any case fatal for using the device further...

coroner21:
Alright, I think I handled all of the issues and did another review by myself including some testing on my devices. @MaxKellermann: Can you let me know how to proceed? I would propose to close this PR and open a new clean one based on a direct diff against master. Do you want this to be a drop-in replacement? I would still prefer to have the old plugin as a fallback at the moment.

MaxKellermann:
Just force-push your new commit into this PR (i.e. in your branch "master"). We don't need to have a new PR; this PR documents the story of this new plugin.
I don't think it's useful to have two plugins doing the same thing with the same system API. Nobody will use your new plugin this way, and so nobody will report bugs.
So if your new plugin is better, there's no reason to keep the old one.
But I already stated that I don't like your general approach, by writing a new plugin from scratch instead of improving the old plugin. But it's too late to complain about this - you spent a lot of time with this code, and I hope you'll take care for your code and support it (the old code isn't maintained/supported by anybody). We shouldn't go back from here, and so we should merge it once we are confident that it's really better.
So yes, the old plugin should go away, and the new one should replace it automatically. Existing old configurations should continue to work properly - maybe better, but never worse.

Once your new code is up, let's check it again, and focus on compatibility with the old plugin.

In any case, thanks for your diligent work. It's good to finally have a maintainer for the macOS code. You've left the "writing new code is exciting" stage behind, and you're in the "maintaining existing code is sometimes boring" stage. I've seen too many people give up long before reaching that second stage, and unfortunately their new code is worthless that way.

coroner21:
Thanks. Hope the force push was successful. As I said, this still has some drawbacks in comparison with the old plugin but should be an improvement for 99% of the use cases.

To make further improvements later, could you let me know if there is a way to:
- reconfigure the output from within the output plugin (i.e. trigger a re-open of the output)?
- check on close whether it was triggered by a format change (I do not want to close the stream in most of the cases which takes a lot of time in integer mode) or by the 'stop' command?

I would be happy if you could point me to some existing code that does something similar already.

MaxKellermann:
Both is not possible.
"reconfigure the output from within the output plugin" - why would you want to do that? Does the "client" (i.e. the MPD core) need to be aware that the output plugin does something which it considers "reopening itself" internally? Or do you want to renegotiate the audio format with the MPD core after a hardware change?
"check on close whether it was triggered by a format change" - I can understand the need for that, but the plugin API isn't capable of doing that. There should be a method which, if implemented, attempts to change the audio format on-the-fly, but such a method doesn't exist yet. But you're welcome to send a PR with an API proposal (virtual method + documentation), and once we agree on the design, I'll merge it, and you can implement your part, which will be unused until I hook it up in the MPD core (and actually call it).

coroner21:
For the first point: Yes it is basically hardware changes. I would like to reconfigure the output from within the output plugin if the default output device changes (e.g. plugging in the external sound card). CoreAudio API is callback based therefore you are supposed to listen to a lot of properties and react accordingly. Also plugging in headphones apparently triggers a sample rate change (I do not know why) which can only be dealt with by detecting this and then re-configuring the output. The old code using AudioUnit was able to handle such situations since the AudioUnit just reconfigures itself on the fly (output is then not bit perfect any more of course). With direct device access and integer mode more work is needed.

Regarding second point: I will think about an additional "ReOpen" method to extend the output plugin API.

src/mixer/plugins/MacOSMixerPlugin.cxx
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	int GetVolume() override;
+	void SetVolume(unsigned volume) override;
+};
+
+int
+MacOSMixer::GetVolume()
+{
+	return macos_output_get_volume(output);
```
Wang-Yue: It's better to move get/set volume code to here rather than letting the output plugin handle it.

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,614 @@
...
+void
+CoreAudioDevice::Open(const char *device_name) {
+	
+	device_id = FindAudioDevice(device_name);
+	Enumerate();
+
+	AudioObjectPropertyAddress aopa;
+	aopa.mScope = kAudioDevicePropertyScopeOutput;
+	aopa.mElement	= 0;
+	aopa.mSelector = kAudioHardwareServiceDeviceProperty_VirtualMasterVolume;
```
Wang-Yue: don't use audio toolbox API. Instead, refer to https://github.com/MusicPlayerDaemon/MPD/pull/306 and create a hardware mixer.

src/lib/coreaudio/CoreAudioDevice.cxx
```diff
@@ -0,0 +1,614 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "CoreAudioDevice.hxx"
+#include "CoreAudioHelpers.hxx"
+#include "util/RuntimeError.hxx"
+#include <AudioToolbox/AudioToolbox.h>
```
Wang-Yue: remove. Use 
        AudioObjectPropertyAddress aopa = {
                .mSelector      = kAudioDevicePropertyVolumeScalar,
                .mScope         = kAudioObjectPropertyScopeOutput,
                .mElement       = kAudioObjectPropertyElementMaster,
        };

to set device volume.

src/output/plugins/MacOSOutputPlugin.cxx
```diff
@@ -0,0 +1,523 @@
...
+	catch (...) {
+		AudioConverterDispose(ca_converter);
+		ca_converter = nullptr;
+		throw;
+	}
+}
+
+void
+MacOSOutput::Setup(AudioFormat &audio_format) {
+	if (device.SetFormat(audio_format, integer_mode)) {
```
Wang-Yue: Can this plugin sync hardware sample rate / bit depth according to the current output format?  Old plugin has that functionality. 

How can user opt in/out this feature?

src/lib/coreaudio/CoreAudioStream.cxx
```diff
@@ -0,0 +1,226 @@
...
+
+void
+CoreAudioStream::SetPhysicalFormat(AudioStreamBasicDescription desc) {
+	if (!stream_id)
+		return;
+
+	AudioObjectPropertyAddress aopa;
+	aopa.mScope = kAudioObjectPropertyScopeGlobal;
+	aopa.mElement = kAudioObjectPropertyElementMaster;
+	aopa.mSelector = kAudioStreamPropertyPhysicalFormat;
```
Wang-Yue: please keep in mind that setting physical format is time consuming. System/DACs may skipped a few samples if you set it. That's part of the reason I deleted relevant code for cmus https://github.com/cmus/cmus/pull/456/commits/995e0673522e191628e9319be6d553f116c21a48 a while ago. 

If you want to have this feature, please make it an opt-in feature.

src/output/plugins/MacOSOutputPlugin.cxx
```diff
@@ -0,0 +1,523 @@
...
+	AudioStreamBasicDescription in_format;
+	// The format CoreAudio requests for IO
+	AudioStreamBasicDescription out_format;
+
+	bool pause;
+	
+	/** Required to support DoP. No
+	 *	other features currently used.
+	 */
+	Manual<PcmExport> pcm_export;
```
Wang-Yue: If it's a DoP only feature, guard it behind a flag. Also hide it within a if (dop) switch. 
Otherwise the code will be running for all audio renderer loops. 
It's always a good idea to optimize the audio renderer loop and make it do as few things as possible.

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Unfortunately the property you used in #306 does not work on my Built-In Output Device. The above has the nice feature that it simply uses either the volume property of each channel in sync or the master device volume if available.

src/mixer/plugins/MacOSMixerPlugin.cxx
coroner21: Why? This mixer plugin is more or less an exact copy of the old OSXOutput mixer plugin. Please also note that apparently in #306 you saw no need to change this...

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: See above, this does not work with all devices

src/output/plugins/MacOSOutputPlugin.cxx
coroner21: This runs also by default in the alsa output and I see no need to have a complex ´if´ (or `#ifdef`) logic around it for using either PcmExport or direct chunk copying. Also note that this runs inside the Play method but not in the audio render callback.

src/mixer/plugins/MacOSMixerPlugin.cxx
Wang-Yue: #306 was my first change, and there will be another coming to get the logic into the mixer plugin. 

I concede Old OSXOutput doesn't have it, but it's not the right thing to do. If I were the original author of old plugin, I wouldn't make it this way. 

Mixer functionality/code should live in the mixer, not the output plugin. It's why they are separate classes

src/lib/coreaudio/CoreAudioDevice.cxx
Wang-Yue: #306 works on my built in output device flawlessly. 

I tested the code on both my Mac Pro and Macbook Pro, with both cmus and mpd running the same code.

src/lib/coreaudio/CoreAudioDevice.cxx
Wang-Yue: See above.

src/output/plugins/MacOSOutputPlugin.cxx
Wang-Yue: Got it. Thanks.

src/lib/coreaudio/CoreAudioStream.cxx
Wang-Yue: You can play some music with any music player app you have, and go to "Audio Midi Setup" app, and change your DAC setting for sample rate / bit depth. You can clearly hear the hiccup when you change it.

src/output/plugins/MacOSOutputPlugin.cxx
Wang-Yue: ALSA has that running doesn't mean it's mandatory for your plugin to do so. 
The first priority in programming an audio render loop is to make sure it consumes as little CPU resources as possible, and avoid any hard to predict latency.

src/mixer/plugins/MacOSMixerPlugin.cxx
coroner21: Understood. However the mixer plugin needs to get the device anyways from the output therefore there can never be real separation. Not sure if it is worth the additional effort.

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: Can have another try but I remember being disappointed that the property was a no-op on my MacBook Pro.

src/lib/coreaudio/CoreAudioStream.cxx
coroner21: Well, this is clear since you are then changing the format while music is playing. This will never be done by this plugin though. Sample rate / bit depth is only changed in beginning of playback of a new song which has a different format than the previous one.

src/lib/coreaudio/CoreAudioDevice.cxx
Wang-Yue: could you check in #306 or build https://github.com/cmus/cmus and see if the code works for your Built-in Output? if either one works for you, probably there's an issue with your code.

src/output/plugins/MacOSOutputPlugin.cxx
```diff
@@ -0,0 +1,523 @@
...
+		 *	to make sure that MPDs output engine sends
+		 *	the physical format. This will get transformed
+		 *	to the virtual format in final conversion
+		 *	directly in the render callback (to float
+		 *	samples in case integer_mode is not
+		 *	configured or not supported).
+		 */
+		
+		memset(&in_format, 0, sizeof(in_format));
+		memset(&out_format, 0, sizeof(out_format));
```
Wang-Yue: why memset is needed? in/out_format will be completely copied from device.GetPhys/IOFormat

src/output/plugins/MacOSOutputPlugin.cxx
```diff
@@ -0,0 +1,523 @@
...
+	 */
+	UInt32 frames = requested / oo->out_format.mBytesPerFrame;
+	// Number of bytes to pop from ring_buffer (MPD format framesize times number of frames)
+	UInt32 in_bytes = std::min(frames * oo->in_format.mBytesPerFrame, (UInt32) oo->ring_buffer->read_available());
+	UInt32 available_frames = in_bytes / oo->in_format.mBytesPerFrame;
+	
+	if(available_frames < frames)
+		FormatDebug(macos_output_domain, "Frames available (%d) less than requested (%d) by device.", available_frames, frames);
+	
+	// Usage of converter
```
Wang-Yue: I don't like the idea of using a converter. If you let AudioUnit to handle the stream, it will automatically convert the format for you. The conversion is transparent --- if your device's output format is identical to the file, it will just copy the bytes. 

I don't know if it's a better approach here to do the conversion by yourself and manipulate various objects (such as virtual format). The original code can do pretty much everything you are trying to achieve here  --- #271  could be easily solved by apply a patch that's similar to the reversion of https://github.com/cmus/cmus/pull/456

Wang-Yue:
I don't quite understand the motivation behind this PR. 

> This is a stab at #271 adding all the API calls required to change bit depth and sample rate

Well, as explained in #271, it is quite simple to add the bit depth synchronization back to MPD. One just need to apply a patch that's similar to the reversion of cmus/cmus#456

> It has the additional feature of integer mode: Passing directly SInt samples to the device which is a desired feature by audiophiles

Oh audiophiles --- Are those the people who can tell a difference of audio quality between using thermo power and nuclear power?

For PCM wav form, Integer and Float are identical --- One wouldn't gain audio quality by switching from Float to Int. 

Most lossless audio format (such as FLAC) only support Float --- the flac callback only outputs 32bit float, no matter what the source wav format is. So not many people will use this feature. 

Even if the source format is wav form in Integer, CoreAudio use 32bit Float under the hood, and provides far better precision than 16/24 bit integer. so conversion is transparent. 

I don't think we gained anything in supporting this weird feature. On the contrary, it makes the code very complex and hard to support. 

> It directly accesses the devices buffers in the callback so no overhead from AudioUnit framework

AudioUnit (which is a simple abstraction) is very transparent and efficient. The CMUS CoreAudio plugin has similar design as the old plugin, and I benchmarked it last year, and FLAC decoding and bit conversion (in case you are interested, the code is here: https://github.com/cmus/cmus/blob/master/ip/flac.c#L181 . MPD has a more performant implementation thanks to C++ templates, which CMUS cannot use) take much much more CPU than AudioUnit overhead. I would like to see your benchmark numbers before calling your approach an optimization. 

Also I would recommend you to take a look at the original osx_renderer function here 

https://github.com/MusicPlayerDaemon/MPD/blob/acd61a0e8ecfa5504381361b0616baeeeaf0a594/src/output/plugins/OSXOutputPlugin.cxx#L605

and compare with yours. 

The original code is far simpler, and easier to understand and maintain. Unless you are willing to maintain your code indefinitely, I don't think it's a better approach to take.

MaxKellermann:
Thanks @Wang-Yue for joining the discussion, and of course thanks @coroner21 for spending time on writing MPD code. This is a hard one, and I have difficulties deciding what to do from here. In any case, I'm happy that there's a now a big chance to improve MPD's macOS support, after a long time of stagnancy.

I know nearly nothing about macOS, which means I can complain about coding style and general concepts, but not about why/how/whether the new plugin is better. As I already said at the beginning, I prefer incremental improvements to old code instead of a rewrite from scratch, because a rewrite can hardly be reviewed; the new code is an opaque block for me.

Is the new plugin better? No idea. I can't "diff" it to see improvements. That worries me.
There may be some things that are better, but these could have been added to the old plugin as @Wang-Yue already wrote. Reviewing such an improvement would be easy, and the nature of the improvement would be obvious enough. But not with this PR.

From anecdotal experience, I know that rewrites cause a lot of pain. Even if old code is bad, and it sometimes takes more time to improve it incrementally, the result in the end is always better. Unfortunately, writing new code is exciting, and improving old code is boring...

I don't know where to go from here. There's the conflict between "this PR is not how I prefer it to be done" and "@coroner21 wants to contribute, and I should make that possible, and avoid frustrating him after spending so much time".
One thing is sure: there will not be two competing plugins for one system API. If the new plugin is merged, we must be sure (as sure as we can be) that it is better, has all old features, automatic transition of all users without any fallout, no regressions.
And I think it is easier to "port" the improvements from this new plugin to the old plugin, one by one, incrementally. Even if that appears to be unnecessary additional work - it is an easy way to make sure that there are no regressions, because all changes can be reviewed easily. Making sure that the old code is mature enough is a huge task compared to that.

My final blurp about audio formats: MPD is targeted at audiophiles, but not at superstitious audiophiles. If @Wang-Yue is right and this "integer mode" is useless (because everything's going to be converted to float anyway), then this integer mode is not worth implementing/supporting. We don't need a feature checklist full of useless features just to make superstitious people happy. They're not MPD's target audience.
(I'm not completely opposed to having integer mode, and I'd accept a PR implementing it. But I consider it a worthless feature, which will be removed if it causes problems which are not be fixed in a timely manner.)
But MPD should (by default) do its best to retain the audio format from the input file and push it bit-by-bit through the whole chain up until the DAC. Deviate from this policy only if the user has asked to do so (e.g. by configuring a `format`). (I don't like how the old plugin violates this policy by turning it around, and I hope we're about to fix that.)

src/output/plugins/MacOSOutputPlugin.cxx
coroner21: Thanks, makes sense to delete the memset calls indeed!

src/lib/coreaudio/CoreAudioDevice.cxx
coroner21: I tried it again. Not every device has a master volume and my Built-in Output of a 2012 MacBook Pro 15" (non-retina) unfortunately has none. **Which means that #306 will not work for some users.**

I agree though that AudioToolbox API just for setting the volume might be overkill. This should therefore probably be redesigned to just loop over all (output) channels of the device and set the volume separately.

Wang-Yue: You probably can't set the master volume using kAudioObjectPropertyElementMaster for all device. Can you try setting the channels separately? Something like 

		AudioObjectPropertyAddress aopa = {
			.mSelector	= kAudioDevicePropertyVolumeScalar,
			.mScope		= kAudioObjectPropertyScopeOutput,
			.mElement	= coreaudio_stero_channels[i]
		};

in https://github.com/cmus/cmus/blob/ef4ca2a2e3c677305d6af891aba5a82c8bcca566/op/coreaudio.c#L754 should work. I've never heard complaints  from user with that piece of code.

MaxKellermann:
@coroner21 I understand your frustration, but please also understand that your approach of rewriting everything from scratch was a pretty poor choice. My first reply to this PR was:
> commit message should point out why this rewrite was done. What problems does it solve? How is it better? Why did you do a rewrite from scratch instead of improving the old one? Complete rewrites are almost always a bad idea.

You didn't reply to that. And so I still don't know!

Your rewrite what really caused unnecessary duplicate effort, because the effort of writing an output plugin for macOS had been done already years ago.
What @Wang-Yue did was pick up the some improvements of your plugin and submit minimal changes to the old plugin. That's good, I like that! Of course I would have preferred to have these small PRs earlier, before you started to work.

And of course I happily merge @Wang-Yue's work. I understand those PRs well enough to say: that's good, that's an improvement. And it doesn't look "desperate" at all, and we don't need a "conclusion" for this PR before I merge improvements for the old plugin. If @Wang-Yue likes to improve the old one, that should be done, as long as the old one exists.

And you know my preference: I'd prefer to keep the old one, and see it improved. Maybe up to the point where it eventually transforms completely into the code which is now your new plugin. During this transformation, the incremental changes can be reviewed, and mistakes can be found more easily. It is likely that the final code will look different from your current code, because we'll learn from the mistakes on the way.

This is the point where hacking open source can seem frustrating - I know how this feels. But please try to push the frustration away, and let's find a way to turn this into something constructive. Sometimes, one has to take many detours until one finds the best solution/approach for a problem. This may seem like wasting time, but it's sometimes an inevitable part of our work, and frustration is also sometimes inevitable.

About the choice of system API (CoreAudio vs. AudioUnit) and Integer Mode: for me, it's ok to migrate to a more complicated lower-level API and add complexity to support things like Integer Mode, if an advantage of doing so can be explained, and a maintainer volunteers for supporting the new code.
For example, I've recently converted the ALSA plugin to asynchronous (https://github.com/MusicPlayerDaemon/MPD/commit/b1c7649edb80285c113f4710dd4c19d8f7c1aef7). This adds quite some complexity, but has serious advantages on the other hand.

coroner21:
@MaxKellermann, thank you for your understanding and I can only say that I am trying to be as constructive as possible here. I think it is already constructive to spend hours of rare free time to work on this and clearly I did this hoping that it benefits MPD and its users (me being one of them) and not for having long discussions about the approach and why and how something is better or worse.

> You didn't reply to that. And so I still don't know!

I thought I provided in the meantime a lot of examples why I think this is an improvement over the old plugin, but let me repeat all of them condensed together:
- This is a stab at #271 adding all the API calls required to change bit depth and sample rate
- It has the additional feature of integer mode: Passing directly signed Integer samples to the device which is a desired feature by audiophiles
- It directly accesses the devices buffers in the callback so no overhead from AudioUnit framework
- It has class interfaces for the main objects used (CoreAudio device, streams of the device) and a meaningful separation of code into different files
- It introduces some useful helper functions (converting MPD sample format to CoreAudio ASBD and back, printing out ASBD formats, printing CoreAudio error codes)
- It is fully compatible with the existing plugin

Does that justify a rewrite with a big bang instead of small improvements? I think some of the things above can only be achieved going for a more complete overhaul than just gradual improvements. As I said for me the decision was easy since officially the old code was unmaintained and unsupported. In creating this I sticked to the existing plugin as much as possible (simply copying all files renaming them to MacOS since this is how the OS is called now). So I did not start from scratch but completely used the available code whenever appropriate. So much so that @Wang-Yue is now complaining that I did not change the layout of the mixer code which is a valid point. Also I took over a lot from Kodi (whose CoreAudio plugin has many similarities with this PR).

Why did I do this at all? Simple: I like MPD very much (it is my main program used for serious music listening) and I think the OpenSource concept is great in the sense that if you want to get something done / if you want a feature to be available then you can simply implement it and in that way even contribute back to the community. Before, I wanted to be able to output DSD via DoP method on MacOS and implemented that into the existing plugin. Here, I wanted support for Integer Mode which is available by other software on MacOS aimed at audiophiles.

Now that this is done any time spent on the existing plugin seems a waste of time for me. As I said this time would be better used ironing out the kinks of this new code.

So where do we go from here? I take from the comments that there are some things that still can be worked on:
- Setting the volume should be done using the CoreAudio API directly instead of AudioToolbox
- Mixer code should be moved to mixer plugin
- If it helps, an additional render path using AudioUnit could be implemented
I am happy to take additional feedback and improve this further. And clearly if it is ever merged I expect to receive even more feedback since eventually more users test it (as of now not even @Wang-Yue tested this I guess). I will however only continue here if this might be merged eventually. Step by step moving over parts of the features / code to the old plugin is something I will not do as it means redoing all of the work.

Wang-Yue:
while you could discuss more about your design choices, one thing I want to say is I did tested your code --- apart from a linking problem that could be easily solved by adding "--framework Foundation", everything seems to run well. 

with that said, I still don't understand the need for integer mode --- Yes there's extra conversion for using AU, but 16 or 24 bit integer to 32 bit float conversion is just bit operation, as 32bit integer has 23 bit integer width and 1 bit sign. Such conversion is exactly like what MPD does in a lot of situation already. For instance, MPD  pack all 24 bit data inside a 32 bit size (S24_P32), similar to all decode library such as flac does. If you really wish to eliminate overhead there's a lot more you need to change. 

for 16 bit samples, extend to 24 bit is always encouraged as it provides transparent volume control without losing precision. thus, sending the device exactly the same bits as original is only a good evaluation of audio path transparency, but not a useful thing to do.

coroner21:
Thanks @Wang-Yue, which compiler are you using? With clang I did not have to include this linker option. Should probably then add it in general to be on the safe side.

With regards to integer mode: To be honest I am not even convinced myself that it is needed for anything. Currently I mainly listen mainly with having it disabled because the mode changes just take so long (unfortunately two virtual mode changes required due to lack of a method to signal the output plugin that it is only closed for audio format change).

Still, it is something that Apple offers in CoreAudio API and all the audiophiles apparently cannot sleep well in case they are uncertain whether the music might be taking any detour from hard drive to DAC. In general it is just more direct playback even if all conversions involved are bit perfect. If you look around on the internet you find a lot of people that seem to care about this feature and I think it is also some kind of marketing opportunity for MPD (now that it also has DSD under macOS). I would hope that integer mode convinces some users to ditch roon, Audirvana or JRiver or whatever proprietary solution (iTunes :-;) they use and switch to MPD instead.

Wang-Yue:
I'm using the clang bundled in latest Xcode and latest macOS.

```
wangyue-macpro:~ wangyue$ clang --version
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

and it was complained that unable to link because CFRelease() and so on is missing. 

also is it possible for you to create a abstract factory so user can opt in to use AU or the low level way?

MaxKellermann:
> Does that justify a rewrite with a big bang instead of small improvements?

No. All of this could have done with incremental improvements to the existing code. Of course, a migration from AudioUnit to CoreAudio would be a larger patch, but I would see only the necessary changes in the diff, and all those unchanged lines of boilerplate code would not have to be reviewed.

Why do you believe this justifies a rewrite? This is a serious question, because I really have no idea why you believe anything justifies a rewrite. And apparently on the other hand you cannot imagine why I believe a rewrite is a bad idea.

coroner21:
@MaxKellermann, so this is what being constructive means for you? Fine, let me give you some more background even though this seems like a big waste of time and arguing just for the arguments sake since we are not making any progress:
- I went with this approach also because I thought it would be a good idea to keep the old code around as a fall back for people that might experience issues with the new code
- I wanted to rename the plugin completely because OSX was renamed

Iterative approach might be easier for you to review and assess, but it would have been a lot harder for me. Making sure that the plugin always is fully functional while transforming it in little increments from the old to the new code is in my view very challenging and given the status of the old code (I say it again: officially unmaintained and unsupported) was not worth the effort for me.

Please think about some of the past increments that you just merged for the old code:
- My pause implementation caused CPU load to go to 100% (an issue I caught while testing this rewrite)
- @WangYue‘s volume patch now makes volume setting impossible for various users (depending on their sound card driver)

How did it help here that these were small increments and you could review them? Why should the risk of issues being introduced with this rewrite be a lot higher?

I just want to help and do the right thing here. You can be sure that I keep working on this and will be fixing problems in case we get any user feedback. Either way, we need to make progress in my view so please let me know how to proceed!

Wang-Yue:
No, My patch use the audiotoolbox, so it's equivalent as yours. It might be better to follow cmus's implementation (set both channels' volume using core audio), however.

Wang-Yue:
btw, if you can implement to support Native DSD (doable on Mac OS X, as proofed by Decibel), I encourage a complete rewrite. Now functionality wise your code is almost equivalent as the old one.

coroner21:
@Wang-Yue, yes, apparently now after you changed it in some other unrelated pull request which btw is not really transparent. In the original patch that @MaxKellermann merged you used the master volume scalar that is not available for many devices.

coroner21:
Regarding native DSD and Decibel: They support this for one DAC (some exasound device) with a proprietary API (ASIO) which is nothing that would make sense for MPD.

coroner21:
@MaxKellermann, I would like to revisit this and either close it or make sure it will get merged eventually. Could you let me know your conclusion?
And no, I will not work on the old plugin in little increments. I would give it one more shot and rebase this on the current master and make sure it still works, but if you do not like it then please simply close this pull request.

MaxKellermann:
I'm unable to review this big code dump.
Sad to hear that you've ultimately decided not to refactor your code dump into something which can be reviewed.
This is a hard decision for me, because you put a lot of time into this PR (even after I told you this was the wrong way of doing this). On the other hand, you never cared for the migration path and what happens with the old plugin.
Drive-by contributions of small, incremental patches are ok, I can merge that easily - but with a code dump like yours, there needs to be long-term maintainership by somebody with an eye for the big picture. That is more important than the actual code. A big improvement has a negative value if there are regressions, and those regression aren't fixed quickly.
I've seen this happen so many times - somebody drops code, looks good, but turns out to be flawed (of course, every new chunk of code is bugged). That somebody's gone by now, and I have to get my hands dirty.
Your decision leaves me no good choice, and my only remaining acceptable choice is to close this PR.

PR #308, opened 2018-07-13
Enable OSX to set both hardware sample rate and bit depth 

### LINKED ISSUE #271: osx sync_sample_rate does not sync bit depth

Wang-Yue:
This PR will fix #271.

special thanks to @coroner21 who contributed a nice way to score hardware supported format in #292

Also, The DSD related code are all guarded with ENABLE_DSD  flag.

### DIFF
--- a/doc/user.xml
+++ b/doc/user.xml
@@ -4440,18 +4440,6 @@ run</programlisting>
                   output device it is playing through, and no other program can access it.
                 </entry>
               </row>
-              <row>
-                <entry>
-                  <varname>sync_sample_rate</varname>
-                  <parameter>yes|no</parameter>
-                </entry>
-                <entry>
-                  Synchronize the sample rate. It will try to set the output device sample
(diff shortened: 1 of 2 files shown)


MaxKellermann:
Commit messages like "Update OSXOutputPlugin.cxx" look very bad in the shortlog, and please don't do merge commits in PRs - I merged it now, but please watch out next time.

PR #310, opened 2018-07-15
Add an option to OSXPlugin to allow user set the buffer time.

Wang-Yue:


coroner21:
Microseconds and milliseconds are not the same. Setting buffer_time does not really make much of a difference since it just changes how often MPD pushes samples into the output but does not have any effect on the number of times the CoreAudio API calls the callback.

Wang-Yue:
> Microseconds and milliseconds are not the same.

Fixed. I must be sleeping while writing that code

> Setting buffer_time does not really make much of a difference since it just changes how often MPD pushes samples into the output but does not have any effect on the number of times the CoreAudio API calls the callback.

Well, I'm not trying to use it this way. On the contrary,  a small enough buffer space is encouraged as it reduces latency. For instance, you may want to seek to another position in your track. With smaller buffer space it can seek much faster, as there're less old bytes to process in the buffer. That's why having it as an option is useful. For me 0.1s is too long.

MaxKellermann:
I'm not comfortable with this setting; I'm not comfortable with the existing setting in the ALSA plugin, and I'm not comfortable with the existing global audio buffer size settings. Users do stupid things with those settings, even though none of them ever had to change them. Usually, changing the defaults results in worse performance, because the default values are just fine.
Note that the setting in the ALSA plugin controls ALSA's (hardware) buffer, not MPD's ring buffer. This may cause some more confusion.
Formal complaint: please use `std::chrono` to avoid obvious mistakes like the one in your first version.
And:
> On the contrary, a small enough buffer space is encouraged as it reduces latency. For instance, you may want to seek to another position in your track. With smaller buffer space it can seek much faster, as there're less old bytes to process in the buffer

This is wrong. Seeking will cause a call to `Cancel()` which clears the ring buffer. In the ALSA plugin, this clears both the internal ring buffer and ALSA's hardware buffer.
In the ALSA plugin, a larger buffer indeed causes a longer latency at the start of playback; but only because of a different setting which MPD links with the hardware buffer size: the "start threshold". But the same does not apply to this plugin.
After this clarification, do we really need this setting?

Wang-Yue:
I don't know how to dispose both the boost ring buffer in the plugin, and CoreAudio's hardware buffer. 

-  boost::lockfree::spsc_queue does not have a flush API so it may be hard to reset the buffer in Cancel(). 

- CoreAudio provides no public documentation to flush the internal buffer other than dispose the existing AU and reinitiate a new one.  

If we just reset the boost buffer, it is possible the renderer only reads half of a single sample frame while a buffer gets flushed, and may cause it to render wrong data when new data is pushed into the buffer.

coroner21:
You can call AudioOutputUnitStop / Start before and after flushing ring buffer to prevent calls to the callback in between.

Wang-Yue:
boost::lockfree::spsc_queue is not flushable.  @MaxKellermann , Do you think it's worthwhile importing a new lock-free ring buffer library (such as the one provided by PortAudio) to resolve this issue?

Wang-Yue:
#312 is created.

PR #311, opened 2018-07-15
Support timed_wait in microseconds level

Wang-Yue:
Some Audio plugin (such as ALSA, and soon CoreAudio) already support microsecond level buffer time. However, current interval less than 1000 microseconds will cause a bug that the code treated as 0 ms.

### DIFF
--- a/src/thread/PosixCond.hxx
+++ b/src/thread/PosixCond.hxx
@@ -75,13 +75,13 @@ public:
 	}
 
 private:
-	bool timed_wait(PosixMutex &mutex, unsigned timeout_ms) noexcept {
+	bool timed_wait(PosixMutex &mutex, long timeout_us) noexcept {
 		struct timeval now;
 		gettimeofday(&now, nullptr);
 
 		struct timespec ts;
-		ts.tv_sec = now.tv_sec + timeout_ms / 1000;
-		ts.tv_nsec = (now.tv_usec + (timeout_ms % 1000) * 1000) * 1000;
+		ts.tv_sec = now.tv_sec + timeout_us / 1000000;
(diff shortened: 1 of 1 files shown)


MaxKellermann:
This is bad for two reasons:
1. it converts the parameter to `signed`
2. it converts the parameter to `long` which may be 32 bit or 64 bit

Strictly speaking, `unsigned` may be 16 bit, so it wasn't a good choice already - but one which worked on all targets MPD ever ran on (but I'd be happy to change it). But your switch to `long` suggests that you think it should have a bigger range (which sounds reasonable), but that is not what the code really does on many architectures where MPD usually runs. Even more dangerously: by switching it to `signed`, you reduce the range.

Wang-Yue:
what about something like uint32_t?

MaxKellermann:
I'd really like to use `std::condition_variable` and drop all of our custom implementations, but unfortunately GCC doesn't implement it on Windows.
Yes, you can use `uint32_t`; for parameters, it is usually better to use `uint_least32_t`. But since you chose `long`, I thought you wanted at least 64 bit, why else would you switch to `long`?

Wang-Yue:
Changed to uint_least32_t. 

I chose long since tv_usec is long. But 32bit should already enough for the wait --- 2^32 microsec is 4000second, which should be a wait long enough.

MaxKellermann:
`tv_usec`'s maximum legal value is 999999. This cumbersome `struct timeval` was invented because there was no integer type which could map all valid time values with microsecond precision. Today, 64 bit is enough for everybody ;-)
Much of the C standard library was designed at a time when nobody really cared about correct types. By today's standards, it's a lot of "WTF". I mean, how can a wait time possibly be negative, even less so its fractional value?

PR #312, opened 2018-07-16
Implement Cancel() for OSXOutputPlugin

Wang-Yue:


MaxKellermann:
The `reset()` method which was added in Boost 1.58, but MPD is currently declared to be compatible with Boost 1.54. I'd agree to raise the minimum version to 1.58 to take advantage of this method (which is also useful for other plugins).
Your `Cancel()` implementation misses a call to `PcmExport::Reset()`. That method is currently empty, but eventually code will be added there, so making this correct in advance is a good idea.

Wang-Yue:
Done

PR #317, opened 2018-07-20
Convert user documentation to restructuredText

carnager:


carnager:
There are some tables where paragraphs had to be squashed to a single one. Not sure if restructuredText  allows paragraphs in tables at all.

PR #324, opened 2018-07-26
configure.ac: clarify sidplay related info

skidoo23:


MaxKellermann:
I've cherry-picked the one commit, to skip the merge commit you made. Please don't do merge commits. Until a PR is merged, it's a good idea to keep your branch rebased to the target branch, i.e. rebase is ok, but merge is not.

PR #327, opened 2018-07-30
decoder/sidplay: support basic and kernal rom (libsidplayfp)

skidoo23:
Some SID tunes - eg Commodore Basic Demos from HVSC - require basic and/or kernal rom to play but dumps are not embedded in libsidplayfp - please note https://sourceforge.net/p/sidplay-residfp/news/2013/01/released-libsidplayfp-100beta1/ 

Patch adds kernal and basic configuration parameters to load corresponding rom dumps.

### DIFF
--- a/src/decoder/plugins/SidplayDecoderPlugin.cxx
+++ b/src/decoder/plugins/SidplayDecoderPlugin.cxx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2003-2017 The Music Player Daemon Project
+ * Copyright 2003-2018 The Music Player Daemon Project
  * http://www.musicpd.org
  *
  * This program is free software; you can redistribute it and/or modify
@@ -25,6 +25,10 @@
 #include "DetachedSong.hxx"
 #include "fs/Path.hxx"
 #include "fs/AllocatedPath.hxx"
+#ifdef HAVE_SIDPLAYFP
+#include "fs/io/FileReader.cxx"
+#include "util/RuntimeError.hxx"
+#endif
 #include "util/Macros.hxx"
 #include "util/StringFormat.hxx"
 #include "util/Domain.hxx"
@@ -65,6 +69,21 @@ static unsigned default_songlength;
 
 static bool filter_setting;
 
+#ifdef HAVE_SIDPLAYFP
+static constexpr unsigned rom_size = 8192;
+static uint8_t *kernal, *basic = nullptr;
+
+static void loadRom(const Path rom_path, uint8_t *dump)
+{
+	FileReader romDump(rom_path);
+	if (romDump.Read(dump, rom_size) != rom_size)
+	{
+		throw FormatRuntimeError
+			("Could not load rom dump '%s'", rom_path.c_str());
+	}
+}
+#endif
+
 static SidDatabase *
 sidplay_load_songlength_db(const Path path)
 {
@@ -100,6 +119,24 @@ sidplay_init(const ConfigBlock &block)
 
 	filter_setting = block.GetBlockValue("filter", true);
 
+#ifdef HAVE_SIDPLAYFP
+	/* read kernal rom dump file */
+	const auto kernal_path = block.GetPath("kernal");
+	if (!kernal_path.IsNull())
+	{
+		kernal = new uint8_t[rom_size];
+		loadRom(kernal_path, kernal);
+	}
+
+	/* read basic rom dump file */
+	const auto basic_path = block.GetPath("basic");
+	if (!basic_path.IsNull())
+	{
+		basic = new uint8_t[rom_size];
+		loadRom(basic_path, basic);
+	}
+#endif
+
 	return true;
 }
 
@@ -107,6 +144,11 @@ static void
 sidplay_finish() noexcept
 {
 	delete songlength_database;
+
+#ifdef HAVE_SIDPLAYFP
+	delete[] basic;
+	delete[] kernal;
+#endif
 }
 
 struct SidplayContainerPath {
@@ -202,6 +244,8 @@ sidplay_file_decode(DecoderClient &client, Path path_fs)
 
 #ifdef HAVE_SIDPLAYFP
 	sidplayfp player;
+
+	player.setRoms(kernal, basic, nullptr);
 #else
 	sidplay2 player;
 #endif
--- a/doc/user.rst
+++ b/doc/user.rst
@@ -1328,6 +1328,10 @@ C64 SID decoder based on `libsidplayfp <https://sourceforge.net/projects/sidplay
      - This is the default playing time in seconds for songs not in the songlength database, or in case you're not using a database. A value of 0 means play indefinitely.
    * - **filter yes|no**
      - Turns the SID filter emulation on or off.
+   * - **kernal**
+     - Only libsidplayfp. Roms are not embedded in libsidplayfp - please note https://sourceforge.net/p/sidplay-residfp/news/2013/01/released-libsidplayfp-100beta1/ But some SID tunes require rom images to play. Make C64 rom dumps from your own vintage gear or use rom files from Frodo or VICE emulation software tarballs. Absolute path to kernal rom image file.
+   * - **basic**
+     - Only libsidplayfp. Absolute path to basic rom image file.
 
 sndfile
 ~~~~~~~


src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -65,6 +66,38 @@ static unsigned default_songlength;
 
 static bool filter_setting;
 
+#ifdef HAVE_SIDPLAYFP
+static const unsigned romSize = 8192;
```
MaxKellermann: Use `constexpr`

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -40,6 +40,7 @@
 #include <sidplayfp/builders/resid.h>
 #include <sidplayfp/builders/residfp.h>
 #include <sidplayfp/SidDatabase.h>
+#include <fs/io/FileReader.cxx>
```
MaxKellermann: This is an internal header; should have double quotes instead of pointed quotes.

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -65,6 +66,38 @@ static unsigned default_songlength;
 
 static bool filter_setting;
 
+#ifdef HAVE_SIDPLAYFP
+static const unsigned romSize = 8192;
+static std::unique_ptr<uint8_t[]> kernal(new uint8_t[romSize]());
+static std::unique_ptr<uint8_t[]> basic(new uint8_t[romSize]());
+
+static void loadRom(std::unique_ptr<uint8_t[]> &dump, const Path romPath)
```
MaxKellermann: There's no point in passing the `std::unique_ptr` here, because that pointer will not be modified by the function; just pass `uint8_t *`; and I'd prefer to have the `Path` parameter first. Also, don't use camelCase for variable names; underscore instead.

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -65,6 +66,38 @@ static unsigned default_songlength;
...
+		{
+			romDump.Read(dump.get(), romSize);
+			FormatInfo(sidplay_domain, "LOAD \"%s\",8,1",
+				romPath.c_str());
+		}
+	}
+	catch (...)
+	{
+		/* We break for nobody. Most tunes play without roms */
+		FormatWarning(sidplay_domain,
```
MaxKellermann: If you catch all exceptions here, how does the caller know whether the buffers contain valid data? Maybe the best idea would be to return a `std::unique_ptr<uint8_t[]>`, and return `nullptr` on failure?
Or let the caller catch the exception? I think startup should fail if a ROM path was configured but failed to load. Even if some files play without a ROM - it's a misconfiguration.

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -65,6 +66,38 @@ static unsigned default_songlength;
...
+		FileReader romDump(romPath);
+		if (romDump.GetSize() != romSize)
+		{
+			FormatWarning(sidplay_domain,
+				"Corrupted rom dump %s - %u bytes expected",
+				romPath.c_str(), romSize);
+		}
+		else
+		{
+			romDump.Read(dump.get(), romSize);
```
MaxKellermann: Forgot to check the `Read()` return value.

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -65,6 +66,38 @@ static unsigned default_songlength;
...
+			FormatInfo(sidplay_domain, "LOAD \"%s\",8,1",
```
MaxKellermann: What's the point of this log message?

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -202,6 +250,9 @@ sidplay_file_decode(DecoderClient &client, Path path_fs)
 
 #ifdef HAVE_SIDPLAYFP
 	sidplayfp player;
+
+	player.setRoms((const uint8_t*)kernal.get(),
```
MaxKellermann: Is this cast really necessary?

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -202,6 +250,9 @@ sidplay_file_decode(DecoderClient &client, Path path_fs)
...
+		(const uint8_t*)basic.get(), nullptr);
```
MaxKellermann: How do you know if these buffers have valid data? (See above)

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -65,6 +66,38 @@ static unsigned default_songlength;
 
 static bool filter_setting;
 
+#ifdef HAVE_SIDPLAYFP
+static const unsigned romSize = 8192;
+static std::unique_ptr<uint8_t[]> kernal(new uint8_t[romSize]());
+static std::unique_ptr<uint8_t[]> basic(new uint8_t[romSize]());
```
MaxKellermann: I don't like that these buffers get allocated even if no ROM path is configured. Also why call the constructor on the allocated array?

src/decoder/plugins/SidplayDecoderPlugin.cxx
skidoo23: Agreed. MPD should fail if option is misconfigured.

src/decoder/plugins/SidplayDecoderPlugin.cxx
MaxKellermann: Convention. No technical reason.

src/decoder/plugins/SidplayDecoderPlugin.cxx
skidoo23: Oops. 

Is the size check sufficient or should additional MD5/SHA2 checksums be implemented?

src/decoder/plugins/SidplayDecoderPlugin.cxx
skidoo23: Come on, its only 16 kb ;) - Please allow some time for making requested changes.

MaxKellermann:
Please fold those commits into one. I can't review these fixup commits.

skidoo23:
done.

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -65,6 +69,21 @@ static unsigned default_songlength;
 
 static bool filter_setting;
 
+#ifdef HAVE_SIDPLAYFP
+constexpr unsigned rom_size = 8192;
```
MaxKellermann: Make this `static` or else the compiler will generate a global symbol.

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -100,13 +119,36 @@ sidplay_init(const ConfigBlock &block)
 
 	filter_setting = block.GetBlockValue("filter", true);
 
+#ifdef HAVE_SIDPLAYFP
+	/* read kernal rom dump file */
+	const auto kernal_path = block.GetPath("kernal");
+	if (!kernal_path.IsNull())
+	{
+		kernal = new uint8_t[rom_size - 1];
```
MaxKellermann: Why `-1`?

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -100,13 +119,36 @@ sidplay_init(const ConfigBlock &block)
...
+		loadRom(kernal_path, kernal);
+	}
+
+	/* read basic rom dump file */
+	const auto basic_path = block.GetPath("basic");
+	if (!basic_path.IsNull())
+	{
+		basic = new uint8_t[rom_size - 1];
```
MaxKellermann: `-1`?

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -65,6 +69,21 @@ static unsigned default_songlength;
 
 static bool filter_setting;
 
+#ifdef HAVE_SIDPLAYFP
+constexpr unsigned rom_size = 8192;
+uint8_t *kernal, *basic = nullptr;
```
MaxKellermann: Should also be `static` to avoid namespace collision.

skidoo23:
Thanks for merging and your advice.

PR #329, opened 2018-08-01
Add support for unix socket to HttpdOutputPlugin

1848:
Just a quick patch to enable unix sockets.
I did no checking for HAVE_UN since this is already done in AddPath.
"Works fine for me".
For now socket permissions are hardcoded to 0666 but I will create a PR tomorrow (or so) for adding support to set socket permissions in mpd.conf...

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1023,6 +1023,7 @@ libconf_a_SOURCES = \
 	src/config/Migrate.cxx src/config/Migrate.hxx \
 	src/config/Templates.cxx src/config/Templates.hxx \
 	src/config/Domain.cxx src/config/Domain.hxx \
+	src/config/Net.cxx src/config/Net.hxx \
 	src/config/Option.hxx
 
 # the Song library
--- a/src/config/Net.cxx
+++ b/src/config/Net.cxx
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2003-2017 The Music Player Daemon Project
+ * http://www.musicpd.org
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include "config.h"
+#include "Net.hxx"
+#include "event/ServerSocket.hxx"
+#include "Path.hxx"
+#include "fs/AllocatedPath.hxx"
+
+void
+ServerSocketAddGeneric(ServerSocket &server_socket, const char *address, unsigned int port)
+{
+	if (address == nullptr || 0 == strcmp(address, "any")) {
+		server_socket.AddPort(port);
+	} else if (address[0] == '/' || address[0] == '~') {
+		server_socket.AddPath(ParsePath(address));
+	} else {
+		server_socket.AddHost(address, port);
+	}
+}
--- a/src/config/Net.hxx
+++ b/src/config/Net.hxx
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2003-2017 The Music Player Daemon Project
+ * http://www.musicpd.org
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef MPD_CONFIG_NET_HXX
+#define MPD_CONFIG_NET_HXX
+
+class ServerSocket;
+
+/**
+ * Sets the address or unix socket of a ServerSocket instance
+ * There are three possible ways
+ * 1) Set address to a valid ip address and specify port.
+ *    server_socket will listen on this address/port tuple.
+ * 2) Set address to null and specify port.
+ *    server_socket will listen on ANY address on that port.
+ * 3) Set address to a path of a unix socket. port is ignored.
+ *    server_socket will listen on this unix socket.
+ *
+ * Throws #std::runtime_error on error.
+ *
+ * @param server_socket the instance to modify
+ * @param address the address to listen on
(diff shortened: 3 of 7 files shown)


MaxKellermann:
If your PR needs text, then your commit message is bad. And indeed it is.

src/output/plugins/httpd/HttpdOutputPlugin.cxx
```diff
@@ -63,12 +64,14 @@ HttpdOutput::HttpdOutput(EventLoop &_loop, const ConfigBlock &block)
 	clients_max = block.GetBlockValue("max_clients", 0u);
 
 	/* set up bind_to_address */
-
```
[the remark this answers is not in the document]
1848: accidental

src/output/plugins/httpd/HttpdOutputPlugin.cxx
```diff
@@ -63,12 +64,14 @@ HttpdOutput::HttpdOutput(EventLoop &_loop, const ConfigBlock &block)
...
 	const char *bind_to_address = block.GetBlockValue("bind_to_address");
-	if (bind_to_address != nullptr && strcmp(bind_to_address, "any") != 0)
-		AddHost(bind_to_address, port);
-	else
+	if(bind_to_address == nullptr || !strcmp(bind_to_address, "any"))
 		AddPort(port);
+	/* check for (abstract) unix socket */
+	else if(bind_to_address[0] == '/' || (bind_to_address[0] == '@' && bind_to_address[1] == '/'))
```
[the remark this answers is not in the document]
1848: Abstract socket (https://linux.die.net/man/7/unix)
Same as in AllocatedSocketAddress::SetLocal[76]

1848:
Alright, merged the code for parsing the listen address.
ServerSocket::AddGeneric(const char *address, unsigned int port) is used for that now.
The name of the function is somewhat placeholder since I had no better idea.
Is that what you were talking about?

src/event/ServerSocket.cxx
```diff
@@ -35,6 +35,7 @@
 #include "util/Domain.hxx"
 #include "util/ScopeExit.hxx"
 #include "Log.hxx"
+#include "config/Path.hxx"
```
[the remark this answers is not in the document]
1848: I neither like pulling in big deps but I had no better idea.
We need someting like ParsePath() there for home dirs (only allowing absolute paths would not be nice)...
And I am not sure if its a good idea to move AddGeneric away from ServerSocket.
What do you suggest?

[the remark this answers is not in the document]
MaxKellermann: I had a similar idea like your commit a while ago, but didn't have a quick idea how to solve this dependency problem. Resolving home directories must not be done here because that may require accessing configuration data from a global variable; during startup, MPD may still run as "root", but "~/" must resolve to the configured user's home directory, not to root's.
Maybe the config library needs a function which expands a string which is possibly a path, but leaves it as UTF-8 string (instead of using the FS charset / `class Path`). But then it may need to convert the configured UTF-8 string to FS charset, expand the path, convert back to UTF-8, and then the ServerSocket converts it to FS charset again. Ugly!
Another idea: implement a function in the config library which uses a configured (UTF-8) string to set up a given `ServerSocket` instance. No change to `ServerSocket` needed, and the heavy-dependency code lives in the config library.

1848:
Alright. Created ServerSocketAddGeneric in the config library. This is just a placeholder name for the moment.
Put it in a new file called Util.{cxx,hxx} for now.
What do you think?

Makefile.am
```diff
@@ -1018,7 +1018,8 @@ libconf_a_SOURCES = \
 	src/config/Migrate.cxx src/config/Migrate.hxx \
 	src/config/Templates.cxx src/config/Templates.hxx \
 	src/config/Domain.cxx src/config/Domain.hxx \
-	src/config/Option.hxx
+	src/config/Option.hxx \
+	src/config/Util.cxx src/config/Util.hxx
```
MaxKellermann: If you move this before the `Option.hxx`, then you save two lines of diff.

src/config/Util.hxx
```diff
@@ -0,0 +1,28 @@
...
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef MPD_CONFIG_UTIL_HXX
+#define MPD_CONFIG_UTIL_HXX
+
+class ServerSocket;
+
+void
+ServerSocketAddGeneric(const char *address, unsigned int port, ServerSocket &server_socket);
```
MaxKellermann: The `ServerSocket` should be the first parameter here - by convention again, and because this is some kind of `this` pointer which is also the first parameter passed to a method.

src/config/Util.cxx
```diff
@@ -0,0 +1,17 @@
+#include "config.h"
+#include "Util.hxx"
+#include "event/ServerSocket.hxx"
+#include "config/Path.hxx"
```
MaxKellermann: Don't need the `config/` prefix here.

src/config/Util.hxx
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef MPD_CONFIG_UTIL_HXX
+#define MPD_CONFIG_UTIL_HXX
+
+class ServerSocket;
+
```
MaxKellermann: API documentation?

MaxKellermann:
I'd call it `src/config/Net.cxx` because this is networking-specific code in the config library.
Can you make this a single commit based on current master? I don't want to merge the intermediate commit.

src/config/Net.hxx
```diff
@@ -0,0 +1,45 @@
...
+ *    server_socket will listen on ANY address on that port.
+ * 3) Set address to a path of a unix socket. port is ignored.
+ *    server_socket will listen on this unix socket.
+ *
+ * Throws #std::runtime_error on error.
+ *
+ * @param server_socket the instance to modify
+ * @param address the address to listen on
+ * @param port the port to listen on
+ * @param error location to store the error occurring
```
MaxKellermann: Copy'n'paste error? Don't copy wrong/obsolete API documentation!

src/config/Net.cxx
```diff
@@ -0,0 +1,17 @@
+#include "config.h"
```
MaxKellermann: Copyright header missing

src/output/plugins/httpd/HttpdOutputPlugin.cxx
```diff
@@ -58,17 +59,11 @@ HttpdOutput::HttpdOutput(EventLoop &_loop, const ConfigBlock &block)
...
 	clients_max = block.GetBlockValue("max_clients", 0u);
 
 	/* set up bind_to_address */
 
-	const char *bind_to_address = block.GetBlockValue("bind_to_address");
-	if (bind_to_address != nullptr && strcmp(bind_to_address, "any") != 0)
-		AddHost(bind_to_address, port);
-	else
-		AddPort(port);
+	ServerSocketAddGeneric(*this, block.GetBlockValue("bind_to_address"), block.GetBlockValue("port", 8000u));
```
MaxKellermann: This is a feature change. I'd prefer to have this in a separate commit, along with a NEWS line and user documentation.

1848:
Are you good with this "user documentation"-like thing or do you think there should be more (e.g. syntax for specifying a socket)?

MaxKellermann:
Your addition to the user documentation is enough for now.
Documenting the syntax would be good, but this part could refer to the global `bind_to_address` documentation (there's no point in repeating this here). But abstract sockets aren't yet documented there... improvements to existing documentation are always welcome.

1848:
Acknowledged.
Will (hopefully) be included in next PR since I need to update it anyway.
Thanks for merging.

PR #331, opened 2018-08-04
Use fchmod instead of chmod for unix sockets to prevent TOCTTOU

1848:
Quick fix to use fchmod for unix sockets to prevent TOCTTOU (TimeOfCheckToTimeOfUse).
Its better to set permissions before the socket is finally created.

Also it would be good to let the user decide which permissions to apply instead of hardcoding 0666.
Have this working on a local branch, maybe worth another PR?

Btw:
How about calling unlink() for a (maybe existing) socket in socket_bind_listen() or SocketDescriptor::Bind(), instead of ServerSocket::AddPath()?
So we would have creation and deletion together. No disadvantage in the current approach, its just a feeling.

### DIFF
--- a/src/net/SocketUtil.cxx
+++ b/src/net/SocketUtil.cxx
@@ -23,6 +23,8 @@
 #include "SocketError.hxx"
 #include "UniqueSocketDescriptor.hxx"
 
+#include <sys/stat.h>
+
 UniqueSocketDescriptor
 socket_bind_listen(int domain, int type, int protocol,
 		   SocketAddress address,
@@ -31,6 +33,14 @@ socket_bind_listen(int domain, int type, int protocol,
 	UniqueSocketDescriptor fd;
 	if (!fd.CreateNonBlock(domain, type, protocol))
 		throw MakeSocketError("Failed to create socket");
+
+
+#ifdef HAVE_UN
+	if (domain == AF_UNIX) {
+		/* allow everybody to connect */
+		fchmod(fd.Get(), 0666);
+	}
+#endif
 
 	if (!fd.SetReuseAddress())
 		throw MakeSocketError("setsockopt() failed");
--- a/src/event/ServerSocket.cxx
+++ b/src/event/ServerSocket.cxx
@@ -39,7 +39,6 @@
 #include <string>
 #include <algorithm>
 
-#include <sys/stat.h>
 #include <string.h>
 #include <unistd.h>
 #include <assert.h>
@@ -183,13 +182,6 @@ OneServerSocket::Open()
 	auto _fd = socket_bind_listen(address.GetFamily(),
 				      SOCK_STREAM, 0,
 				      address, 5);
-
-#ifdef HAVE_UN
-	/* allow everybody to connect */
-
-	if (!path.IsNull())
-		fchmod(_fd.Get(), 0666);
-#endif
 
 	/* register in the EventLoop */	
 
--- a/src/event/ServerSocket.cxx
+++ b/src/event/ServerSocket.cxx
@@ -188,10 +188,10 @@ OneServerSocket::Open()
 	/* allow everybody to connect */
 
 	if (!path.IsNull())
-		chmod(path.c_str(), 0666);
+		fchmod(_fd.Get(), 0666);
 #endif
 
-	/* register in the EventLoop */
+	/* register in the EventLoop */	
 
 	SetFD(_fd.Release());
 }


src/net/SocketUtil.cxx
```diff
@@ -32,8 +34,21 @@ socket_bind_listen(int domain, int type, int protocol,
...
 		throw MakeSocketError("Failed to create socket");
 
+#ifdef HAVE_UN
+	if (domain == AF_UNIX) {
+		/* allow everybody to connect */
+		if(fchmod(fd.Get(), 0666) != 0)
+			throw MakeSocketError("fchmod() failed");
+	}
+	else {
+		/* SO_REUSEADDR has no effect on unix sockets */
```
MaxKellermann: This is an unrelated change. And it duplicates code.

1848: Why is it unrelated? Do you prefer to keep the code in OneServerSocket::Open()?
Which code is duplicated, what do you mean?

MaxKellermann: Wrong question: you need to explain how this new runtime check for `SetReuseAddress()` is related to switching to `fchmod()`.
You copy'n'pasted the `SetReuseAddress()` call. There are two of them now, one in each `#if` branch. Code duplication is bad.

1848: Sure, its possible to remove that duplication but then SetReuseAddress() would run even if its a unix socket. Thats not a problem at all but its also not necessary.
So so question is duplicate code or have code to execute "nops".
I thought the first option would be better but I am fine if you prefer the second.

1848: Are you fine with moving that code to SocketUtil or do you want it to stay in ServerSocket?

MaxKellermann: Since this patch is about changing `chmod()` to `fchmod()`, it should do exactly that, and not move code around. Don't do two things at the same time - you make patch review harder than necessary. If you want to move code around or do other things (like adding a runtime check to `SetReuseAddress()`), make that a separate commit.

1848:
Okay, splitted into 2 commits. The second commit about moving the function away has not much value yet (I guess) but it will maybe when the following points are decided:

1. There is no error checking on chmod/fchmod yet. Is it bad enough to throw an error if it fails? I guess for now it is not because worst case is that the user cannot connect. But when there are non-hardcoded permissions someday then the user might care about permissions (e.g. security related). So shall we keep it this way or do we want to throw when the call fails? I like the latter because its always better to handle "errors" but its your call of course.

2. If I understood right you want to keep calling SetReuseAddress() on unix socket which has no effect to prevent duplicated code?

3. What about unlink() in ServerSocket::AddPath(). Shouldnt it be called from socket_bind_listen() too? So we would have creation and deletion together.

4. Another "additional check vs calling noop function"-like question. unlink() has no effect on abstract sockets (since they get destroyed when the process exits). Ignore and keep calling or check path if its an abstract socket?

5. What do you think in general about letting the user decide which permissions to apply?
Not for this PR of course, just want to know for the near future.

MaxKellermann:
> There is no error checking on chmod/fchmod yet. Is it bad enough to throw an error if it fails?

Even though I could not imagine this could ever fail, I didn't think this was a condition worth implementing error handling. The worst case is harmless. But you are right, once there are configurable permissions, they should be applied reliably, and failure should be noticable.

> If I understood right you want to keep calling SetReuseAddress() on unix socket which has no effect to prevent duplicated code?

No. That was a good change, but it was an undocumented part of the `fchmod` commit. That bothered me, because it obscures my view on your change.

> What about unlink() in ServerSocket::AddPath(). Shouldnt it be called from socket_bind_listen() too?

I don't think so. That's a generic and very low-level library which shouldn't know or care about specialties of certain address/protocol families.

> unlink() has no effect on abstract sockets

Unlink on abstract sockets is bad. Abstract sockets start with a null byte, therefore the unlink will attempt to delete a file with an empty name; this will fail with no side effects. But if you call unlink before replacing the leading `@` with the null byte, this may cause data loss... if you happen to have a file in the current directory with that name.

> What do you think in general about letting the user decide which permissions to apply?

Why not.

1848:
First, thanks for merging.
1) Will change that when I work on the permissions stuff.

2) I am sorry, I guess I misunderstood. The current merge keeps calling SetReuseAddress() even if its a unix socket. Do you want me to change this again? Otherwise I would do this in a seperate commit when I am on the permissions stuff if thats okay for you.

3) Good point.

Issue #337, opened 2018-08-10
Inconsistent behavior with fchmod()


1848:
The recently switch to fchmod() for socket permissions seems to be not working well on some machines.
While everything worked perfectly fine when I tested, now it does not seem to work (I am now on a different machine but same arch, same dist, minor kernel version ahead).
Can anyone please confirm that a unix sockets does not get the right permission 0666 on creation?

While reading the docs and the internet I can also see inconsistent behavior.
Some ppl are confirming its works, some say its a "noop". For me its not a noop, but the permissions are not correct (e.g. getting 0755 instead of 0666).
unix7 manpage does not mention fchmod for setting permissions although its says fchmod is not working for abstract socket. Not sure why its mentioned there when its also not working for non-abstract sockets.
Anyway, will do more research but for now I think there is a need to revert this change.
@MaxKellermann can you do this? I will provide a PR to revert the changes asap just in case you prefer to not rollback that commit.
chmod() is not an optimal solution but it seems like its still the best solution for now.

Relevant commit:
84054203afeb9f3c0ece9b85922bb357f9419d14

Thanks.

Edit:
Related PR:  #339

MaxKellermann:
Just from a rough look at the kernel sources:
`fchmod()` can be used to reduce the permission bits on a local socket, but the umask is applied when `bind()` is called:
https://github.com/torvalds/linux/blob/master/net/unix/af_unix.c#L1008..L1010
So `fchmod()` can be used as a security measure to fix race conditions, but to be sure, it must be called again after `bind()` to undo the effects of applying the umask.

1848:
Do you have any success with calling fchmod() AFTER bind()?
On my machine fchmod() after bind() is a noop.

For me it seems like
- fchmod(0) // prevent race conds
- bind()
- chmod(0666)

is the (best?) way solving this without messing with umask().

MaxKellermann:
I think with local sockets, the socket and the file system inode are distinct objects, therefore `fchmod()` changes the socket's permissions, but not the file system inode. The inode inherits the socket's permissions upon creation (i.e. `bind()`), but the umask is applied.
So yes, if fixing the race condition is important, your approach seems good. But right now, without stricter permissions, there is no race condition; MPD wants to broaden the socket permissions, but never narrows it.

1848:
Of course you are completely right that there is no real need to secure this right now but there is also no disadvantage.
So I would create a PR tomorrow to implement this approach if you are good with this?

[...]

PR #339, opened 2018-08-13
fixed setting unix socket permissions

1848:
Call fchmod() first to prevent TOCTTOU.
Then apply permissions using hmod().

This is related to issue #337

### DIFF
--- a/src/net/SocketUtil.hxx
+++ b/src/net/SocketUtil.hxx
@@ -32,6 +32,10 @@ class SocketAddress;
 /**
  * Creates a socket listening on the specified address.  This is a
  * shortcut for socket(), bind() and listen().
+ * When a unix socket is created (domain == AF_UNIX), its
+ * permissions will be stripped down to prevent unauthorized
+ * access. The caller is responsible to apply proper permissions
+ * at a later point.
  *
  * Throws #std::system_error on error.
  *
--- a/src/event/ServerSocket.cxx
+++ b/src/event/ServerSocket.cxx
@@ -184,6 +184,13 @@ OneServerSocket::Open()
 				      SOCK_STREAM, 0,
 				      address, 5);
 
+#ifdef HAVE_UN
+	/* allow everybody to connect */
+
+	if (!path.IsNull())
+		chmod(path.c_str(), 0666);
+#endif
+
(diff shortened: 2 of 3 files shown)


MaxKellermann:
I'm not so happy that the generic `socket_bind_listen()` now creates sockets which are completely inaccessible. This requires every caller to adjust the permissions. In the MPD code base, there is just the one caller, but this is still a heavy restriction on this low-level function, and should at least be documented.
And it is only there to avoid a problem with a feature that doesn't even exist (yet).

1848:
Added a bit of documentation to socket_bind_listen.

I understand your concerns but in my opinion that is not a big problem since everyone using unix sockets should care about security. Actually this is not specific to unix socket but to security related things in general. I think its better to prevent access when someone forgets about permissions than giving full access. Also it is way more easier to detect not given access than givien full access.
As always its your call. I am fine if you prefer the other way but then we need to find a good way to solve this.
One possibility would be to "overload" the function but I am not a big fan of this (for this type of problem).
Other possibility could be a flag which can be set (e.g. in SetPath()) which control whether permissions get striped or not.
What do you prefer?

src/net/SocketUtil.hxx
```diff
@@ -32,6 +32,9 @@ class SocketAddress;
 /**
  * Creates a socket listening on the specified address.  This is a
  * shortcut for socket(), bind() and listen().
+ * When a unix socket is created (domain == AF_UNIX), its
+ * permissions will be striped down to prevent unauthorized
```
MaxKellermann: Typo "strip**p**ed".

src/net/SocketUtil.hxx
```diff
@@ -32,6 +32,9 @@ class SocketAddress;
...
+ * access until proper permissions are applied at a later point.
```
MaxKellermann: Should clarify that the caller is responsible for this "applying at a later point". Important parts of the contract should be clear enough so nobody can possibly misunderstand.

1848: Just removed that "manually" word before pushing because I thought it would be sufficient this ways :)
But valid point, give me a minute.

PR #341, opened 2018-08-17
Enable Curl plugin on win32

jwise:
The Win32 instance of the Curl plugin had bit-rotted a bit when Expat became a requirement; this adds Expat to the Windows build, and fixes up a CFLAGS requirement when statically linking Curl.

### DIFF
--- a/python/build/libs.py
+++ b/python/build/libs.py
@@ -364,6 +364,15 @@ curl = AutotoolsProject(
     patches='src/lib/curl/patches',
 )
 
+libexpat = AutotoolsProject(
+    'https://github.com/libexpat/libexpat/releases/download/R_2_2_6/expat-2.2.6.tar.bz2',
+    '17b43c2716d521369f82fc2dc70f359860e90fa440bea65b3b85f0b246ea81f2',
+    'lib/libexpat.a',
+    [
+        '--disable-shared', '--enable-static',
+    ],
+)
+
 libnfs = AutotoolsProject(
     'https://github.com/sahlberg/libnfs/archive/libnfs-3.0.0.tar.gz',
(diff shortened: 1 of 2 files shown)


MaxKellermann:
I don't understand the `CURL_STATICLIB` thing. What does it do, and what really causes the linker problem? Your commit message doesn't explain it, and your code contains no information at al. If it turns out to be necessary, there should be a prominent code comment explaining why it's there.

Btw. Expat has always been a requirement for the Curl **storage** plugin ever since the plugin was added to MPD. It has never been a requirement for for the Curl **input** plugin.

MaxKellermann:
Real fix for the `CURL_STATICLIB` linker problem: https://github.com/MusicPlayerDaemon/MPD/commit/533a3def9f300a2c4cd6f5baa303e6e6f2ae0e14
This was a `Makefile.am` bug, and while your patch appeared to make the symptom go away, it did not address the actual bug. I'm always cautious with under-documented workaround patches.

I cherry-picked the libexpat commit to v0.20.x: https://github.com/MusicPlayerDaemon/MPD/commit/a5cc13b0c5327ab742e2d3a66de63e1d0e02e7a6

jwise:
Makes sense.  I knew something was hokey about that `CURL_STATICLIB` thing, but I wasn't able to figure out how to convince the build system to do it better -- and, especially, I don't think I ever would have found the `CURL_CXXFLAGS`!  And, yes -- I've only been trying the Curl storage plugin, which I can imagine gets relatively little use.

Probably there is an analogous thing that has to happen for the Android build, too, but I haven't been willing to subject myself to the agony of messing with this stuff Android quite yet :-)

Thanks for the feedback.

PR #348, opened 2018-08-28
First draft of the protocol as rst

rain0r:
Started converting the protocol from Docbook to rst. Internal links are not working (yet).

Leave some notes / comments / criticism.

Tools I used:

https://github.com/wojdyr/db2rst
https://github.com/rst2pdf/rst2pdf


doc/protocol.rst
```diff
@@ -0,0 +1,1100 @@
...
+
+where ``version`` is a version identifier such as
+0.12.2.  This version identifier is the version of the protocol
+spoken, not the real version of the daemon.  (There is no way to
+retrieve this real version identifier from the connection.)
+
+Requests
+========
+
+CMDSYN:  COMMAND ARG
```
MaxKellermann: I suppose `CMDSYN` is translated from DocBook's `<cmdsynopsis>`, but this loses all meta information. In DocBook, this was:

      <cmdsynopsis>
        <command>COMMAND</command>
        <arg rep="repeat"><replaceable>ARG</replaceable></arg>
      </cmdsynopsis>

The "repeat" was lost, and this is important. Of course we need to decide how this should look like in RST, but in any case this is bad.

doc/protocol.rst
```diff
@@ -0,0 +1,1100 @@
...
+  some (hopefully) informative text that describes the
+  nature of the error.
+
+foo
+~~~
+
+An example might help. Consider the following sequence
+sent from the client to the server.
+
+CMDSYN:  command_list_begin volume 86 play 10240 status command_list_end
```
MaxKellermann: Line breaks were lost here.

MaxKellermann:
Is this going anywhere?

doc/protocol.rst
rain0r: I'm not sure how to reproduce this with RSt. Any suggestions?

PR #384, opened 2018-10-30
output/Alsa: workaround 'next is not a member' (Boost 1.67)

skidoo23:
MPD from git/master does not build on 18.10 *buntus. Workaround for build error ‘next’ is not a member of ‘boost’ build error related to Boost 1.67.


MaxKellermann:
This should better be fixed by Ubuntu, but I don't see a bug report about it in the Ubuntu bug tracker.
Btw. Boost 1.68 was released two months before Ubuntu 18.10. Blame Ubuntu for shipping a known-buggy Boost version.

PR #388, opened 2018-10-31
include hostname in zeroconf_name (fixes #387)

### LINKED ISSUE #387: zeroconf name should include hostname
This is forwarded from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=516974, which used to live at http://bugs.musicpd.org/view.php?id=4293

Joey Hess writes:

```
# The service name to publish. This name should be unique on
# your local network.
#
#zeroconf_name "Music Player"
```

There needs to be a way to include the hostname in this.
Ideally something like %h that expands to it, and
a default string like "Music Player on %h".

As it stands, people are very unlikely to change the default,
which is fine until they end up on a network with more than
one mpd. Note that zeroconf networks are not planned; they evolve.
So this will probably happen to any portable device running mpd
eventually. The result is that such networks end up with

```
Music Player Deamon
    Music Player
    Music Player #2
```

etc, which is not very clear.

If you look at other servers that support zeroconf, they invariably do
so by including some form of the hostname in the service name. For
example, my network has:

```
Workstation
    gnu
    dodo
    penguin
    box

Internet Printer
    HP Deskjet @ penguin
```

fschlich:
expand %h to the system hostname and default to a zeroconf string that
contains %h

### DIFF
--- a/doc/mpd.conf.5
+++ b/doc/mpd.conf.5
@@ -99,7 +99,8 @@ information will be published with Zeroconf.  The default is yes.
 .B zeroconf_name <name>
 If Zeroconf is enabled, this is the service name to publish.  This name should
 be unique to your local network, but name collisions will be properly dealt
-with.  The default is "Music Player".
+with.  The default is "Music Player @ %h", where %h will be replaced with the
+hostname of the machine running MPD.
 .TP
 .B audio_output
 See \fBDESCRIPTION\fP and the various \fBAUDIO OUTPUT PARAMETERS\fP sections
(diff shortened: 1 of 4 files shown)


fschlich:
Please check carefully and mangle as required, this is meant as a suggestion more than a fully-developed feature. I have very little experience with c++ and this may contain stupid mistakes!

MaxKellermann:
Using `std::regex` bloats the MPD executable by more than 100 kB (due to a badly designed libstdc++ which compiles a copy of the regex engine into every program, instead of having one copy in the shared library). I'd like to avoid that. In ncmpc, I've decided to use libpcre instead of `std::regex` only for that reason.

PR #406, opened 2018-11-04
doc/protocol.rst: update information about volume since -1 is not sent anymore

nroi:
Since 9cc37bdea2a55ca022c92902f943443badb94bce, MPD's status command does not emit the value -1 for the volume if the player is stopped. Instead, volume information is completely absent from the response sent by MPD.

Also, two backticks were missing.

### DIFF
--- a/doc/protocol.rst
+++ b/doc/protocol.rst
@@ -395,7 +395,7 @@ Querying ``MPD``'s status
     - ``consume`` [#since_0_15]_: ``0`` or ``1``
     - ``playlist``: 31-bit unsigned integer, the playlist version number
     - ``playlistlength``: integer, the length of the playlist
-    - ``state``: ``play``, ``stop, or ``pause``
+    - ``state``: ``play``, ``stop``, or ``pause``
     - ``song``: playlist song number of the current song stopped on or playing
     - ``songid``: playlist songid of the current song stopped on or playing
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Your documentation change breaks backwards compatibility with older versions. If somebody writes a client according to your text, and receives a `-1`, what wil it do?

nroi:
I assumed that the protocol specification always refers to the most recent stable version.

The reason why I submitted this PR is because I maintain a client library which worked well until the recent upgrade to version MPD v0.21. The client library was written with the assumption that there is *always* a line in the form of `volume: <integer>` sent as response to `status`. If the protocol documentation must support older versions of MPD, perhaps there is another place to document such changes? The `NEWS` file, maybe? Documenting it somewhere would make things easier for library maintainers.

MaxKellermann:
Your change removes something from the documentation which is deprecated, but keeping this rule in client code is a good idea even if it's deprecated, because it avoids breaking backwards compatibility. I'd agree to change the wording to mention that `-1` is deprecated, but not remove it completely from the document.
Your argument is not wrong, it's just that I value backwards compatibility more than keeping documentation clean of deprecated things. This is my subjective choice, not the absolute truth.
I've cherry-picked your first commit, but the second isn't a good improvement. Anyway thanks for far for caring about the documentation, we need more of that!

PR #435, opened 2018-12-09
Add meson dependencies

jacobvosmaer:
I couldn't get MPD to build on macOS with MacPorts until I added these. I am not a Meson expert and I don't know if what I'm doing here makes sense.

### DIFF
--- a/src/db/meson.build
+++ b/src/db/meson.build
@@ -45,6 +45,9 @@ db_glue = static_library(
   'db_glue',
   db_glue_sources,
   include_directories: inc,
+  dependencies: [
+    boost_dep,
+  ],
 )
 
 db_glue_dep = declare_dependency(
--- a/src/event/meson.build
+++ b/src/event/meson.build
@@ -16,6 +16,9 @@ event = static_library(
   'Thread.cxx',
   'Loop.cxx',
   include_directories: inc,
+  dependencies: [
+    boost_dep,
+  ],
 )
 
 event_dep = declare_dependency(
--- a/src/input/meson.build
+++ b/src/input/meson.build
@@ -7,6 +7,9 @@ input_api = static_library(
   'AsyncInputStream.cxx',
   'ProxyInputStream.cxx',
   include_directories: inc,
+  dependencies: [
+    boost_dep,
+  ],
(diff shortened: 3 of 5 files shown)


MaxKellermann:
What was the problem?

jacobvosmaer:
Thanks for asking @MaxKellermann , I forgot to give you enough context.

I got failures like this:

```
[5/64] Compiling C++ object 'src/event/src@event@@event@sta/DeferEvent.cxx.o'.
FAILED: src/event/src@event@@event@sta/DeferEvent.cxx.o 
c++ -Isrc/event/src@event@@event@sta -Isrc/event -I../src/event -Isrc -I../src -I. -I../ -Xclang -fcolor-diagnostics -pipe -Wall -Winvalid-pch -Wnon-virtual-dtor -std=c++14 -g -Wall -Wextra -fvisibility=hidden -ffast-math -ftree-vectorize -fno-threadsafe-statics -fmerge-all-constants -Wmissing-declarations -Wshadow -Wpointer-arith -Wcast-qual -Wwrite-strings -Wsign-compare -Wno-non-virtual-dtor -Wno-noexcept-type -funwind-tables -D_GNU_SOURCE  -MD -MQ 'src/event/src@event@@event@sta/DeferEvent.cxx.o' -MF 'src/event/src@event@@event@sta/DeferEvent.cxx.o.d' -o 'src/event/src@event@@event@sta/DeferEvent.cxx.o' -c ../src/event/DeferEvent.cxx
In file included from ../src/event/DeferEvent.cxx:20:
../src/event/DeferEvent.hxx:25:10: fatal error: 'boost/intrusive/list_hook.hpp' file not found
#include <boost/intrusive/list_hook.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Mostly for boost, one time for pcre.

jacobvosmaer:
I should add that meson struggled a bit with my boost to begin with. I had to set `BOOST_ROOT=/opt/local` both for meson and ninja (because ninja may dive back into meson).

I still don't really understand why I needed this in my environment while few others (apparently) do.

PR #444, opened 2018-12-27
Haiku fixes

extrowerk:
Various fixes for the build system to get the MPD build with meson on Haiku host

### DIFF
--- a/src/haiku/meson.build
+++ b/src/haiku/meson.build
@@ -2,8 +2,8 @@ rc = find_program('rc')
 xres = find_program('xres')
 
 rsrc = custom_target(
-  'mpd.rsrc',
-  output: 'mpd.rsrc',
+  'mpd2.rsrc',
+  output: 'mpd2.rsrc',
   input: 'mpd.rdef',
   command: [rc, '-q -o', '@OUTPUT@', '@INPUT@'],
 )
--- a/src/haiku/meson.build
+++ b/src/haiku/meson.build
@@ -1,11 +1,11 @@
-rc = meson.find_program('rc')
-xres = meson.find_program('xres')
+rc = find_program('rc')
+xres = find_program('xres')
 
 rsrc = custom_target(
   'mpd.rsrc',
   output: 'mpd.rsrc',
   input: 'mpd.rdef',
-  command: [rc, '-o', '@OUTPUT@', '@INPUT@'],
+  command: [rc, '-o -q', '@OUTPUT@', '@INPUT@'],
 )
 
 custom_target(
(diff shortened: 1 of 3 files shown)


extrowerk:
With this fixes MPD builds fine,but while it compiles the Haiku output and mixer parts, somehow it they are missing from the binary:
```
Output plugins:
 null fifo pipe openal httpd recorder
```

extrowerk:
RSRC still missing, the gen fix is just my stupidity.

extrowerk:
It seems libmixer_plugins.a have the required Haiku symbols, but the generated executable doesn't.

src/haiku/meson.build
```diff
@@ -1,11 +1,11 @@
...
-xres = meson.find_program('xres')
+rc = find_program('rc')
+xres = find_program('xres')
 
 rsrc = custom_target(
   'mpd.rsrc',
   output: 'mpd.rsrc',
   input: 'mpd.rdef',
-  command: [rc, '-o', '@OUTPUT@', '@INPUT@'],
+  command: [rc, '-o -q', '@OUTPUT@', '@INPUT@'],
```
MaxKellermann: What's this? Unrelated? Does this even work, adding two options separated by a space in one argument?

MaxKellermann:
I've cherry-picked two simple commits, but the others confuse me a bit. Can you please rebase, and fold those patches together which are just fixups for previous unmerged commits?

extrowerk:
Will do.

src/net/IPv6Address.hxx
```diff
@@ -53,7 +53,10 @@ class IPv6Address {
 						   uint16_t c, uint16_t d,
 						   uint16_t e, uint16_t f,
 						   uint16_t g, uint16_t h) noexcept {
-		return {{{
+		return {{
```
alaviss:
```suggestion
		return { .s6_addr = {
```

This should be the most portable option as defined in the POSIX specification.

MaxKellermann: This isn't valid C++. Designated initializers are C99 (and will maybe come to C++ with C++20).

MaxKellermann: This is even more portable, and valid C++14: https://github.com/MusicPlayerDaemon/MPD/commit/1fa99da3c254a6a641bdf4eca35d781281c4c268

mmuman:
Ugh, I just noticed this PR now after redoing it all myself… bleh :^)

MaxKellermann:
So this one is obsolete now? If any pieces are missing, please submit a new PR.

PR #454, opened 2019-01-13
Fix link_args for mDNSResponder on non-darwin platforms

riggs-:
The `-framework` linker argument is unavailable on non-darwin platforms, leading to linker errors, e.g. with LLD. On other platforms the default link argument `-ldns_sd` works fine, though.


MaxKellermann:
Curious: what non-Darwin platform supports the Bonjour API? Obviously Avahi with its Bonjour API emulation, but we don't want to use that.

PR #459, opened 2019-01-19
shout output plugin: add support for TLS

JakobOvrum:
Adds an optional config option for the "shout" output to support use of libshout's `shout_set_tls` functionality. Defaults to "disabled", which is the same as the behavior before this patch: no attempt at TLS will be made.

To use with icecast, set `<ssl>1</ssl>` on a listen socket in the icecast configuration. When testing, I was successful in getting it to work using "auto", "auto_no_plain" and "rfc2818" modes with the [icecast 2.5 beta](https://ftp.osuosl.org/pub/xiph/releases/icecast/icecast-2.5-beta2.tar.gz). I also tested with icecast 2.4.3 with the same config but couldn't get any of the TLS modes to work.

[This Xiph wiki page](https://wiki.xiph.org/Icecast_Server/known_https_restrictions) seems to claim that TLS should work since icecast 2.4.1 but also that icecast 2.5 contains a bunch of TLS-related fixes, FWIW.

For testing with a self-signed cert I used OpenSSL's SSL_CERT_FILE environment variable. libshout has functions `shout_set_ca_directory` and `shout_set_ca_file` which could be used for the same purpose, but this patch doesn't support them. It appears that libshout can also support client certificate authentication through `shout_set_client_certificate` but I haven't tested it and it's not supported by this patch.

### DIFF
--- a/doc/plugins.rst
+++ b/doc/plugins.rst
@@ -974,6 +974,8 @@ You must set a format.
      - Set the timeout for the shout connection in seconds. Defaults to 2 seconds.
    * - **protocol icecast2|icecast1|shoutcast**
      - Specifies the protocol that wil be used to connect to the server. The default is "icecast2".
+   * - **tls disabled|auto|auto_no_plain|rfc2818|rfc2817**
+     - Specifies what kind of TLS to use. The default is "disabled" (no TLS).
    * - **mount URI**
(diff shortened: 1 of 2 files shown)


MaxKellermann:
btw. this caused a build breakage, see 754f4048a8ab2e4c698ef74c042dd839f369a47b

Issue #405, opened 2018-11-04
mpd 0.21: meson: missing config option for tremor (libvorbisidec)


joerg-krause:
Commit 94592c14062d5afc9482d11baa401648082022c0 dropped autotools in favour of meson build system. Unfortunately, support for the tremor (libvorbisidec) got lost by this change.

MaxKellermann:
That was intentional, even though I did not document it. It looked like Tremor was mostly unmaintained (last release 1.0.2 was 16 years ago), and I had never heard of anybody using it. When it was developed, there were many low-end CPUs without FPU, but 6 years ago, even the very first Raspberry Pi had an FPU.
Do you believe Tremor support in MPD is still important?

riggs-:
I wouldn't mind dropping it. Even the very low-end CPUs are fine decoding with the floating point libvorbis.

joerg-krause:
tremor is still maintained, it received the last commit some month ago: http://git.xiph.org/?p=tremor.git.

There are still a lot of soft-floating CPUs around, so I would say, yes, it is still important for those CPUs. I am using an i.MX28-based board myself. This CPUs has a longevity until late 2020.

I don't agree, that it is decoding fine with libvorbis. It requires a lot more CPU usage, making it impossible to do other stuff, while decoding vorbis streams.

Furthermore, it doesn't cause any trouble to still keep it, right?

MaxKellermann:
If there is a maintainer, the code will stay. Currently, there is none.
I'm already overloaded with work and too few people are helping with MPD/libmpdclient/mpc/ncmpc development, I didn't want to spend any time on bringing Tremor support to the new Meson build system. So anybody interested in Tremor support should become maintainer, do the Meson rules, and probably write a Travis-CI rule to do a Tremor-enabled build.

joerg-krause:
I see, thanks! I will have a look and maybe give it a try...

PR #460, opened 2019-01-21
meson: add fixed-point Vorbis (Tremor) decoder support

joerg-krause:
Re-add build support for the fixed-point Vorbis (Tremor) decoder, which was dropped when switching from Autotools to Meson.

Note, that it is not possible to build both, the Vorbis and the Tremor decoder.

Closes: #405


MaxKellermann:
This PR caused a crash bug, see commit c18cd941aaa4c16ac713bc4f1e16ada5281d689e

Issue #465, opened 2019-01-26
Bug playlist filename with zero length


alandrees:
Client: ncmpcpp 0.8.2
Server: 0.21.4
Distro: Debian (testing)

Steps to Reproduce
--------------------------
1. Create a playlist file called `.m3u`
2. Connect to mpd as a client (I have been using nc for testing and just issuing commands directly)
3. Issue an `lsinfo` command for the root music directory

### Observed Results
In the file listing which is returned there is a playlist item which has no playlist name, and will look similar to:
```
playlist:
Last-Modified: <ISO-8601 Timestamp>
```
 
Background
----------------
Up until recently I had been using the 0.19.x, and everything had worked fine.  I did an update to 0.21.4, and noticed that I could no longer browse my music directory in ncmpcpp, all the other features worked fine, but browsing would actually cause the application to exit.

I wound up getting this error, with the application aborting on exception:

```
terminate called after throwing an instance of 'MPD::ClientError'
what():  Malformed entity response line
Aborted
```

Eventually I figured out that when I was doing a root music directory request it was returning a playlist like:

```
lsinfo
directory: flac
Last-Modified: 2016-06-30T14:09:14Z
directory: mp3
Last-Modified: 2017-07-04T09:41:06Z
directory: wav
Last-Modified: 2016-06-30T14:09:28Z
playlist: 
Last-Modified: 2015-04-03T10:35:14Z
OK
```

Checking the playlist directory I found a lone playlist with the name '.m3u'.

The reason it was unprocessable was because it expected a label/name for the playlist, and it was probably splitting on last instance of '.'

Conclusion
---------------
After some code spelunking:

From here:
https://github.com/MusicPlayerDaemon/MPD/blob/master/src/command/OtherCommands.cxx#L179-L184

To Here:
https://github.com/MusicPlayerDaemon/MPD/blob/master/src/PlaylistFile.cxx#L157-L175

Finally here:
https://github.com/MusicPlayerDaemon/MPD/blob/master/src/PlaylistFile.cxx#L157-L175

I found that doing a check for name length to be sufficient to stop the problem from happening:

```
	try {
		info.name = name.ToUTF8Throw();
	} catch (...) {
		return false;
        }

+        if (info.name.empty())
+                return false;
```

Created PR #466

PR #466, opened 2019-01-26
Fixed playlist filename zero-length bug

alandrees:
This is a fix for a bug which happens when returning playlists that are named '.m3u' via the `lsinfo` command.

### DIFF
--- a/src/PlaylistFile.cxx
+++ b/src/PlaylistFile.cxx
@@ -149,6 +149,9 @@ LoadPlaylistFileInfo(PlaylistInfo &info,
 	} catch (...) {
 		return false;
 	}
+	
+	if (info.name.empty())
+		return false;
 
 	info.mtime = fi.GetModificationTime();
 	return true;


MaxKellermann:
Which bug?

alandrees:
#465 

tl;dr 
Playlist files which are named .m3u cause the entry for that playlist file to be malformed due to the 'name' being blank, so you'd have a playlist entry like:

playlist: 
Last-Modified: 2015-04-03T10:35:14Z

This trips up libmpdclient's validation.

MaxKellermann:
But the crash in #465 is a client bug.

a-drees:
Then the server generates valid MPD protocol entities in this case?

If that's the case, this is for sure the wrong place to fix this, and the validation should be changed in the client library.

alandrees:
If that's valid output from the server side, then the fix should be made to the libmpdclient code instead.

I guess it depends on the spec for the mpd protocol.

MaxKellermann:
62a129c18f56b3199c938f2cfabd58d344cfe513

PR #488, opened 2019-02-22
Unix Sockets audio output

terorie:
__WIP, don't review/merge__

 * Adds a new audio output plugin (`sock`)
 * Will be available on BSDs, Linux and macOS
 * Like, `fifo` but supports multiple clients

### Motivation

Most MPD visualizers work with the `fifo` audio output right now.
Imo, `fifo` has some limitations
 * Only one visualizer per named pipe
 * No way to transmit sample rate/stream config

Unix Sockets would fix those limitations and are fairly easy to implement.
Of course, client-side support will be needed.
I'm working on client-side support for `sock` on as many visualizers I can find.

### Features

This PR is based on `src/output/plugins/FifoOutputPlugin.cxx` and will aim to implement these features
 - [x] New audio output plugin (Name  `sock`/`SockOutputPlugin`)
 - [x] `SOCK_STREAM` socket in file system
 - [ ] Deletion of old/existing sockets on start
 - [ ] Raw audio stream to every connected client
 - [ ] Buffer overrun handling (Client reads too slow)
 - [ ] Buffer underrun handling? (Lag/Server misses a write)
 - [ ] Implement stream encoding

For maximum platform compatibility, the actual streaming is done over `write/O_NONBLOCK` by iterating over all FDs in a single thread. (As opposed to `epoll/kqueue`, or multithreading).
The stream is read-only. No `read` calls are made.

### DIFF
--- a/src/output/Registry.cxx
+++ b/src/output/Registry.cxx
@@ -23,6 +23,7 @@
 #include "plugins/AlsaOutputPlugin.hxx"
 #include "plugins/AoOutputPlugin.hxx"
 #include "plugins/FifoOutputPlugin.hxx"
+#include "plugins/SockOutputPlugin.hxx"
 #include "plugins/SndioOutputPlugin.hxx"
 #include "plugins/httpd/HttpdOutputPlugin.hxx"
 #include "plugins/HaikuOutputPlugin.hxx"
@@ -51,6 +52,9 @@ const AudioOutputPlugin *const audio_output_plugins[] = {
 #endif
 #ifdef HAVE_FIFO
 	&fifo_output_plugin,
+#endif
+#ifdef HAVE_SOCK
+  &sock_output_plugin,
 #endif
 #ifdef ENABLE_SNDIO
 	&sndio_output_plugin,
--- a/src/output/plugins/SockOutputPlugin.cxx
+++ b/src/output/plugins/SockOutputPlugin.cxx
@@ -1,0 +1,187 @@
+#include "SockOutputPlugin.hxx"
+#include "../OutputAPI.hxx"
+#include "../Timer.hxx"
+#include "fs/AllocatedPath.hxx"
+#include "fs/FileSystem.hxx"
+#include "fs/FileInfo.hxx"
+#include "util/Domain.hxx"
+#include "util/RuntimeError.hxx"
+#include "Log.hxx"
+#include "open.h"
+
+#include <sys/socket.h>
+#include <sys/stat.h>
+#include <sys/un.h>
+#include <errno.h>
+#include <unistd.h>
+
+class SockOutput final : AudioOutput {
+  const AllocatedPath path;
+  std::string path_utf8;
+
+  int fd = -1;
+  bool created = false;
+  Timer *timer;
+
+public:
+  SockOutput(const ConfigBlock &block);
+
+  ~SockOutput() {
+    CloseSock();
+  }
+
+  static AudioOutput *Create(EventLoop &,
+            const ConfigBlock &block) {
+    return new SockOutput(block);
+  }
+
+private:
+  void Create();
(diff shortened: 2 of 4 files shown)


terorie:
Considering Stream Encoding

 * __Raw PCM__
    * _Pros_: No overhead, easy to code
    * _Cons_: Unknown sample rate & bit format, annoying to use
 * __WAV__
    * _Pros_: Widely supported format, possibility for `nc sock | visualize --wav`
    * _Cons_: Stream length unknown but is required; `ChunkSize` field is 32-bit (max stream length is 6,5 hrs CD quality)
 * __Matroska__ or __OggPCM__
    * _Pros_: Standard-conformant, perfect `ffmpeg`, `dd` (streaming) support
    * _Cons_: Potentially very complex to parse, less supported than WAV (ngl OGG is pretty basic though)
    * Might as well add a ffmpeg audio output type instead
 * __Custom__
    * _Pros_: Tiny overhead, easy to code
    * _Cons_: Non-standard, no support for other tools

Opinions on this?

MaxKellermann:
Since you asked for an opinion: why not use `httpd` instead of writing a new plugin? `httpd` can already do everything on your TODO list (and more), and it has been existing for quite a number of years.

Your draft code is not only redundant feature-wise, but also duplicates lots of code, for example by implementing socket listener code from scratch, instead of reusing the libraries already existing in MPD.

There's so much to do in mPD, and I wish you'd write useful new code instead of redoing something that already exists.

terorie:
@MaxKellermann Thanks for the feedback! I’m closing this PR as it’s considered redundant.

> why not use httpd instead

HTTP/TCP overhead feels a bit large just for inter-process streaming, which is why some visualizers don’t support it (yet).

> Your draft code is not only redundant feature-wise

I have to disagree. `fifo` has problems I pointed out earlier, and HTTP streaming is a very large overhead for simple visualizers. Imo, `sock` is the missing gap between the two.

> but also duplicates lots of code, for example by implementing socket listener code from scratch

The code is a minimal implementation of the audio output interface. Reusing the code from fifo is not suitable, as fifo has different logic. Using other socket listeners is a good idea for conventional streaming, but I was planning to implement socket-only streaming with very minimal buffering.

> There's so much to do in mPD, and I wish you'd write useful new code instead of redoing something that already exists.

Sadly, I don’t have time nor know the codebase of MPD well enough. I figured the sock listener I need could be useful for other people too, which is why I made the PR.

MaxKellermann:
> HTTP/TCP overhead feels a bit large just for inter-process streaming, which is why some visualizers don’t support it (yet).

TCP overhead: wrong, the `httpd` plugin can bind to local sockets.
HTTP overhead: (mostly) wrong, because the `httpd` plugin has **no overhead at all** after response headers have been sent. It's a raw stream from here.

> HTTP streaming is almost impossible to implement without relying on additional libraries

Really? Send `GET /\n\n` and wait for an empty line, start decoding from there. Sounds not so impossible.

> Using other socket listeners is a good idea for conventional streaming, but I was planning to implement socket-only streaming with very minimal buffering.

This sounds like you don't know the difference between a listener socket and a connection socket.

But my complaint about duplicating code was just the start; it was only about code you had already written (which was very little). Going further, you'd have had to solve lots of complicated problems, problems which were already solved by the `httpd` plugin many years ago.

> I don’t have time

That's why I would not like to see you waste time with solving problems which were already solved by somebody else, repeating old mistakes. And making new ones - yes, I found trivial bugs in your code which you made because you tried to write redundant code from scratch instead of reusing existing code...

Now that I've convinced you to abandon this idea, you have lots of time left which you would have put into this plugin. Please leave your mark, and do something useful with that time I saved you.

terorie:
I will continue developing it on my fork

Issue #455, opened 2019-01-13
[Request] allow loading playlist from absolute path


miccoli:
Under version 0.21.4 the `load {NAME} [START:END]` command will lookup

1. `{NAME}.m3u` relative to the playlist directory
1. `{NAME}` relative to the music directory

and fail if `{NAME}` is absolute (starting with `/`).

It would be very useful to be able to load an absolute playlist path, say `/tmp/tmp.bnsUAw5rXO/dynamic.pls`, when the command is issued from a local client. (Usage case: query some web-radio directory, save returned playlist, load in MPD queue.)

Since local clients are able to load absolute sound files with the `add` command, I was wondering if there is some reason (security issues?) for forbidding to load absolute playlist files.

miccoli:
I'm having a try at implementing this feature: miccoli@e2b289705ad8fb1afca8b8f6a3bae6cfd4b72151

After some testing I will open a PR. If there is no interest for this feature, please feel free to close this issue.

PR #492, opened 2019-02-24
Allow local clients to load playlists in arbitrary absolute location

miccoli:
Adresses #455.

Before this patch, 
```
load /foo/bar.m3u
```
always fails with `ACK [50@0] {load} No such playlist`

With this patch we have
- remote clients: always fails with `ACK [4@0] {} Access denied`
- local clients:
    - success `OK` if the files exists
    - otherwise fails `ACK [52@0] {} Failed to access /foo/bar.m3u: No such file or directory`.

Behaviour for relative paths and remote URI is not changed.

### DIFF
--- a/src/playlist/PlaylistAny.hxx
+++ b/src/playlist/PlaylistAny.hxx
@@ -34,7 +34,7 @@ class Storage;
  * music or playlist directory.
  */
 std::unique_ptr<SongEnumerator>
-playlist_open_any(const char *uri,
+playlist_open_any(const LocatedUri &located_uri,
 #ifdef ENABLE_DATABASE
 		  const Storage *storage,
 #endif
--- a/src/playlist/Print.hxx
+++ b/src/playlist/Print.hxx
@@ -34,6 +34,6 @@ struct Partition;
 bool
 playlist_file_print(Response &r, Partition &partition,
 		    const SongLoader &loader,
-		    const char *uri, bool detail);
+		    const LocatedUri &uri, bool detail);
 
 #endif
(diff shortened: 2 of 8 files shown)


src/playlist/PlaylistAny.hxx
```diff
@@ -21,6 +21,7 @@
 #define MPD_PLAYLIST_ANY_HXX
 
 #include "config.h"
+#include "src/LocateUri.hxx"
```
MaxKellermann: Use forward declarations instead of adding an include dependency in a header.

src/playlist/PlaylistAny.hxx
```diff
@@ -34,7 +35,7 @@ class Storage;
  * music or playlist directory.
  */
 std::unique_ptr<SongEnumerator>
-playlist_open_any(const char *uri,
+playlist_open_any(const struct LocatedUri &located_uri,
```
MaxKellermann: Don't use `struct` here - it's not necessary in C++.

src/playlist/Print.hxx
```diff
@@ -20,6 +20,8 @@
 #ifndef MPD_PLAYLIST__PRINT_HXX
 #define MPD_PLAYLIST__PRINT_HXX
 
+#include "LocateUri.hxx"
```
MaxKellermann: Forward-declaration - see above.

src/playlist/PlaylistAny.hxx
miccoli: Fixed here and in other places.

src/playlist/PlaylistAny.hxx
miccoli: Fixed here and in other places

src/playlist/Print.hxx
miccoli: Fixed

miccoli:
@MaxKellermann thanks for reviewing.
I force pushed a new patch that I hope should address all concerns.

PR #520, opened 2019-03-27
macOS: fix Bonjour build

Nyx0uf:


MaxKellermann:
Please improve the commit message, explaining why suddenly that other framework is needed instead of "dnssd".

Nyx0uf:
I don't know why, nor since when, I did not find any explanation.

There's no more dnssd framework, instead there is a `/usr/lib/system/libsystem_dnssd.dylib` but we can't directly link to it because it yields an error saying we have to link with `System.framework`.

MaxKellermann:
There must be some documentation explaining which library shall be linked. I don't know that, because I don't have any Apple products. But I worry about having things in the MPD repository which nobody understands, so I'd like to have explanations in commits like that.

Nyx0uf:
Yeah I understand. I searched on Apple dev forums, didn't find.
The only doc I found don't mention what to link : https://developer.apple.com/documentation/dnssd
I also looked at the `dns_sd.h` header file but there's nothing more there :/

Since when Zeroconf has been disabled for macOS ?

MaxKellermann:
Can't review this PR without pointers to Apple documentation.

PR #529, opened 2019-04-06
Fix move doc args

cotko:
I think this is more correct?


MaxKellermann:
1bc78e9f2c9e459a52ca7fa2d44a59b297a86d0d

PR #536, opened 2019-04-16
Fix meson.build to work properly with '-Ddatabase=false'

eugene2k:
Currently building with '-Ddatabase=false' fails. This fix rectifies that.


MaxKellermann:
Cherry-picked to v0.21.x 8d18b4c24b776fe62c39ad9ce72e1e57ac0aa3ea

PR #537, opened 2019-04-16
Make ffmpeg input plugin handle http/s sources if curl is disabled

eugene2k:
ffmpeg is capable of reading http and https sources, however MPD doesn't allow to use HTTP(S) sources unless MPD is compiled with libcurl as a dependency. This removes the need to compile MPD with libcurl if ffmpeg is enabled.


eugene2k:
superseeded by #538

PR #538, opened 2019-04-17
Runtime schema enumeration support

eugene2k:
This PR adds support for enumerating input schemas at runtime. Basically, this is needed because FFMPEG libraries can be compiled with different flags and thus may or may not support gopher or rtmp, depending on the user's choice. It's a better solution, IMHO, than hardcoding supported protocols into MPD.

### DIFF
--- a/src/input/InputPlugin.cxx
+++ b/src/input/InputPlugin.cxx
@@ -21,15 +21,54 @@
 #include "util/StringCompare.hxx"
 
 #include <assert.h>
+#include <algorithm>
+#include <iterator>
 
 bool
 InputPlugin::SupportsUri(const char *uri) const noexcept
 {
-	assert(prefixes != nullptr);
-
-	for (auto i = prefixes; *i != nullptr; ++i)
-		if (StringStartsWithIgnoreCase(uri, *i))
-			return true;
-
+	assert(prefixes || protocols);
+	if (prefixes != nullptr) {
+		for (auto i = prefixes; *i != nullptr; ++i)
+			if (StringStartsWithIgnoreCase(uri, *i))
+				return true;
+	} else {
+		for (auto schema : protocols()) {
+			if (StringStartsWithIgnoreCase(uri, schema.c_str())){
+				return true;
+			}
+		}
+	}
 	return false;
+}
+
+// Note: The whitelist has to be ordered alphabetically
+constexpr static const char *whitelist[] = {
+	"ftp",
+	"ftps",
+	"gopher",
+	"http",
+	"https",
+	"mmsh",
+	"mmst",
+	"rtmp",
+	"rtmpe",
+	"rtmps",
+	"rtmpt",
+	"rtmpte",
+	"rtmpts",
+	"rtp",
+	"scp",
+	"sftp",
+	"smb",
+	"srtp",
+};
+
+bool
+protocol_is_whitelisted(const char *proto) {
+	auto begin = std::begin(whitelist);
+	auto end = std::end(whitelist);
+	return std::binary_search(begin, end, proto, [](const char* a, const char* b) {
+		return strcasecmp(a,b) < 0;
+	});
 }
--- a/src/input/InputPlugin.hxx
+++ b/src/input/InputPlugin.hxx
@@ -22,6 +22,9 @@
 
 #include "Ptr.hxx"
 #include "util/Compiler.h"
+#include <assert.h>
+#include <set>
+#include <string>
 
 struct ConfigBlock;
 class Mutex;
@@ -63,6 +66,11 @@ struct InputPlugin {
 	InputStreamPtr (*open)(const char *uri, Mutex &mutex);
 
 	/**
+	 * return a set of supported protocols
+	 */
+	std::set<std::string> (*protocols)();
+
+	/**
 	 * Prepare a #RemoteTagScanner.  The operation must be started
 	 * using RemoteTagScanner::Start().  Returns nullptr if the
 	 * plugin does not support this URI.
@@ -76,6 +84,25 @@ struct InputPlugin {
 
 	gcc_pure
 	bool SupportsUri(const char *uri) const noexcept;
+
+	template<typename F>
+	void ForeachSupportedUri(F lambda) const noexcept {
+		assert(prefixes || protocols);
+
+		if (prefixes != nullptr) {
+			for (auto schema = prefixes; *schema != nullptr; ++schema) {
+				lambda(*schema);
+			}
+		}
+		if (protocols != nullptr) {
+			for (auto schema : protocols()) {
+				lambda(schema.c_str());
+			}
+		}
+	}
 };
+
+bool
+protocol_is_whitelisted(const char *proto);
 
 #endif
--- a/src/input/plugins/FfmpegInputPlugin.cxx
+++ b/src/input/plugins/FfmpegInputPlugin.cxx
@@ -25,8 +25,13 @@
 #include "lib/ffmpeg/Init.hxx"
 #include "lib/ffmpeg/Error.hxx"
 #include "../InputStream.hxx"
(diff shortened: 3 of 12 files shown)


src/input/InputPlugin.cxx
```diff
@@ -25,11 +25,18 @@
...
-		if (StringStartsWithIgnoreCase(uri, *i))
-			return true;
-
+	assert(prefixes || enumerate);
+	if (prefixes != nullptr) {
+		for (auto i = prefixes; *i != nullptr; ++i)
+			if (StringStartsWithIgnoreCase(uri, *i))
+				return true;
+	} else {
+			void* iter = nullptr;
```
MaxKellermann: Wrong indent

src/input/InputPlugin.hxx
```diff
@@ -62,6 +63,11 @@ struct InputPlugin {
 	 */
 	InputStreamPtr (*open)(const char *uri, Mutex &mutex);
 
+	/**
+	 * Dynamically enumerate supported protocols
+	 */
+	const char* (*enumerate)(void** opaque);
```
MaxKellermann: I don't like this API, because I don't like `void` pointers. And your API isn't properly documented:
- how shall this `opaque` parameter be initialized?
- when are you allowed to call this function? Does it require initializing the plugin first?
- can you leave the enumeration loop early, or does this result in memory leaks?
- can you store an `opaque` pointer and reuse it later? Can you copy it?
- what is the lifetime of the returned pointer? (yes, I'm talking about your global variable hack in the FFmpeg plugin...)
This is error-prone to implement and error-prone to use.

Let's design something which is safe to use. For example:
```
std::set<std::string> (*get_prefixes)();
```
This suffers from the usual C++ bloat, but is easy to understand and is nearly impossible to go wrong.

src/input/plugins/FfmpegInputPlugin.cxx
```diff
@@ -71,6 +72,66 @@ input_ffmpeg_init(EventLoop &, const ConfigBlock &)
...
+is_whitelisted(const char *str) {
+	for (auto i = whitelist; *i != nullptr; ++i)
+		if (StringIsEqualIgnoreCase(*i, str))
+			return true;
+	return false;
+}
+
+static const char*
+input_ffmpeg_enumerate(void **opaque) {
+	static char protocol_name[get_max_len() + 4];
```
MaxKellermann: No global variables, please!

src/ls.cxx
```diff
@@ -22,28 +22,41 @@
...
 {
 #ifdef HAVE_UN
 	fprintf(fp, " file://");
 #endif
+	std::set<std::string> protocols;
 	input_plugins_for_each(plugin)
-		for (auto i = plugin->prefixes; *i != nullptr; ++i)
-			fprintf(fp, " %s", *i);
+		plugin->ForeachSupportedUri([&](const char* uri) {
+			protocols.insert(uri);
```
MaxKellermann: `emplace()` is a little bit better.

eugene2k:
I totally forgot about needing to have a proper interface that not only ffmpeg could use, sorry. Hopefully, this latest commit changes things for the better.

MaxKellermann:
This PR now suddenly contains the commits of your other pull request; they don't belong here, they're already merged in v0.21.x and will soon be merged into master. And if you correct an unmerged commit, don't add another commit - amend the original commit. This PR is very hard to review this way.

eugene2k:
Well, this is odd... I don't remember rebasing this branch yesterday...

eugene2k:
I think I came up with a less invasive approach, that involves creating a list at init and making `InputPlugin::prefixes` point to it by passing `prefixes` as an additional argument to `InputPlugin::init` and `InputPlugin::finish`. Working on it now, so please don't merge this yet.

eugene2k:
The 'less invasive' approach turned out to be more invasive, and required adding some sort of plugin registration function that would be called at the start of Main because the schemas are enumerated before plugins are initialized.

MaxKellermann:
Are you still working on this?

eugene2k:
No, I was waiting for you to review the current commit. Unless you plan on having dynamically loaded plugins doing more makes little sense.

MaxKellermann:
So your latest commit contains an abstract class with two virtual methods. This is tailored after FFmpeg's `avio_enum_protocols()` function, but is complicated to implement if the library happens to have a different method of enumerating its supported protocols. What was your reason not to implement my suggestion?

eugene2k:
I think a set is overkill for cases like ffmpeg, where you can be certain that if an element is not unique then it's a bug. You could just enumerate all the elements and put them inside an array to produce a list. At the same time, in a case where an input handler library actually produces duplicate elements, using a set makes sense. Using an iterator seems like a good way of abstracting the underlying implementation away.

When you say "complicated to implement if the library happens to have a different method of enumerating its supported protocols" do you have any examples in mind?

MaxKellermann:
> I think a set is overkill for cases like ffmpeg, where you can be certain that if an element is not unique then it's a bug. You could just enumerate all the elements and put them inside an array to produce a list.

Your own commit converts the iterator to `std::set`.

> Using an iterator seems like a good way of abstracting the underlying implementation away.

This isn't a good way at all. It's complicated to implement, error prone to use (e.g.: what's the lifetime of the returned string pointer?), and needs to be converted to `std::set` later anyway (as your own code demonstrates). You named no single advantage of an abstract iterator class.

> When you say "complicated to implement if the library happens to have a different method of enumerating its supported protocols" do you have any examples in mind?

Yes, for example this one: https://curl.haxx.se/libcurl/c/curl_version_info.html
Try to implement this using your iterator concept.

eugene2k:
>Your own commit converts the iterator to std::set.
My commit adds items from the iterator to a set when a set of unique protocols is needed. There is no need to store protocols of each plugin in individual sets.

>It's complicated to implement
I have yet to encounter complications  to be honest.
>error prone to use (e.g.: what's the lifetime of the returned string pointer?)
That's not a property of the iterator. The iterator could just as easily made to return std::string.

>and needs to be converted to std::set later anyway (as your own code demonstrates).
And an implementation that returns a set would need to add the items from each individual set to a combined set of protocols anyway. I don't see how returning a set is useful.

>Yes, for example this one: https://curl.haxx.se/libcurl/c/curl_version_info.html
>Try to implement this using your iterator concept.
I did. I have no idea what you're talking about as `version_info::protocols` contains an array that could easily be iterated over. If you meant something like "supported protocols are stored as flags in a struct" then in that case all that needs to be done is to have the iterator creation function create an array to iterate over and store it inside the iterator structure. That's not complicated either.

I'll rewrite this to return a set of strings, but I still fail to see the efficiency.

eugene2k:
Please review.

MaxKellermann:
Build failure:
```
../../src/input/plugins/TidalInputPlugin.cxx:255:1: error: invalid conversion from ‘std::unique_ptr<RemoteTagScanner> (*)(const char*, RemoteTagHandler&)’ to ‘std::set<std::__cxx11::basic_string<char> > (*)()’ [-fpermissive]
../../src/input/plugins/QobuzInputPlugin.cxx:223:1: error: invalid conversion from ‘std::unique_ptr<RemoteTagScanner> (*)(const char*, RemoteTagHandler&)’ to ‘std::set<std::__cxx11::basic_string<char> > (*)()’ [-fpermissive]
```

eugene2k:
@MaxKellermann should be fixed now. The travis build script should probably configure the project to build with all plugins set to enabled.

MaxKellermann:
This PR caused a serious regression: https://github.com/MusicPlayerDaemon/MPD/issues/599
It looks like you never tested your code....

eugene2k:
Shouldn't the test suite run automatically on travis?

MaxKellermann:
It does, but you did not write a unit test.

PR #573, opened 2019-06-04
output/httpd: reject some well-known request paths

cathugger:
Return `404 not found` for some common well-known paths, as clients requesting them usually do that automatically and don't expect endless audio stram.
Also use strncmp instead of memcmp, as memcmp use may result in out of bounds access.

### DIFF
--- a/src/output/plugins/httpd/HttpdClient.cxx
+++ b/src/output/plugins/httpd/HttpdClient.cxx
@@ -71,10 +71,10 @@ HttpdClient::HandleLine(const char *line) noexcept
 	assert(state != State::RESPONSE);
 
 	if (state == State::REQUEST) {
-		if (memcmp(line, "HEAD /", 6) == 0) {
+		if (strncmp(line, "HEAD /", 6) == 0) {
 			line += 6;
 			head_method = true;
-		} else if (memcmp(line, "GET /", 5) == 0) {
+		} else if (strncmp(line, "GET /", 5) == 0) {
 			line += 5;
 		} else {
 			/* only GET is supported */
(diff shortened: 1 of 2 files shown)


src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -71,10 +71,10 @@ HttpdClient::HandleLine(const char *line) noexcept
 	assert(state != State::RESPONSE);
 
 	if (state == State::REQUEST) {
-		if (memcmp(line, "HEAD /", 6) == 0) {
+		if (strncmp(line, "HEAD /", 6) == 0) {
```
MaxKellermann: What does this have to do with rejecting well-known paths?

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -129,14 +141,21 @@ HttpdClient::SendResponse() noexcept
...
+			"404 not found";
+	} else if (metadata_requested) {
 		allocated =
 			icy_server_metadata_header(httpd.name, httpd.genre,
 						   httpd.website,
 						   httpd.content_type,
 						   metaint);
 		response = allocated.c_str();
-       } else { /* revert to a normal HTTP request */
+	} else { /* revert to a normal HTTP request */
```
MaxKellermann: Why a whitespace change here? What does this have to do with rejecting certain paths?

MaxKellermann:
The commit is bad because it mixes more than one change into one commit.

MaxKellermann:
f9ca2f52c17aa037c5cffb28b9052a4003cad36d

cathugger:
Oh, sorry, I guess it is, I just fixed whatever else bothered me while I was at it, but I guess putting that into single commit is worse for anyone later wanting to make sense of them.
Thanks for splitting it so I didn't need to.

Issue #578, opened 2019-06-08
DAC hiccup noise when seeking DSD over PCM


Wang-Yue:
Hi Max, I have a question. During MPD's DoP playback, when seeking in a track, I can hear a noise over essentially all USB DACs I have (either ESS or AKM based, either XMOS based or custom implementation). 

Latest version of MPD on both Linux and Mac all presented the same issue. Which means this may not be an issue in the output plugin, but an issue related to general DoP handling. 

A few other players has the same issue, such as Foobar2000. 
Some players are free of this problem, such as JRiver media center, or HQPlayer. 

I was wondering if you can give some insights on what exactly does MPD do when seeking a DoP track? Are the marker bits still in order? I assume if we have a locally noncontinuous marker bits stream the DAC would switch to PCM and then switch back, which would cause the hiccup noise.  I could be wrong. Are there any start/stop command send to USB DAC during seeking?

Wang-Yue:
I checked the DSD marker bits and they are right. The problem happens in Cancel(). If I clear the Cancel() function implementation everything works correctly on both platforms. So probably there're something wrong with DSD handling in Cancel().

MaxKellermann:
This is about a bug in some DoP-capable DACs, and not about a MPD bug, right?
If anything, then this demonstrates how bad the whole idea of DoP was, and why this bad kludge should be removed from MPD. Instead, people should use DSD directly.

Wang-Yue:
> If anything, then this demonstrates how bad the whole idea of DoP was, and why this bad kludge should be removed from MPD. Instead, people should use DSD directly.

I don't believe so. native DSD is not supported on platform such as Mac,  iOS, or various *NIX like systems (such as FreeBSD). On Linux it requires a kernel patch to support certain DAC's altsettings. On Windows it needs specific ALSA driver. 

And this is just USB. On TOSLINK/COAXIAL/etc only DoP is supported. 

In any cace, this is unrelated to the bug report. I personally hate DSD as it limits all DSP possibilities. I believe if we get rid of DSD altogether, the world would become a better place.  However this is also unrelated to the bug report. 

> This is about a bug in some DoP-capable DACs, and not about a MPD bug, right?

[...]

PR #579, opened 2019-06-08
Fix DoP hiccups on Mac OS X

Wang-Yue:
see https://github.com/MusicPlayerDaemon/MPD/issues/578
Here're a few findings:

whenever AudioOutputUnitStop and AudioOutputUnitStart is called, DACs will reset their state. This will surely cause hiccups for DoP, no matter we reset the buffers or not. 

I test the following Cancel 

```
void OSXOutput::Cancel() noexcept {
  AudioOutputUnitStop(au);
  AudioOutputUnitStart(au);
}
```

And that's already suffient to make the problem occur on various DACs (either AKM or ESS based chip. either XMOS or custom USB interface). 

In this [post](https://www.forum.rme-audio.de/viewtopic.php?pid=142008#p142008), Matthias Carstens from RME believes "a player that stops sending DSD during seek operation is flawed", when I told him many players "go through stop, empty the buffer, fill the buffer with new data, start" during seek. 

I understand the reasoning behind using Start/Stop in Cancel() --- we don't have assurance that OS will consume multiple of complete frames, so without start/stop, we have no idea how many packet are left. If OS is not consuming multiple of frame packets (in DoP's case it's 4 complete frame packets), the audio stream will be broken. So I have a reply [here](https://www.forum.rme-audio.de/viewtopic.php?pid=142026#p142026) . 

At current stage, because most DACs will change mode and produce noise when AudioOutputUnitStop/AudioOutputUnitStart is called, they should not be used in DoP's case. When no Start/Stop is used, it is then impossible to count how many packets are left and drop maximum of multiples of 4 complete frames, because it has to be an atomic operation. 

So in this case, we can do nothing about it. I would rather have the 0.1s latency here rather than a corrupted stream.

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -922,12 +922,14 @@ bool OSXOutput::Pause() {
 void
 OSXOutput::Cancel() noexcept
 {
-	AudioOutputUnitStop(au);
-	ring_buffer->reset();
 #ifdef ENABLE_DSD
-	pcm_export->Reset();
+        pcm_export->Reset();
+        if (!dop_enabled) {
+                ring_buffer->reset();
+        }
+#else
+        ring_buffer->reset();
 #endif
-	AudioOutputUnitStart(au);
 }
 
 int
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -923,13 +923,15 @@ void
 OSXOutput::Cancel() noexcept
 {
 #ifdef ENABLE_DSD
-        pcm_export->Reset();
-        if (!dop_enabled) {
-                ring_buffer->reset();
-        }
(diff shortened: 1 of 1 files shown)


MaxKellermann:
If a PR needs a text, this means that the commit messages are bad, and indeed they are.
I don't understand the two commits; the sequence doesn't make sense. The first one deletes the stop/start calls, and the second one re-adds them. That's confusing, that can't be right.
And there are no code comments.
Complicated rationales like the one in your PR text need to be made permanent, however GitHub PR texts are lost after merging.
Also, since this appears to solve a real problem, it should be applied to the stable branch (v0.21.x) instead of master.

Wang-Yue:
I will make a new commit with comment in code tomorrow. You can either wait I clean up the commit history, or you can close this one and I'll create a new one instead.

Wang-Yue:
Close this PR because we're going to do it via option to not stop/drop/start during Cancel()

PR #582, opened 2019-06-13

Wang-Yue:

- Have an atomic boolean monitoring the cancelling process. The Audio Unit is no longer stopped and restarted.
- Because the AU is no longer stopped,  in the render function, we always copy multiple of complete frames (in DoP even number of frames, to preserve the DoP marker bits pattern) into the hardware buffer.
- During the cancel process, the ring buffer only keeps a small portions of front side buffers, and the rest of the old data are dropped.
- We can now safely enlarge the ring buffer to hold 1 second of frames. It's now decoupled with the device hardware buffer and its size no longer depends on the latter size. The performance is improved. 
- Because of the previous change, unused osx_output_set_buffer_size's function parameters are removed.
- Properly manage the memory lifecycle of ring_buffer.


Wang-Yue:
fix https://github.com/MusicPlayerDaemon/MPD/issues/578

Wang-Yue:
Please note that

> During the cancel process, the ring buffer only keeps a small portions of front side buffers, and the rest of the old data are dropped.

Keeping the front samples is necessary to make sure the render function has something to consume when called again shortly after. It is likely that Play hasn't finished populating data after cancel when osx_render is called.  If we don't do this, the render function ends up reading no data. 

In many cases, osx_render fails to read sufficient samples will cause clocking issues (such as jitter). Many DACs with internal clocks (such USB DACs) may even believe the stream is stopped, and change from DoP mode to PCM mode.

Wang-Yue:
This PR is inherently flawed. Just realized that Cancel() will be called before Stop() or Pause() called.
In case of skipping, we want to preserve the front buffers to make it continuous. In Stop() or Pause() we don't. However there's no way to tell which condition it is working at.  MPD's design makes it super complex. I would rather remove Cancel() all together.

MaxKellermann:
If you believe MPD's output API design is too complex, suggest a better one. Your suggestion is to remove `Cancel()` completely doesn't sound so good - for example, it means MPD that will continue playing music for a while after the user has pressed "pause". That's a serious regression, and certainly not an improvement.

Wang-Yue:
On Mac the ring buffer size is 0.1s and such latency is hard to notice. 

I would rather give user an option to disable cancel for now.

Wang-Yue:
A PR to remove Cancel() is posted in https://github.com/MusicPlayerDaemon/MPD/pull/583. 
I'll leave both PR open. Please let me know which implementation you prefer. 

My personal opinion is, Cancel() is a bug, and shouldn't exist.

Wang-Yue:
Reopen this PR. 
Waiting for Max to implement the WriteSilence function discussed here: https://github.com/MusicPlayerDaemon/MPD/issues/578#issuecomment-502387458

MaxKellermann:
Writing silence if MPD's ring buffer runs empty is orthogonal to this PR. Don't wait.

Wang-Yue:
because the while loop is still causing hiccups in audio stream here: https://github.com/MusicPlayerDaemon/MPD/pull/582/files#diff-ce7ecec9ea9ca3df90d9c290cb3ef9d4R689

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -795,6 +795,7 @@ OSXOutput::Open(AudioFormat &audio_format)
 	PcmExport::Params params;
 	params.alsa_channel_order = true;
 	bool dop = dop_setting;
+	params.dsd_mode = PcmExport::DsdMode::NONE;
```
MaxKellermann: This is redundant. It's the default-initialized value.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -685,8 +685,24 @@ osx_render(void *vdata,
...
-	// Wait until data is available.
-	while (!od->ring_buffer->read_available()); 
+	// If data is not available, fill the buffer with silence,
+	if (!od->ring_buffer->read_available()) {
+#ifdef ENABLE_DSD
+		if (od->dop_enabled) {
+			const auto silence = od->pcm_export->GetSilence();
+			size_t count = in_number_frames * od->asbd.mBytesPerFrame / silence.size * silence.size;
+			for (size_t i = 0; i < count; ++i) {
+				memcpy((uint8_t *)buffer_list->mBuffers[0].mData + i * silence.size, silence.data, silence.size);
```
MaxKellermann: This assumes that previously, everything was already written at "block" boundary; but what if the previous `osx_render()` call wrote only the first frame of a DoP "block"?

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Done

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: See https://github.com/MusicPlayerDaemon/MPD/pull/582/files/3744a59da56b07fb5b4cccbf17d0787c3c1ef5e0..b6bcb97a527de660e6cb5cc68a70b077ded00182#diff-ce7ecec9ea9ca3df90d9c290cb3ef9d4R713

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -795,6 +795,7 @@ OSXOutput::Open(AudioFormat &audio_format)
 	PcmExport::Params params;
 	params.alsa_channel_order = true;
```
MaxKellermann: btw. unrelated to this commit, not your code.... why is ALSA channel order only used with DoP? Does OSX always need ALSA channel order or never? It doesn't appear to make sense to have it only with DoP.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: the osx_render function always write complete dop blocks, so we can safely write silence here.

MaxKellermann: Oh, good! ALSA can't do that, unfortunately, because a period can theoretically have an odd number of frames.

Wang-Yue:
Please be aware the PR is work in progress. I tested it with PCM audio and it works (albeit the delay is longer than when we busy wait, as we fill the entire CoreAudio buffer with silence). I can test DoP next week once I settle down and have access to a DSD capable DAC. If it works flawlessly I'll let you know and you can merge the PR.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: I'll mark it as resolved.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: No idea what alsa's channel order is. We allow users to configure their channel map via https://github.com/MusicPlayerDaemon/MPD/blob/1ffc2c6c5d0c3f6a09059de21762bb203d46d809/src/output/plugins/OSXOutputPlugin.cxx#L291

Wang-Yue:
PR is tested on both PCM and DSD. ready to be reviewed or merged.

MaxKellermann:
It's nearly impossible to review this PR, because it contains a merge commit and various fixup commits fixing mistakes in earlier new commits. Please fold those which belong together, and get rid of the merge commit.

Wang-Yue:
Closing this one. new PR is #595.

PR #583, opened 2019-06-14
Remove Cancel() in CoreAudio plugin

Wang-Yue:
see reasons in https://github.com/MusicPlayerDaemon/MPD/pull/582#issuecomment-502117387

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -101,7 +101,6 @@ private:
 	std::chrono::steady_clock::duration Delay() const noexcept override;
 	size_t Play(const void *chunk, size_t size) override;
 	bool Pause() override;
-	void Cancel() noexcept override;
 };
 
 static constexpr Domain osx_output_domain("osx_output");
@@ -917,17 +916,6 @@ bool OSXOutput::Pause() {
 		AudioOutputUnitStop(au);
 	}
 	return true;
-}
-
-void
-OSXOutput::Cancel() noexcept
-{
-	AudioOutputUnitStop(au);
-	ring_buffer->reset();
-#ifdef ENABLE_DSD
-	pcm_export->Reset();
-#endif
-	AudioOutputUnitStart(au);
 }
 
 int


MaxKellermann:
> However, energy wise they are different. Getting 1s DoP64 streams on my Mac would make it suddenly use 15% CPU cycle for a very short period of time. A lot of times this caused CPU frequency raise. On mobile chips, long but less intensive tasks saves more power/battery than short but intensive tasks.

This is wrong, it's a fallacy to believe that a higher CPU frequency consumes more power. The opposite is true!

At a lower CPU frequency, a batch of work takes longer to finish, which keeps the CPU busy for a longer duration. If you instead raise the CPU frequency, it finishes more quickly, and allows the CPU to stay in deeper sleep states for greater amounts of time, and at the bottom line, this actually saves power! 

So really, MPD's approach to do larger batches of work saves electric power. MPD's decoder uses a similar strategy: it wakes up the decoder thread only at a certain low water mark, to combine its work into larger batches, to be executed more efficiently.

> The cancel() behavior is just wrong in many aspects, especially on Linux, where a 1s alsa buffer is used:
> 
>     1. Most importantly, Every time you pause (which cancel the 1s buffer) and start the music from the same position, **1s of music data is lost**.  You said letting user hear music after pause is "a serious regression, and certainly not an improvement". But you don't care whether user will lose 1s of music each time they pause? To me the existence of Cancel() is a **bug**!

What you observe indeed is a bug, but while `Cancel()` plays a role in it, it isn't the bug per se. Without `Cancel()`, you'd still lose everything in the kernel/hardware buffer. `Cancel()` just amplifies this problem, but removing it isn't the solution.

>     2. users either need to wait for the 1s latency, or they're unable to seek without interrupting the stream. The reason is provided above --- right after dropping the data, OS Audio stack has nothing to read, and this caused a delay/hiccup  --- this is particularly bad in multi device environment when clocks needs to be locked/synced.

Interrupting the stream (to play something else instead) is the whole point of seeking.
The ALSA plugin generates silence when MPD's ring buffer runs empty - which is a good solution to avoid the xrun until the seeked-to data gets available. The OSX plugin doesn't do that, and I have always assumed that CoreAudio takes care for this, but I don't know.

I don't understand your point about clock locking/syncing.

>     3. The 1s latency also don't have outstanding performance benefit, compared to say 0.1s. It may consume more power/battery on mobile devices. discussed above.

Your discussion above was wrong. But other than being more efficient, it also allows MPD to have longer deadlines, which increases its reliability.

>     4. It cause latency problem as well --- UI displays faster than playing. So User may see their player's time stamp is 1s rather than 0s immediately after playing. Also when continue to next track, they see the UI switches track 1s faster than what they hear.

True, MPD should really be keeping track on the real position which is being played, and not the position which was submitted to the output. There are several layers of buffers, and each adds to latency. There's still too much latency at 0.1s ring buffer, so making the ring buffer smaller instead of doing proper time stamp management just makes the problem less visible, but doesn't solve it.

> My personal opinion is, Cancel() is a bug in API design. It sounds very clever and complex in a software point of view (I've never heard other players do that), but doesn't work at all in real world

I have. mocp and xmms2 both do the same. I wanted to check other music players (amarok, audacious), but found they didn't have any ALSA specific code.

None of your arguments are specific to the OSX plugin, but all your plugin does is remove `Cancel()` from that plugin. So even measured by your own arguments, this PR is bad. Given that I none of your arguments convince me that removing `Cancel()` is the right approach to the real problems you named, I hereby reject this PR.

As an overview, these are the real problems you named and my suggested real solution:

- workaround for pop noise bug in some DACs: add option to avoid the DAC reset (but still drop internal buffers - i.e. best-effort while doing the one thing that works around the DAC bug)
- time stamp: keep track of real playback position in output plugin
- data loss on pause: resubmit lost data after unpausing

Wang-Yue:
> This is wrong, it's a fallacy to believe that a higher CPU frequency consumes more power. The opposite is true!

Mac has an activity monitor which shows the energy impact of a program. What you said is not true. 

> The ALSA plugin generates silence when MPD's ring buffer runs empty - which is a good solution to avoid the xrun until the seeked-to data gets available. 

Then DOP’s marker pattern is broken. 

> As an overview, these are the real problems you named and my suggested real solution:

They make the code even more complex, while not solve the real problem

> workaround for pop noise bug in some DACs: add option to avoid the DAC reset (but still drop internal buffers - i.e. best-effort while doing the one thing that works around the DAC bug)

PR #582 just did exactly that (without an option). But sorry it just don’t work in the first place. As long as the buffer is dropped, the DAC reverts back to PCM because OS audio stack can’t keep up with DAC’s clock. So DAC assumes the stream is stopped when OS failed to provide data in time. This is proven on two of my above mentioned DACs which complains loudly when it happens. 

> data loss on pause: resubmit lost data after unpausing
 
But as you said it doesn’t solve the issue of losing hardware/kernel buffers because you don’t know how much is lost. 

> time stamp: keep track of real playback position in output plugin

You can’t do it solely in output plugin. In case of a track change output plugin has no idea what track is playing. 


> I wanted to check other music players

You don’t need to check their alsa implementation. Just check if they have cancel in plugin API. I don’t believe Cmus, MPV, VLC has that.

MaxKellermann:
> Just check if they have cancel in plugin API. I don’t believe Cmus, MPV, VLC has that.

Voila: https://github.com/cmus/cmus/blob/master/op/alsa.c#L227-L231
Called while seeking: https://github.com/cmus/cmus/blob/master/player.c#L1279-L1280
Oh and yes they have a CoreAudio implementation of that "cancel" method (which they call "drop"): https://github.com/cmus/cmus/blob/master/op/coreaudio.c#L714-L716

This very much looks just like MPD.

Wang-Yue:
> This very much looks just like MPD.

The cmus code was written by me. It’s not similar to MPD.

MaxKellermann:
And here's mpv: https://github.com/mpv-player/mpv/blob/master/audio/out/ao_alsa.c#L1084-L1096
Called while seeking: https://github.com/mpv-player/mpv/blob/master/player/playloop.c#L349
CoreAudio implementation: https://github.com/mpv-player/mpv/blob/master/audio/out/ao_coreaudio.c#L320-L323

This also very much looks like what MPD does.
Need more proof that you were wrong, or do you admit it?

Wang-Yue:
In cmus’s case , the code is written by me a few years ago. 
I remember it’s not related to pause or seek. 
It’s related to format change— during format change we need to dump the buffer, other wise if there is a bit change we would consume for instance the remaining 24 bits in a new 16 bits setting. And it just produce static garbage noise. 

In mpv’s case. I do not object seeking deletes data. I objects pausing do. 

Let me further clarify:
In seeking it’s fine to delete some data but should not let the stream stop. 
In pausing it’s fine to stop but should not delete data. 

Also the reset function you pointed out is the synonyms for stop. For some reason they chose a bad function name. What it essentially used is during pausing or stopping it stops the audio. I don’t believe it’s the same reset you see in the seeking code. 

So no, you have not pursuaded me.

Wang-Yue:
See https://github.com/mpv-player/mpv/blob/master/audio/out/ao_coreaudio.c#L421

It’s just a bad function name

MaxKellermann:
> In cmus’s case , the code is written by me a few years ago.
> I remember it’s not related to pause or seek.

Your memory is wrong, and I posted the link proving that it gets called while seeking.
Seriously, I took the time to check various other players and proved they do essentially the same thing as MPD - and you keep believing in your "alternative facts". I feel this was a bad investment of my time.
I havn't persuaded you, you havn't persuaded me, so let's keep it that way.

Wang-Yue:
> Your memory is wrong, and I posted the link proving that it gets called while seeking.

You are right. I woke up browsing the stuff on my phone so didn't notice that becuase of the tiny screen. 

Sorry.

PR #595, opened 2019-07-01

Wang-Yue:

- Have an atomic boolean monitoring the cancelling process. The Audio Unit is no longer stopped and restarted.
- During the cancel process, and the old data is dropped. in case the buffer is not ready has no data to read, the osx_render function just outputs silence.
- the ring buffer size is decoupled with the device hardware buffer size. Unused osx_output_set_buffer_size's function parameters are removed.
- Properly manage the memory lifecycle of ring_buffer.

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -40,8 +40,6 @@
 #include <CoreServices/CoreServices.h>
 #include <boost/lockfree/spsc_queue.hpp>
 
-#include <memory>
-
 static constexpr unsigned MPD_OSX_BUFFER_TIME_MS = 100;
 
 static StringBuffer<64>
@@ -76,6 +74,7 @@ struct OSXOutput final : AudioOutput {
 	 */
 	bool dop_setting;
 	bool dop_enabled;
+	std::atomic<bool> cancel;
 	Manual<PcmExport> pcm_export;
 #endif
 
@@ -129,7 +128,8 @@ osx_output_test_default_device(void)
 }
 
 OSXOutput::OSXOutput(const ConfigBlock &block)
-	:AudioOutput(FLAG_ENABLE_DISABLE|FLAG_PAUSE)
+	:AudioOutput(FLAG_ENABLE_DISABLE|FLAG_PAUSE),
+	 ring_buffer(nullptr)
 {
 	const char *device = block.GetBlockValue("device");
 
@@ -384,10 +384,10 @@ osx_output_set_device_format(AudioDeviceID dev_id, const AudioStreamBasicDescrip
 	};
 
 	UInt32 property_size;
-	OSStatus err = AudioObjectGetPropertyDataSize(dev_id, &aopa, 0, NULL, &property_size);	
-	if (err != noErr) {	
-		throw FormatRuntimeError("Cannot get number of streams: %d\n", err);	
-	}	
+	OSStatus err = AudioObjectGetPropertyDataSize(dev_id, &aopa, 0, NULL, &property_size);
+	if (err != noErr) {
+		throw FormatRuntimeError("Cannot get number of streams: %d\n", err);
+	}
 	
 	const size_t n_streams = property_size / sizeof(AudioStreamID);
 	static constexpr size_t MAX_STREAMS = 64;
@@ -395,30 +395,30 @@ osx_output_set_device_format(AudioDeviceID dev_id, const AudioStreamBasicDescrip
 		throw std::runtime_error("Too many streams");
 
 	AudioStreamID streams[MAX_STREAMS];
-	err = AudioObjectGetPropertyData(dev_id, &aopa, 0, NULL, &property_size, streams);	
-	if (err != noErr) {	
-		throw FormatRuntimeError("Cannot get streams: %d\n", err);	
-	}	
+	err = AudioObjectGetPropertyData(dev_id, &aopa, 0, NULL, &property_size, streams);
+	if (err != noErr) {
+		throw FormatRuntimeError("Cannot get streams: %d\n", err);
+	}
 	
 	bool format_found = false;
 	int output_stream;
 	AudioStreamBasicDescription output_format;
 
 	for (size_t i = 0; i < n_streams; i++) {
-		UInt32 direction;	
+		UInt32 direction;
 		AudioStreamID stream = streams[i];
-		aopa.mSelector = kAudioStreamPropertyDirection;	
-		property_size = sizeof(direction);	
-		err = AudioObjectGetPropertyData(stream,	
-						 &aopa,	
-						 0,	
-						 NULL,	
-						 &property_size,	
-						 &direction);	
+		aopa.mSelector = kAudioStreamPropertyDirection;
+		property_size = sizeof(direction);
+		err = AudioObjectGetPropertyData(stream,
+						 &aopa,
+						 0,
+						 NULL,
+						 &property_size,
+						 &direction);
 		if (err != noErr) {
-			throw FormatRuntimeError("Cannot get streams direction: %d\n", err);	
+			throw FormatRuntimeError("Cannot get streams direction: %d\n", err);
 		}
-		if (direction != 0) {	
+		if (direction != 0) {
 			continue;
 		}
 
@@ -462,23 +462,23 @@ osx_output_set_device_format(AudioDeviceID dev_id, const AudioStreamBasicDescrip
 	}
 
 	if (format_found) {
-		aopa.mSelector = kAudioStreamPropertyPhysicalFormat;	
+		aopa.mSelector = kAudioStreamPropertyPhysicalFormat;
 		err = AudioObjectSetPropertyData(output_stream,
-						 &aopa,	
-						 0,	
-						 NULL,	
-						 sizeof(output_format),	
-						 &output_format);	
-		if (err != noErr) {	
-			throw FormatRuntimeError("Failed to change the stream format: %d\n", err);	
-		}	
-	}	
+						 &aopa,
+						 0,
+						 NULL,
+						 sizeof(output_format),
+						 &output_format);
+		if (err != noErr) {
+			throw FormatRuntimeError("Failed to change the stream format: %d\n", err);
+		}
+	}
 
 	return output_format.mSampleRate;
 }
 
 static OSStatus
-osx_output_set_buffer_size(AudioUnit au, AudioStreamBasicDescription desc, UInt32 *frame_size)
+osx_output_set_buffer_size(AudioUnit au)
 {
 	AudioValueRange value_range = {0, 0};
 	UInt32 property_size = sizeof(AudioValueRange);
@@ -499,7 +499,7 @@ osx_output_set_buffer_size(AudioUnit au, AudioStreamBasicDescription desc, UInt3
 				   &buffer_frame_size,
 				   sizeof(buffer_frame_size));
 	if (err != noErr)
-                FormatWarning(osx_output_domain,
+		FormatWarning(osx_output_domain,
 			      "Failed to set maximum buffer size: %d",
 			      err);
 
@@ -511,18 +511,10 @@ osx_output_set_buffer_size(AudioUnit au, AudioStreamBasicDescription desc, UInt3
 				   &buffer_frame_size,
 				   &property_size);
 	if (err != noErr) {
-                FormatWarning(osx_output_domain,
+		FormatWarning(osx_output_domain,
 			      "Cannot get the buffer frame size: %d",
 			      err);
 		return err;
-	}
-
-	buffer_frame_size *= desc.mBytesPerFrame;
-
-	// We set the frame size to a power of two integer that
-	// is larger than buffer_frame_size.
-	while (*frame_size < buffer_frame_size + 1) {
-		*frame_size <<= 1;
 	}
 
 	return noErr;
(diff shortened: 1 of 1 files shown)


src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -76,6 +74,7 @@ struct OSXOutput final : AudioOutput {
 	 */
 	bool dop_setting;
 	bool dop_enabled;
+	std::atomic<bool> cancel;
```
MaxKellermann: This requires `<atomic>`; also, `std::atomic_bool` is a shortcut which might be useful here.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -129,7 +128,8 @@ osx_output_test_default_device(void)
 }
 
 OSXOutput::OSXOutput(const ConfigBlock &block)
-	:AudioOutput(FLAG_ENABLE_DISABLE|FLAG_PAUSE)
+	:AudioOutput(FLAG_ENABLE_DISABLE|FLAG_PAUSE),
+	 ring_buffer(nullptr)
```
MaxKellermann: This doesn't need to be initialized here. If you do anyway with an arbitrary value like `nullptr`, you lose the chance to detect uninitialized variables with tools like valgrind.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -384,41 +384,41 @@ osx_output_set_device_format(AudioDeviceID dev_id, const AudioStreamBasicDescrip
 	};
 
 	UInt32 property_size;
-	OSStatus err = AudioObjectGetPropertyDataSize(dev_id, &aopa, 0, NULL, &property_size);	
-	if (err != noErr) {	
-		throw FormatRuntimeError("Cannot get number of streams: %d\n", err);	
-	}	
+	OSStatus err = AudioObjectGetPropertyDataSize(dev_id, &aopa, 0, NULL, &property_size);
```
MaxKellermann: Whitespace-only changes don't belong here. They just obfuscate the real changes and make review hard.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -384,41 +384,41 @@ osx_output_set_device_format(AudioDeviceID dev_id, const AudioStreamBasicDescrip
...
+		throw FormatRuntimeError("Cannot get number of streams: %d\n", err);
+	}
 	
 	const size_t n_streams = property_size / sizeof(AudioStreamID);
 	static constexpr size_t MAX_STREAMS = 64;
 	if (n_streams > MAX_STREAMS)
 		throw std::runtime_error("Too many streams");
 
 	AudioStreamID streams[MAX_STREAMS];
-	err = AudioObjectGetPropertyData(dev_id, &aopa, 0, NULL, &property_size, streams);
```
MaxKellermann: More whitespace changes.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -384,41 +384,41 @@ osx_output_set_device_format(AudioDeviceID dev_id, const AudioStreamBasicDescrip
...
+	if (err != noErr) {
+		throw FormatRuntimeError("Cannot get streams: %d\n", err);
+	}
 	
 	bool format_found = false;
 	int output_stream;
 	AudioStreamBasicDescription output_format;
 
 	for (size_t i = 0; i < n_streams; i++) {
-		UInt32 direction;
```
MaxKellermann: And yet more whitespace changes.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -688,7 +680,44 @@ osx_render(void *vdata,
 {
 	OSXOutput *od = (OSXOutput *) vdata;
 
-	int count = in_number_frames * od->asbd.mBytesPerFrame;
+	if (od->cancel) {
+		od->ring_buffer->reset();
```
MaxKellermann: This method is not thread-safe. Concurrent `Play()` calls may wreak havoc.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -688,7 +680,44 @@ osx_render(void *vdata,
...
+		od->cancel = false;
+	}
+
+	// If data is not available, fill the buffer with silence,
```
MaxKellermann: Can we move the silence generation code in a separate commit? Mixing two orthogonal changes in one commit may make bisects harder than necessary, and obscures the patch.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -688,7 +680,44 @@ osx_render(void *vdata,
...
+			size_t count = in_number_frames * od->asbd.mBytesPerFrame / silence.size;
+			for (size_t i = 0; i < count; ++i) {
+				memcpy((uint8_t *)buffer_list->mBuffers[0].mData + i * silence.size, silence.data, silence.size);
+			}
+			buffer_list->mBuffers[0].mDataByteSize = count * silence.size;
+			return noErr;
+		}
+#endif
+		size_t count = in_number_frames * od->asbd.mBytesPerFrame;
+		bzero(buffer_list->mBuffers[0].mData, count);
```
MaxKellermann: I'd prefer to use `PcmSilence()` here. Even though `bzero()` happens to be correct here, doing this in a central code location is better. Right now, only DSD uses a different silence pattern; now imagine one day you add native DSD support here; using `PcmSilence()` means you don't need to refactor this one.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -688,7 +680,44 @@ osx_render(void *vdata,
...
+	size_t copy_frames =
+		od->ring_buffer->read_available() / od->asbd.mBytesPerFrame;
+	if (copy_frames > in_number_frames) {
+		copy_frames = in_number_frames;
+	}
+#ifdef ENABLE_DSD
+	// In DoP, we always cut off at even frames to preserve the
+	// DoP marker bits pattern.
+	if (od->dop_enabled) {
+		copy_frames = copy_frames / 2 * 2;
```
MaxKellermann: I think instead of this special case, the code should use `PcmExport::GetOutputBlockSize()`, which is a generic way to calculate this.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -752,7 +781,10 @@ OSXOutput::Close() noexcept
 {
 	AudioOutputUnitStop(au);
 	AudioUnitUninitialize(au);
-	delete ring_buffer;
+	if (ring_buffer) {
```
MaxKellermann: Superfluous `nullptr` check. Why is it even here? Nobody ever assigns `nullptr` to this variable.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -752,7 +781,10 @@ OSXOutput::Close() noexcept
...
+		delete ring_buffer;
+		ring_buffer = nullptr;
```
MaxKellermann: Assigning `nullptr` here is not necessary. This just hides possible use-after-free bugs.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -841,22 +872,19 @@ OSXOutput::Open(AudioFormat &audio_format)
...
-						   MPD_OSX_BUFFER_TIME_MS * audio_format.GetFrameSize() * audio_format.sample_rate / 1000);
+	size_t ring_buffer_size = MPD_OSX_BUFFER_TIME_MS * audio_format.GetFrameSize() * audio_format.sample_rate / 1000;
 
 #ifdef ENABLE_DSD
-        if (dop_enabled) {
+	if (dop_enabled) {
 		pcm_export->Open(audio_format.format, audio_format.channels, params);
-		ring_buffer_size = std::max<size_t>(buffer_frame_size,
-						   MPD_OSX_BUFFER_TIME_MS * pcm_export->GetOutputFrameSize() * asbd.mSampleRate / 1000);
+		ring_buffer_size = MPD_OSX_BUFFER_TIME_MS * pcm_export->GetOutputFrameSize() * asbd.mSampleRate / 1000;
```
MaxKellermann: Food (for a new commit, not to be mixed into this one): this formula is redundant; use `AudioFormat::TimeToSize()`.
Also I'd prefer to use `std::chrono` instead of those `_MS` suffixes and hard-coding `* 1000`.

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -883,8 +913,11 @@ OSXOutput::Play(const void *chunk, size_t size)
 			throw std::runtime_error("Unable to restart audio output after pause");
 		}
 	}
+	if (cancel) {
+		return 0;
```
MaxKellermann: This silently loses all audio data arriving between the `Cancel()` call and whenever the render thread happens to call `osx_render()` the next time. `Cancel()` should only discard data which arrived until that call; newly added data must not be lost.

Wang-Yue: I assume when Play returns 0, it means at this call, no data is consumed, and next time it calls, it should output buffer at the same position, No?

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: that's why it's wrapped in atomic variable cancel. If cancel is true, Play() will never write to the ring buffer.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Could you do a white space only commit to this file? The IDE tries to fix the formatting each time I open the file.

src/output/plugins/OSXOutputPlugin.cxx
MaxKellermann: Actually, I was wrong, but your explanation doesn't make sense either; if `Play()==0`, when will the next call be?
Data isn't lost with your change; it's simply that `Play()` must never return 0. It must either consume at least one frame (and wait if necessary) or throw an exception.
But I admit that this requirement isn't documented properly; I'll add that.

MaxKellermann: > But I admit that this requirement isn't documented properly

Oh, it is documented after all; API documentation says:

> The method blocks until at least one audio frame is consumed.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Done in https://github.com/MusicPlayerDaemon/MPD/pull/598

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Done in https://github.com/MusicPlayerDaemon/MPD/pull/598

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Mark as resolved.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: I'll mark this as resolved.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: https://github.com/MusicPlayerDaemon/MPD/pull/598 does not contain this change. Will do in another PR.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Will do in future PR, together with the above comment to add silence generation back.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Done

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Done in #598

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Will do in a new PR.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: keep it busy waiting in #598. We may change it to mutex/cond_var in the future.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: Umm... then how can I get number 2 here? GetOutputBlockSize()/GetFrameSize()?

Wang-Yue:
I'll split it to one PR at a time. will keep this PR for reference.

src/output/plugins/OSXOutputPlugin.cxx
MaxKellermann: That would be obscure and fragile. Better always use `PcmExport::GetOutputBlockSize()` and never use the frame size.

Wang-Yue: But core audio provide the number of frames in the method signature.

MaxKellermann:
See https://github.com/MusicPlayerDaemon/MPD/pull/598

PR #598, opened 2019-07-04

Wang-Yue:

-  Have an atomic boolean monitoring the cancelling process. The Audio Unit is no longer stopped and restarted.
- the ring buffer size is decoupled with the device hardware buffer size. Unused osx_output_set_buffer_size's function parameters are removed.

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -40,7 +40,7 @@
 #include <CoreServices/CoreServices.h>
 #include <boost/lockfree/spsc_queue.hpp>
 
-#include <memory>
+#include <atomic>
 
 static constexpr unsigned MPD_OSX_BUFFER_TIME_MS = 100;
 
@@ -77,6 +77,7 @@ struct OSXOutput final : AudioOutput {
 	 */
 	bool dop_setting;
 	bool dop_enabled;
+	std::atomic_bool cancel;
 	Manual<PcmExport> pcm_export;
 #endif
 
@@ -493,8 +494,7 @@ osx_output_set_device_format(AudioDeviceID dev_id,
 }
 
 static OSStatus
-osx_output_set_buffer_size(AudioUnit au, AudioStreamBasicDescription desc,
-			   UInt32 *frame_size)
+osx_output_set_buffer_size(AudioUnit au)
 {
 	AudioValueRange value_range = {0, 0};
 	UInt32 property_size = sizeof(AudioValueRange);
@@ -531,14 +531,6 @@ osx_output_set_buffer_size(AudioUnit au, AudioStreamBasicDescription desc,
 			      "Cannot get the buffer frame size: %d",
 			      err);
 		return err;
-	}
-
-	buffer_frame_size *= desc.mBytesPerFrame;
-
-	// We set the frame size to a power of two integer that
-	// is larger than buffer_frame_size.
-	while (*frame_size < buffer_frame_size + 1) {
-		*frame_size <<= 1;
 	}
 
 	return noErr;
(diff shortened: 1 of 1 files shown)


src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -717,7 +709,28 @@ osx_render(void *vdata,
...
 	OSXOutput *od = (OSXOutput *) vdata;
 
-	int count = in_number_frames * od->asbd.mBytesPerFrame;
+	if (od->cancel) {
+		od->ring_buffer->reset();
+		od->cancel = false;
+	}
+
+	// Wait until data is available.
+	while (!od->ring_buffer->read_available());
```
MaxKellermann: This is a busy loop, churning CPU and thus slowing down the decoder/player threads, leading to further delay. Is this really a good idea?

src/output/plugins/OSXOutputPlugin.cxx
```diff
@@ -914,6 +925,7 @@ OSXOutput::Play(const void *chunk, size_t size)
 			throw std::runtime_error("Unable to restart audio output after pause");
 		}
 	}
+	while (cancel);
```
MaxKellermann: Another busy loop.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: The code will be removed in a follow up PR, as you want to split the silence generation code from the bigger PR. When silence generation is added this will become an if rather than while. 

Also this only busy waits the moment Cancel happens and running for a couple of milliseconds at most. So the impact is very minimal.

src/output/plugins/OSXOutputPlugin.cxx
Wang-Yue: This one would be changed to a condition variable wait. But we will need to do a lot of changes before that happens (Such as silence generation, and other changes you proposed). Better not to do it in this PR

Again it only happens during Cancel, so the impact is very minimal.

src/output/plugins/OSXOutputPlugin.cxx
MaxKellermann: I disagree, it's not "minimal". If `osx_render()` runs with real-time priority, this will completely block one CPU core and will not give anybody else a share.
I don't even know why this loop is part of the commit. It has nothing to do with cancellation. The loop didn't exist previously, and the "real" changes in this commit don't add a need for it.
Just like silence generation didn't belong in this commit - but if you feel the silence generator is necessary to avoid this busy loop, then let's merge the silence generator commit first!

Wang-Yue: i lost interest in keeping up with the ever changing requests.  
Sorry I don’t have that much time to change the code back and forth. 

If someone else has interest in merging the changes in I will be happy to see that happen 

Feel free to close these pull requests

MaxKellermann:
> These commits solves one problem and should be integrated.

This PR introduces a rather critical bug, and I won't merge PRs which have known critical bugs.

> You asked to split and they will leave audible/code issues here or there. It is a waste of time polish the code here and then remove the polished code altogether in the next commit, just to make you happy.

I don't understand this part. I didn't ask you to write code which will be removed later. I suggested merging the silence generator first, so we can avoid having this busy loop at all. My suggestion means omitting bad code completely, instead of adding it and then removing it.

> If someone else (such as you) has interest in merging the changes from the other big PR following your standards I will be happy to see that happen. Personally I don’t mind to have a personal build which removes the cancel code altogether.

I certainly won't work on the OSX plugin, because I (proudly) don't have any Apple hardware. If you give up, this means the OSX plugin is back to status "unmaintained". That's ok for me, MPD's macOS support doesn't mean anything to me, I don't care. But I care for not merging known-bad code, not even in the OSX plugin. In the past, I've often merged known-bad code, hoping for the best, but it has only caused sorrow (as in: lots of bug reports only I care for, lots of work for me, bad reputation for the project). My lesson is that I only merge code submissions which meet certain quality requirements, even if that means losing contributors who don't have enough patience to finish their work.

> i lost interest in keeping up with the ever changing requests. [...] Feel free to close these pull requests

As you wish.

Wang-Yue:
Merging silence generation first doesn’t make sense. If we are stopping the audio, what’s the point (or where) to do silence generation ? If you are not doing the silence generation but leaves the code there but in a dormant state, it’s not useful for future bisect when people wants to find the root problem of a future big. 

Removing the whole hides the problem. But this makes OS X render busy waiting as well, as each time it returns zero data— the audio stack will call it repeatedly. It’s no difference compare to the busy loop inside. 

Again it’s your suggestion doesn’t make sense at all. I can only do a commit have both in the PR otherwise it just doesn’t make sense. Yes you don’t care about OS X and that’s fine. But at least you should not give unreasonable requirements to merging a PR that fix important problem.

Wang-Yue:
I can work on a PR which does not stop and with silence generation. But that has to be done in one commit.

We can leave other stuffs later, Such as time calculation using chrono library. 

But I insist to do the two mentioned above in one PR. If you don’t agree on this that’s the end of this discussion.

Wang-Yue:
ping

PR #605, opened 2019-07-15
Add option to not connect jack ports automatically

derekdreery:
Some people (including me) use services like LADISH and non-session-manager to manage their jack connections. For this reason, it's considered good practice to provide an option in clients to disable auto-connecting ports.

This patch adds a config option to do precisely that.

### DIFF
--- a/doc/plugins.rst
+++ b/doc/plugins.rst
@@ -864,9 +864,10 @@ The jack plugin connects to a `JACK server <http://jackaudio.org/>`_.
      - The names of the JACK source ports to be created. By default, the ports "left" and "right" are created. To use more ports, you have to tweak this option.
    * - **destination_ports A,B**
      - The names of the JACK destination ports to connect to.
-   * - **no_auto_connect yes|no**
-     - If set to yes, then mpd won't make any automatic connections between mpd ports and other
-       jack ports. This means that *destination_ports* will have no effect.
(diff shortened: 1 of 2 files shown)


MaxKellermann:
I don't like that this new option has a negated name, which can result in double negation, making this option easy to misunderstand. Imagine you want to allow MPD to automatically connect ports, and you'd have to write `no_auto_connect "no"` .. huh?
And I don't like that this silently ignores the `destination_ports` option. Yes, that is documented, but it will lead to support cases which are avoidable. Users rarely ever read documentation, and this kind of misconfiguration should result in an error message; or better, redefine the semantics of the new option to eliminate the possibility of a misconfiguration.
The new option doesn't really do what it pretends to do: you wrote "to not connect jack ports automatically", but in addition to not doing that automatically, it doesn't do it at all. This difference is confusing, even though it's documented in the fine print (nobody reads the fine print, see above).
My suggestion: new option `auto_destination_ports` defaulting to `yes`. If set to `no`, then only ports which are explicitly configured will be connected.

derekdreery:
Sounds sensible - I will modify the PR to match what you say.l

derekdreery:
The build failure is spurious I think.

MaxKellermann:
Folded both commits into one: 4f6a713b3212a176ef133b08fbd708fe8729e0cc

PR #611, opened 2019-08-02
Fix JACK plugin outputting mono input only to left channel

Diomendius:
The JACK output plugin would not correctly upmix mono input files when exactly 2 output ports were configured. This fixes that.


MaxKellermann:
Cherry-picked to v0.21.x: b81138bda1c6cb1b3d1fd8ecbf4174ac939b5f44

PR #614, opened 2019-08-04
decoder/sidplay: Fix windows-1252 to utf-8 string conversion

frno7:
High Voltage SID Collection (HVSC) metadata fields are encoded in
windows-1252, as described in DOCUMENTS/SID_file_format.txt:

https://www.hvsc.c64.org/download/C64Music/DOCUMENTS/SID_file_format.txt

C++ exception handling may need to be adjusted before submitting this.

### DIFF
--- a/src/decoder/plugins/SidplayDecoderPlugin.cxx
+++ b/src/decoder/plugins/SidplayDecoderPlugin.cxx
@@ -25,6 +25,7 @@
 #include "song/DetachedSong.hxx"
 #include "fs/Path.hxx"
 #include "fs/AllocatedPath.hxx"
+#include "lib/icu/Converter.hxx"
 #ifdef HAVE_SIDPLAYFP
 #include "fs/io/FileReader.hxx"
 #include "util/RuntimeError.hxx"
@@ -33,6 +34,8 @@
 #include "util/StringView.hxx"
 #include "util/Domain.hxx"
 #include "util/ByteOrder.hxx"
+#include "util/AllocatedString.hxx"
+#include "util/CharUtil.hxx"
 #include "Log.hxx"
 
 #ifdef HAVE_SIDPLAYFP
@@ -437,19 +440,46 @@ sidplay_file_decode(DecoderClient &client, Path path_fs)
 	} while (cmd != DecoderCommand::STOP);
 }
 
+static AllocatedString<char>
+Windows1252ToUTF8(const char *s)
+{
+#ifdef HAVE_ICU_CONVERTER
+	try {
+		std::unique_ptr<IcuConverter>
+			converter(IcuConverter::Create("windows-1252"));
+
+		return converter->ToUTF8(s);
+	} catch (...) { }
+#endif
+
+	/*
+	 * Fallback to not transcoding windows-1252 to utf-8, that may result
+	 * in invalid utf-8 unless nonprintable characters are replaced.
+	 */
+	auto t = AllocatedString<char>::Duplicate(s);
+
+	for (size_t i = 0; t[i] != AllocatedString<char>::SENTINEL; i++)
+		if (!IsPrintableASCII(t[i]))
+			t[i] = '?';
+
+	return t;
+}
+
 gcc_pure
-static const char *
-GetInfoString(const SidTuneInfo &info, unsigned i) noexcept
+static AllocatedString<char>
+GetInfoString(const SidTuneInfo &info, unsigned i)
 {
 #ifdef HAVE_SIDPLAYFP
-	return info.numberOfInfoStrings() > i
+	const char *s = info.numberOfInfoStrings() > i
 		? info.infoString(i)
-		: nullptr;
+		: "";
 #else
-	return info.numberOfInfoStrings > i
+	const char *s = info.numberOfInfoStrings > i
 		? info.infoString[i]
-		: nullptr;
+		: "";
 #endif
+
+	return Windows1252ToUTF8(s);
 }
 
 static void
@@ -457,31 +487,29 @@ ScanSidTuneInfo(const SidTuneInfo &info, unsigned track, unsigned n_tracks,
 		TagHandler &handler) noexcept
 {
 	/* title */
-	const char *title = GetInfoString(info, 0);
-	if (title == nullptr)
-		title = "";
+	AllocatedString<char> title = GetInfoString(info, 0);
 
 	if (n_tracks > 1) {
 		const auto tag_title =
 			StringFormat<1024>("%s (%u/%u)",
-					   title, track, n_tracks);
+					   title.c_str(), track, n_tracks);
 		handler.OnTag(TAG_TITLE, tag_title.c_str());
 	} else
-		handler.OnTag(TAG_TITLE, title);
+		handler.OnTag(TAG_TITLE, title.c_str());
 
 	/* artist */
-	const char *artist = GetInfoString(info, 1);
-	if (artist != nullptr)
-		handler.OnTag(TAG_ARTIST, artist);
+	AllocatedString<char> artist = GetInfoString(info, 1);
+	if (!artist.empty())
+		handler.OnTag(TAG_ARTIST, artist.c_str());
 
 	/* genre */
 	if (!default_genre.empty())
 		handler.OnTag(TAG_GENRE, default_genre.c_str());
 
 	/* date */
-	const char *date = GetInfoString(info, 2);
-	if (date != nullptr)
-		handler.OnTag(TAG_DATE, date);
+	AllocatedString<char> date = GetInfoString(info, 2);
+	if (!date.empty())
+		handler.OnTag(TAG_DATE, date.c_str());
 
 	/* track */
 	handler.OnTag(TAG_TRACK, StringFormat<16>("%u", track).c_str());


MaxKellermann:
Your PR breaks the build if neither libicu nor iconv are available (`HAVE_ICU_CONVERTER` is not defined).

> C++ exception handling may need to be adjusted before submitting this.

Of course, do it! Can't merge this as it is now.

frno7:
> Your PR breaks the build if neither libicu nor iconv are available (HAVE_ICU_CONVERTER is not defined).

Fixed!

> Of course, do it! Can't merge this as it is now.

Done!

Please consider the updated patch for inclusion.

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -437,51 +439,83 @@ sidplay_file_decode(DecoderClient &client, Path path_fs)
...
+
+		return converter->ToUTF8(s);
+	} catch (...) { }
+#endif
+
+	/*
+	 * Fallback to not transcoding windows-1252 to utf-8, that may result
+	 * in invalid utf-8 unless nonprintable characters are replaced.
+	 */
+	AllocatedString<char> t = AllocatedString<char>::Duplicate(s);
```
MaxKellermann: For readability use `auto`

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -437,51 +439,83 @@ sidplay_file_decode(DecoderClient &client, Path path_fs)
...
+
+	for (size_t i = 0; t[i] != AllocatedString<char>::SENTINEL; i++)
+		if (!std::isprint(t[i]))
```
MaxKellermann: `isprint()` is a bad choice, because it is locale-dependent and may return true on non-ASCII characters. It does many things, but prevent invalid UTF-8 sequences is not what it does.
What you could use is `IsPrintableASCII()` from `util/CharUtil.hxx` which guarantees that it's printable *and* valid UTF-8 (because 7-bit ASCII is a subset of UTF-8).

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -437,51 +439,83 @@ sidplay_file_decode(DecoderClient &client, Path path_fs)
...
-static const char *
-GetInfoString(const SidTuneInfo &info, unsigned i) noexcept
+static AllocatedString<char>
+GetInfoString(const SidTuneInfo &info, unsigned i)
 {
 #ifdef HAVE_SIDPLAYFP
-	return info.numberOfInfoStrings() > i
+	const char *s = info.numberOfInfoStrings() > i
 		? info.infoString(i)
-		: nullptr;
```
MaxKellermann: Unlike `std::string`, `AllocatedString<>` is nullable. You don't need to use an empty string as magic value for "no value", but keep `nullptr` here.

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -437,51 +439,83 @@ sidplay_file_decode(DecoderClient &client, Path path_fs)
...
+
+		if (n_tracks > 1) {
+			const auto tag_title =
+				StringFormat<1024>("%s (%u/%u)",
+						   title.c_str(),
+						   track, n_tracks);
+			handler.OnTag(TAG_TITLE, tag_title.c_str());
+		} else
+			handler.OnTag(TAG_TITLE, title.c_str());
+	} catch (...) { }
```
MaxKellermann: Exceptions have already been caught inside `Windows1252ToUTF8()`, so what is this supposed to catch?

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -437,51 +439,83 @@ sidplay_file_decode(DecoderClient &client, Path path_fs)
...
+
+	try {
+		/* artist */
+		AllocatedString<char> artist = GetInfoString(info, 1);
+		if (!artist.empty())
+			handler.OnTag(TAG_ARTIST, artist.c_str());
+	} catch (...) { }
```
MaxKellermann: Same here

src/decoder/plugins/SidplayDecoderPlugin.cxx
```diff
@@ -437,51 +439,83 @@ sidplay_file_decode(DecoderClient &client, Path path_fs)
...
-	/* date */
-	const char *date = GetInfoString(info, 2);
-	if (date != nullptr)
-		handler.OnTag(TAG_DATE, date);
+	try {
+		/* date */
+		AllocatedString<char> date = GetInfoString(info, 2);
+		if (!date.empty())
+			handler.OnTag(TAG_DATE, date.c_str());
+	} catch (...) { }
```
MaxKellermann: .. and here

MaxKellermann:
Your other PRs depend on this one (and they contain this commit). For me, it's okay to have semi-related commits in one PR, as long as the commits have proper separation (not mix two unrelated things in one commit). So if this makes your work easier, you can merge those PRs into one.

src/decoder/plugins/SidplayDecoderPlugin.cxx
frno7: Hmm... I can certainly change to `auto`. [ As a personal preference I consider `auto` _less_ readable though, because then one has to chase around to find what the type actually _is_. C++ isn't JavaScript or Python, after all. :smile: ]

src/decoder/plugins/SidplayDecoderPlugin.cxx
frno7: Hmm... I forgot about the locale stuff. `IsPrintableASCII` looks no good since `0x7f` is DEL, which isn't printable, so I think that must be corrected first:

```
constexpr bool
IsPrintableASCII(char ch) noexcept
{
	return (signed char)ch >= 0x20;
}
```

src/decoder/plugins/SidplayDecoderPlugin.cxx
frno7: Yes, but the previous code had a special case to convert the `nullptr` to the empty string, which is why it was simpler to skip the `nullptr` altogether. I consider the empty string less magic (and much safer and less likely to crash anything) than `nullptr`.

And if the retrieved value happens to be the empty string (which can happen), then it does not seem like it should be used anyway, so why not handle the empty string as the common case without introducing the `nullptr` at all?

src/decoder/plugins/SidplayDecoderPlugin.cxx
frno7: It seems AllocatedString may fail to allocate its string, and hence throw a memory related exception.

src/decoder/plugins/SidplayDecoderPlugin.cxx
MaxKellermann: True, sometimes it's hard to tell; but sometimes, it doesn't matter as long as you get the idea what it can be used for (e.g. iterators), and in this case, it's too obvious (because the full type name is spelled again on the same line).

src/decoder/plugins/SidplayDecoderPlugin.cxx
MaxKellermann: You're right, this function should probably be corrected or renamed. Let's not care for that right now.

frno7:
Yes, the other PRs have code dependencies, but are otherwise conceptually unrelated to this one, which is why I put them into separate PRs. Thanks for allowing me simplify that, although I think we can proceed anyway now.

src/decoder/plugins/SidplayDecoderPlugin.cxx
MaxKellermann: As a general rule, I don't care for memory allocation failures; even if those would really throw an exception (they don't, at least not on Linux), it would be OK for me to abort MPD (via `std::terminate`). Once you're on memory shortage, there's little you can do to recover from the situation from inside the process; the best you can do to let the computer breathe again is just exit the process as quickly as possible. And MPD isn't a particularly important process.

src/decoder/plugins/SidplayDecoderPlugin.cxx
MaxKellermann: I prefer returning `nullptr` if there is no value, but I'd accept your approach.
(Sure, empty values will be discarded later, but I don't want to allocate memory for an empty string when I already know I won't be using it.)

frno7: `AllocatedString<>` could be designed to special case the empty string to not allocate anything at all. Then user code could be maximally fast and safe without the `nullptr`. :smile:

src/decoder/plugins/SidplayDecoderPlugin.cxx
frno7: Ah, I was unaware that memory exceptions were exempted. I can remove those catches then.

MaxKellermann: That was a good chance to improve the developer manual... https://github.com/MusicPlayerDaemon/MPD/commit/349a2ea7eb4df3d24bebe9061d47da37bfc078f4

src/decoder/plugins/SidplayDecoderPlugin.cxx
MaxKellermann: It could, but that would make its implementation very complicated. Right now, it's a plain pointer, and has nearly no overhead.

MaxKellermann:
Cherry-picked to v0.21.x: 7723c481db66f089f5ccde941a0a952cb1b16e2e

PR #618, opened 2019-08-05
decoder/sidplay: Fix song length initialisation during container scan

frno7:
The song length was previously undetermined.

### DIFF
--- a/src/decoder/plugins/SidplayDecoderPlugin.cxx
+++ b/src/decoder/plugins/SidplayDecoderPlugin.cxx
@@ -549,6 +549,10 @@ sidplay_container_scan(Path path_fs)
 		AddTagHandler h(tag_builder);
 		ScanSidTuneInfo(info, i, n_tracks, h);
 
+		const SignedSongTime duration = get_song_length(tune);
+		if (!duration.IsNegative())
+			h.OnDuration(SongTime(duration));
+
 		char track_name[32];
 		/* Construct container/tune path names, eg.
 		   Delta.sid/tune_001.sid */


MaxKellermann:
This uses the song length of the whole file, not the one track it's scanning. Are you sure this is correct?

frno7:
> This uses the song length of the whole file, not the one track it's scanning.

No, because SID tunes are mapped as distinct tracks, as explained below.

> Are you sure this is correct?

Yes, I'm quite sure. As explained in the comment in the piece of code just below the patched part, SID tracks are constructed as for example `Delta.sid/tune_001.sid` so the actual file is `Delta.sid`, and the track is `tune_001.sid` as a (kind of) virtual directory within `Delta.sid` file. The context of the patch is this loop over all tracks in a file:

```
	for (unsigned i = 1; i <= n_tracks; ++i) {
		tune.selectSong(i);

		AddTagHandler h(tag_builder);
		ScanSidTuneInfo(info, i, n_tracks, h);
```

So `i` is the track number within `tune`, and the track is selected with `tune.selectSong(i)`. Then comes the patched part:

```
		const SignedSongTime duration = get_song_length(tune);
		if (!duration.IsNegative())
			h.OnDuration(SongTime(duration));
```

And so the track name (e.g. `Delta.sid/tune_001.sid`) is added to the list of tracks:

```
		char track_name[32];
		/* Construct container/tune path names, eg.
		   Delta.sid/tune_001.sid */
		sprintf(track_name, SUBTUNE_PREFIX "%03u.sid", i);
		tail = list.emplace_after(tail, track_name,
					  tag_builder.Commit());
	}
```

The HVSC maintains a [STIL file](https://www.hvsc.de/download/C64Music/DOCUMENTS/STIL.txt) where individual track names (and other interesting stuff such as comments and composers per track) can be obtained. I have forthcoming MPD patches to make use of the STIL, but it seems prudent to merge fixes for obvious problems first. :smile:

MaxKellermann:
Understood - I didn't know that the current track is considered by `SidDatabase::length()` - my `SongLengths.txt` looked like it didn't have per-track data, but indeed it does have that.

MaxKellermann:
Cherry-picked to the stable branch (v0.21.x) 0ed10542cc0ef7c590e3b8b9717ae26309dd494a

PR #623, opened 2019-08-09
 decoder/sidplay: Fix empty album names by using the SID name field 

frno7:
SID files are generally collections of tunes, so a SID name field makes
sense as an MPD album. The SID tune information list (STIL) has name
and title fields for individual tunes, when such are known, but MPD is
currently not using the STIL.


MaxKellermann:
9654a33218ce45aa03b914f8806e92dd2c554a84

PR #653, opened 2019-09-23
command/file: add command "readlyrics"

NBonaparte:
Implements unsynced lyrics (ID3 "USLT") as a tag.

Not sure if it should be disabled by default (like `TAG_COMMENT`); any feedback/opinions would be appreciated.


MaxKellermann:
How does your change deal with multi-line tag values? Lyrics tags are usually multi-line, aren't they?

prat0088:
What's needed before this can be merged?

NBonaparte:
@prat0088 This version doesn't really work with tags with newlines, since IIRC it strips them out or truncates, can't remember which.

I reimplemented lyrics support as a `readlyrics` command (similar to `readpicture`, etc.) which should properly handle newlines nearly two months ago (AFAIR) on c4f7370ed9b361e0360ee5a977a46efcda5b51cc, but forgot to push to this branch.

MaxKellermann:
You added this as a regular MPD tag. This will still break the protocol, because now we have a regular tag with newlines.

NBonaparte:
I'm unable to reproduce this. Without the lyrics tag in the `metadata_to_use` setting, the lyrics tag cannot be queried (obviously) and the `readlyrics` command works as expected. With `metadata_to_use = +lyrics`, the lyrics show up as tags, but with newlines stripped so everything is in one line.

MaxKellermann:
With a long enough line, libmpdclient will close the connection.
And it doesn't even make sense to have lyrics as a regular MPD tag. Just like the album cover wasn't implemented as a regular tag. Look at the album cover code.

NEWS
```diff
@@ -1947,7 +1948,7 @@ ver 0.12.0 (2006/9/22)
 * New --kill argument for killing MPD if pid_file is specified
 * Removed --update-db argument (use the update function in your client instead)
 * New mpdconf.example
-* New mpd.conf man page 
+* New mpd.conf man page
```
MaxKellermann: Whitespace change?

NEWS
```diff
@@ -1988,7 +1989,7 @@ the write() becomes successful)
 7) If "user" is specified, then convert ~ in paths to the user's home path
 specified by "user" config paramter (not the actual current user running mpd).
 
-ver 0.11.2 (2004/7/5) 
+ver 0.11.2 (2004/7/5)
```
MaxKellermann: Whitespace change?

NEWS
```diff
@@ -2162,7 +2163,7 @@ non existing file.  This causes parsing errors for clients.
 ver 0.8.4 (2003/8/13)
 1) Fix a bug where garbage is returned with errors in "list" command
 
-ver 0.8.3 (2003/8/12) 
+ver 0.8.3 (2003/8/12)
```
MaxKellermann: Whitespace change?

doc/protocol.rst
```diff
@@ -1007,6 +1007,11 @@ The music database
     decoder plugins support it.  For example, on Ogg files,
     this lists the Vorbis comments.
 
+:command:`readlyrics {URI}`
+    Read lyrics from the file specified by "URI". This "URI"
+    can be a path relative to the music directory or an
+    absolute path.
```
MaxKellermann: How does the response look like?

doc/protocol.rst
```diff
@@ -1275,7 +1280,7 @@ Audio output devices
 
 :command:`outputs`
     Shows information about all outputs.
-    
+
```
MaxKellermann: Whitespace change?

src/command/FileCommands.cxx
```diff
@@ -330,3 +330,28 @@ handle_read_picture(Client &client, Request args, Response &r)
...
+class PrintLyricsHandler final : public NullTagHandler {
+	Response &response;
+
+public:
+	explicit PrintLyricsHandler(Response &_response) noexcept
+		:NullTagHandler(WANT_LYRICS), response(_response) {}
+
+	void OnLyrics(StringView value) noexcept override {
+		response.Format("%.*s\n",
+				int(value.size), value.data);
```
MaxKellermann: So this writes the lyrics right into the stream, raw. How does the client know when the lyrics are finished?
What if there are multiple lyrics tags?

MaxKellermann:
Let's reopen this as soon as you have solved the fundamental problems.

ddonindia:
waiting for this for long time :(

jcorporation:
> How does the client know when the lyrics are finished?

One possibility could be to print the size of lyrics as a pair before the lyrics. The client can zehn use the new mpd_recv_raw function to get the lyrics.

PR #667, opened 2019-10-20
Sonos httpd mixer (requires noson lib)

skidoo23:
As requested from Max Kellermann. Published for review not for production.

Build instructions for noson library
```
$ git clone https://github.com/janbar/noson.git
$ cd noson
$ mkdir build
$ cd build
$ cmake -DCMAKE_BUILD_TYPE=Release -DHAVE_FLAC=0 -DHAVE_PULSEAUDIO=0 ..
$ make
# make install
```

Example mpd.conf
```
audio_output {
        type            "httpd"
        tags            "yes"
        name            "Local httpd stream for Sonos"
        always_on       "yes"
        encoder         "lame"                  # no flac or vorbis suppported
        port            "8000"
        bind_to_address "0.0.0.0"               # optional, IPv4 or IPv6
        bitrate         "320"                   # max quality suppported
        format          "44100:16:2"
#       max_clients     "0"                     # optional 0=no limit
        mixer_type      "httpd"
        mixer_zone      ""                      # "" connects for first zone discovered
        mixer_stream    "http://192.168.123.123:8000"    # name or IPv4 of MPD
}
```


MaxKellermann:
Havn't looked at any code yet, but it doesn't compile:
```
../../src/mixer/plugins/HttpdMixerPlugin.cxx:31:6: error: no previous declaration for ‘void handleEventCB(void*)’ [-Werror=missing-declarations]
 void handleEventCB(void* handle)
      ^~~~~~~~~~~~~
../../src/mixer/plugins/HttpdMixerPlugin.cxx: In function ‘void handleEventCB(void*)’:
../../src/mixer/plugins/HttpdMixerPlugin.cxx:31:26: error: unused parameter ‘handle’ [-Werror=unused-parameter]
 void handleEventCB(void* handle)
                    ~~~~~~^~~~~~
```

MaxKellermann:
And of course your libnoson detection is faulty - TravisCI fails even though your feature is disabled.

skidoo23:
> And of course your libnoson detection is faulty - TravisCI fails even though your feature is disabled.

Code compiles with above quoted warnings. Code is taken 1:1 from noson example source. I did not bother to fix because you don't like the noson library. I have to agree with you - sometimes connecting to Sonos devices takes a few seconds and blocks MPD control.

skidoo23:
> And of course your libnoson detection is faulty - TravisCI fails even though your feature is disabled.

Travis complains about missing noson and openssl (linker error without it) dependencies. PR does not include noson library. How to force travis to use both dependencies?

MaxKellermann:
> Travis complains about missing noson and openssl (linker error without it) dependencies.

No. OpenSSL was found, but libnoson was not. The problem is not that any of that is missing - the problem is that your code fails to build if libnoson is not available. Since libnoson is an optional dependency (and should be), there should not be a build failure.

skidoo23:
Understood. I will fix it.

skidoo23:
New PR for noson lib 2.0.0 is on its way.

PR #669, opened 2019-10-24
tag/Type: add tag "Conductor"

smutbert:


MaxKellermann:
You submitted this as 4 commits, but this separation doesn't make sense, so I folded those 4 commits into one.
4ecd4761c205c2f92de96236b8a6c275a8dfd53a

PR #670, opened 2019-10-27
output/Jack: mark ports as terminal

haasn:
This is the correct thing to do for ports from which a signal ultimately
"originates", such as is the case with mpd.


haasn:
I don't actually remember why I wanted or needed this change, nor why I never submitted this patch upstream. I just found it lying around and figured I might as well.

MaxKellermann:
Cherry-picked to v0.21.x
65778a377473bf9034c23a93e8fed5a65af5decf

PR #672, opened 2019-10-28
doc/user.rst: update build dependencies on Debian Buster

mxjeff:
Update Debian/Buster build dep. as used in debian package.

### DIFF
--- a/doc/user.rst
+++ b/doc/user.rst
@@ -62,16 +62,16 @@ In any case, you need:
 Each plugin usually needs a codec library, which you also need to
 install. Check the :doc:`plugins` for details about required libraries
 
-For example, the following installs a fairly complete list of build dependencies on Debian Jessie:
+For example, the following installs a fairly complete list of build dependencies on Debian Buster:
 
 .. code-block:: none
 
-    apt install g++ \
+    apt install meson g++ \
       libpcre3-dev \
       libmad0-dev libmpg123-dev libid3tag0-dev \
-      libflac-dev libvorbis-dev libopus-dev \
+      libflac-dev libvorbis-dev libopus-dev libogg-dev \
       libadplug-dev libaudiofile-dev libsndfile1-dev libfaad-dev \
-      libfluidsynth-dev libgme-dev libmikmod2-dev libmodplug-dev \
+      libfluidsynth-dev libgme-dev libmikmod-dev libmodplug-dev \
       libmpcdec-dev libwavpack-dev libwildmidi-dev \
       libsidplay2-dev libsidutils-dev libresid-builder-dev \
       libavcodec-dev libavformat-dev \
@@ -91,7 +91,9 @@ For example, the following installs a fairly complete list of build dependencies
       libsystemd-dev \
       libgtest-dev \
       libboost-dev \
-      libicu-dev
+      libicu-dev \
+      libchromaprint-dev \
+      libgcrypt20-dev
       
 
 Now configure the source tree:


doc/user.rst
```diff
@@ -62,27 +62,26 @@ In any case, you need:
...
       libpcre3-dev \
       libmad0-dev libmpg123-dev libid3tag0-dev \
-      libflac-dev libvorbis-dev libopus-dev \
+      libflac-dev libvorbis-dev libopus-dev libogg-dev \
       libadplug-dev libaudiofile-dev libsndfile1-dev libfaad-dev \
-      libfluidsynth-dev libgme-dev libmikmod2-dev libmodplug-dev \
+      libfluidsynth-dev libgme-dev libmikmod-dev libmodplug-dev \
       libmpcdec-dev libwavpack-dev libwildmidi-dev \
-      libsidplay2-dev libsidutils-dev libresid-builder-dev \
+      libsidplayfp-dev libsidutils-dev libresid-builder-dev \
```
MaxKellermann: Why switch from libsidplay to libsidplayfp? Both are included in Buster.

doc/user.rst
```diff
@@ -62,27 +62,26 @@ In any case, you need:
...
       libavcodec-dev libavformat-dev \
-      libmp3lame-dev libtwolame-dev libshine-dev \
```
MaxKellermann: Why remove twolame and shine?

doc/user.rst
```diff
@@ -62,27 +62,26 @@ In any case, you need:
...
+      libmp3lame-dev \
       libsamplerate0-dev libsoxr-dev \
       libbz2-dev libcdio-paranoia-dev libiso9660-dev libmms-dev \
       libzzip-dev \
-      libcurl4-gnutls-dev libyajl-dev libexpat-dev \
+      libcurl4-gnutls-dev libyajl-dev ibexpat1-dev \
```
MaxKellermann: Typo?
What's the point of adding the SO version to the package name?

doc/user.rst
```diff
@@ -62,27 +62,26 @@ In any case, you need:
...
       libasound2-dev libao-dev libjack-jackd2-dev libopenal-dev \
       libpulse-dev libshout3-dev \
-      libsndio-dev \
```
MaxKellermann: Why?

doc/user.rst
```diff
@@ -91,7 +90,11 @@ For example, the following installs a fairly complete list of build dependencies
       libsystemd-dev \
       libgtest-dev \
       libboost-dev \
-      libicu-dev
+      libicu-dev \
+      libchromaprint-dev \
+      libgcrypt20-dev \
+      python3-sphinx \
```
MaxKellermann: Sphinx isn't needed to build MPD. And since its dependencies are very heavy, I don't want to make everybody install it without a good reason.

doc/user.rst
```diff
@@ -91,7 +90,11 @@ For example, the following installs a fairly complete list of build dependencies
...
+      libcppunit-dev
```
MaxKellermann: Only needed for developing MPD. Pointless for users.

doc/user.rst
mxjeff: I don't know, [Florian switched to libsidplayfp in 2016](https://salsa.debian.org/mpd-team/mpd/commit/34bef5eb) but no motivation is included in the changelog. Do you recommend libsidplay over libsidplayfp?

MaxKellermann: I use neither, I don't have any SID files (except for MPD development), I have no idea which library fork is better. But your commit was about changes for Buster, and this is not one.
If somebody argues that the MPD documentation should suggest one or the other, that would be okay for me, but this way, it's hidden inside an unrelated commit, and I don't like that.
Maybe we should just remove this library from here, because I think only very few users are interested in SID playback.

doc/user.rst
mxjeff: Not certain about `twolame`, it provides support for "MPEG Audio Layer 2" but I think libmad also provides it.

Regarding `shine` we only enable it on armhf in the debian build but your right, it should stay in the user doc.

doc/user.rst
mxjeff: Right, should be fixed in the debian package as well (libexpat-dev is a virtual package, i guess we missed it).

doc/user.rst
MaxKellermann: I'm ok with removing useless libraries, but not hidden inside a commit pretending to do the Buster update.

doc/user.rst
mxjeff: Build was failing on Stretch with libsndio, can't remember why. We should revert it.

mxjeff:
I updated my changes according to your comments. In case you are ok with it, do you want me to squash my commits?

MaxKellermann:
I prefer to have only known-good commits in a PR, instead of known-bad ones along with fixup commits. The latest overall diff looks good.

mxjeff:
Here it is, thanks Max

MaxKellermann:
Cherry-picked to v0.21.x d9841668ffbeeb756f4c4895b41a06c1376d9613

PR #674, opened 2019-10-29
Build documentation with sphinx_rtd_theme

mxjeff:
With latter version of meson (>= 0.51.0) it could be more elegant using
something like:

    import('python').find_installation(modules : ['sphinx_rtd_theme'])

https://mesonbuild.com/Python-module.html

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -1,6 +1,11 @@
 install_man(['mpd.1', 'mpd.conf.5'])
 
 sphinx = find_program('sphinx-build')
+py = import('python').find_installation()
+if run_command(py, '-c', 'import sphinx_rtd_theme').returncode() != 0
+  error('python module "sphinx_rtd_theme" not found')
+endif
+
 sphinx_output = custom_target(
   'HTML documentation',
   output: 'html',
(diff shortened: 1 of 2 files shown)


doc/meson.build
```diff
@@ -1,6 +1,11 @@
 install_man(['mpd.1', 'mpd.conf.5'])
 
 sphinx = find_program('sphinx-build')
+py = import('python').find_installation()
```
MaxKellermann: This doesn't check if this is the Python version used by Sphinx, i.e. this is fragile.

doc/meson.build
```diff
@@ -1,6 +1,11 @@
...
+if run_command(py, '-c', 'import sphinx_rtd_theme').returncode() != 0
+  error('python module "sphinx_rtd_theme" not found')
```
MaxKellermann: This fails after Sphinx has been detected already. This breaks the whole build for people who are not interested in building documentation, but happen to have Sphinx installed, but this theme is missing.

doc/meson.build
```diff
@@ -10,7 +15,7 @@ sphinx_output = custom_target(
     'protocol.rst',
     'conf.py',
   ],
-  command: [sphinx, '-q', '-b', 'html', '-d', '@OUTDIR@/doctrees', meson.current_source_dir(), '@OUTPUT@'],
+  command: [sphinx, '-q', '-b', 'html', '-d', '@OUTDIR@/doctrees', meson.current_source_dir(), '@OUTPUT@', '-D',  'html_theme=sphinx_rtd_theme'],
```
MaxKellermann: This parameter belongs in `doc/conf.py'.

PR #680, opened 2019-11-07
AlsaMixerPlugin.cxx: prevent alsa assertion failure

borine:
When a hot-pluggable mixer is removed, the mixer handle becomes invalid: this commit handles the SND_CTL_EVENT_MASK_REMOVE event to prevent MPD from using that invalid handle


borine:
Sigh! sorry, this fix is not sufficient - it just moves the failure elsewhere. Please leave this alone while I investigate further.

borine:
Now I think the fix is complete.

borine:
Ugh - so now there is a memory leak. Right now I don't see an easy way for the AlsaMixer to handle pluggable controls. I'm closing the PR; I'll have a rethink, and then either try again or maybe just raise an issue.

PR #686, opened 2019-11-30
Fix typo in documentation

naglis:
Also changed _at the expensive of_ with _at the expense of_ since it is [more common](https://ell.stackexchange.com/a/80057).


MaxKellermann:
Cherry-picked to v0.21.x: 9766ac6db3294075c19c7b7efcc277440b50b3d1

PR #509, opened 2019-03-16
output/plugins/OSXOutputPlugin: add boost meson dependency

jacobvosmaer:
Fixes compilation error:

```
[259/516] Compiling C++ object 'src/output/plugins/0f714b4@@output_plugins@sta/OSXOutputPlugin.cxx.o'.
FAILED: src/output/plugins/0f714b4@@output_plugins@sta/OSXOutputPlugin.cxx.o 
c++ -Isrc/output/plugins/0f714b4@@output_plugins@sta -Isrc/output/plugins -I../src/output/plugins -Isrc -I../src -I. -I../ -Xclang -fcolor-diagnostics -pipe -Wall -Winvalid-pch -Wnon-virtual-dtor -std=c++14 -g -Wall -Wextra -fvisibility=hidden -ffast-math -ftree-vectorize -fno-threadsafe-statics -fmerge-all-constants -Wmissing-declarations -Wshadow -Wpointer-arith -Wcast-qual -Wwrite-strings -Wsign-compare -Wno-non-virtual-dtor -Wno-noexcept-type -funwind-tables -D_GNU_SOURCE  -MD -MQ 'src/output/plugins/0f714b4@@output_plugins@sta/OSXOutputPlugin.cxx.o' -MF 'src/output/plugins/0f714b4@@output_plugins@sta/OSXOutputPlugin.cxx.o.d' -o 'src/output/plugins/0f714b4@@output_plugins@sta/OSXOutputPlugin.cxx.o' -c ../src/output/plugins/OSXOutputPlugin.cxx
../src/output/plugins/OSXOutputPlugin.cxx:41:10: fatal error: 'boost/lockfree/spsc_queue.hpp' file not found
#include <boost/lockfree/spsc_queue.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
ninja: build stopped: subcommand failed.
```

I don't know why this didn't come up around the time I submitted https://github.com/MusicPlayerDaemon/MPD/pull/435; it looks like the same type of problem.

PR #690, opened 2019-12-09
Add missing meson dependencies (fixes MacPorts build)

jacobvosmaer:
I'm not sure why it keeps happening to me, but on my macOS / MacPorts machine, Meson is finnicky about dependencies. Without these changes I get "header not found" errors.

### DIFF
--- a/src/filter/plugins/meson.build
+++ b/src/filter/plugins/meson.build
@@ -23,6 +23,9 @@ filter_plugins = static_library(
   'VolumeFilterPlugin.cxx',
   filter_plugins_sources,
   include_directories: inc,
+  dependencies: [
+    filter_plugins_deps,
+  ],
 )
 
 filter_plugins_dep = declare_dependency(
--- a/src/input/meson.build
+++ b/src/input/meson.build
@@ -40,6 +40,9 @@ input_glue = static_library(
   'cache/Item.cxx',
   'cache/Stream.cxx',
   include_directories: inc,
+  dependencies: [
+    boost_dep,
+  ],
 )
 
 input_glue_dep = declare_dependency(
--- a/src/lib/xiph/meson.build
+++ b/src/lib/xiph/meson.build
@@ -52,6 +52,9 @@ xiph = static_library(
   'VorbisPicture.cxx',
   'XiphTags.cxx',
   include_directories: inc,
+  dependencies: [
+    libvorbis_dep,
+  ],
 )
 
 xiph_dep = declare_dependency(


jacobvosmaer:
This is the same problem I had back in https://github.com/MusicPlayerDaemon/MPD/pull/509, but for different meson.build files.

src/input/meson.build
```diff
@@ -40,6 +40,9 @@ input_glue = static_library(
   'cache/Item.cxx',
   'cache/Stream.cxx',
   include_directories: inc,
+  dependencies: [
+    boost_dep,
```
jacobvosmaer: Example error:

```
In file included from ../src/input/cache/Item.cxx:20:
../src/input/cache/Item.hxx:26:10: fatal error: 'boost/intrusive/list.hpp' file not found
#include <boost/intrusive/list.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

src/lib/xiph/meson.build
```diff
@@ -52,6 +52,9 @@ xiph = static_library(
   'VorbisPicture.cxx',
   'XiphTags.cxx',
   include_directories: inc,
+  dependencies: [
+    libvorbis_dep,
```
jacobvosmaer:
```
../src/lib/xiph/VorbisComments.cxx:33:10: fatal error: 'vorbis/codec.h' file not found
#include <vorbis/codec.h>
         ^~~~~~~~~~~~~~~~
1 error generated.
```

src/filter/plugins/meson.build
```diff
@@ -23,6 +23,9 @@ filter_plugins = static_library(
   'VolumeFilterPlugin.cxx',
   filter_plugins_sources,
   include_directories: inc,
+  dependencies: [
+    filter_plugins_deps,
```
jacobvosmaer:
```
../src/lib/ffmpeg/Filter.hxx:26:10: fatal error: 'libavfilter/avfilter.h' file not found
#include <libavfilter/avfilter.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

jacobvosmaer:
Updated the title to reference MacPorts.

MacPorts is funny in that it puts everything in `/opt/local`. I wonder if "normal" MPD builds, with header files under `/usr` or `/usr/local`, succeed without these additions because the required header files get picked up via default search paths. I don't think `/opt/local` is in a lot of default search paths, so it forces us to be more precise about where the header files are.

This is speculation but I think it's plausible.

MaxKellermann:
First patch merged, the other two cherry-picked to v0.21.x

jacobvosmaer:
Thanks @MaxKellermann !

Issue #691, opened 2019-12-11
MPD crash (boost assertion failure)


jacobvosmaer:
### Describe the bug

Crash while listening to music.

## Expected Behavior

Not crashing

## Actual Behavior

A crash

## Version

```
Music Player Daemon 0.22~git (v0.21.16-650-g8e7ab9653)
Copyright 2003-2007 Warren Dukes <warren.dukes@gmail.com>
Copyright 2008-2018 Max Kellermann <max.kellermann@gmail.com>
This is free software; see the source for copying conditions.  There is NO
warranty; not even MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Database plugins:
 simple proxy

Storage plugins:
 local curl


Decoders plugins:
 [vorbis] ogg oga
 [oggflac] ogg oga
 [flac] flac
 [opus] opus ogg oga
 [sndfile] wav aiff aif au snd paf iff svx sf voc w64 pvf xi htk caf sd2
 [audiofile] wav au aiff aif
 [dsdiff] dff
 [dsf] dsf
 [hybrid_dsd] m4a
 [wavpack] wv
 [modplug] 669 amf ams dbm dfm dsm far it med mdl mod mtm mt2 okt s3m stm ult umx xm
 [ffmpeg] 16sv 3g2 3gp 4xm 8svx aa3 aac ac3 adx afc aif aifc aiff al alaw amr anim apc ape asf atrac au aud avi avm2 avs bap bfi c93 cak cin cmv cpk daud dct divx dts dv dvd dxa eac3 film flac flc fli fll flx flv g726 gsm gxf iss m1v m2v m2t m2ts m4a m4b m4v mad mj2 mjpeg mjpg mka mkv mlp mm mmf mov mp+ mp1 mp2 mp3 mp4 mpc mpeg mpg mpga mpp mpu mve mvi mxf nc nsv nut nuv oga ogm ogv ogx oma ogg omg opus psp pva qcp qt r3d ra ram rl2 rm rmvb roq rpl rvc shn smk snd sol son spx str swf tak tgi tgq tgv thp ts tsp tta xa xvid uv uv2 vb vid vob voc vp6 vmd wav webm wma wmv wsaud wsvga wv wve
 [pcm]

Filters:
 soxr

Tag plugins:
 id3tag

Output plugins:
 null fifo pipe ao openal osx httpd recorder

Encoder plugins:
 null vorbis opus twolame wave flac

Archive plugins:
[...]

PR #696, opened 2019-12-18
input/cache: use simpler intrusive::set API

jacobvosmaer:
This fixes https://github.com/MusicPlayerDaemon/MPD/issues/691

In that issue we found a bug in `InputCacheManager::Get` due to incorrect use of the Boost intrusive::set API. The pair of functions `insert_check`/`insert_commit` allow you to amortize the cost of a tree walk across two function calls. This performance gain comes at the cost of complexity: you may not modify the tree in between `insert_check` and `insert_commit`.

If you do modify the tree, you will get sporadic runtime assertion failures (i.e. crashes). This is the cause of #691.

My gut feeling is that `InputCacheManager::Get` is too complex to use `insert_check` / `insert_commit` correctly. I would have to add comments to remind myself of what is going on, and why things need to happen in just the right order. I also don't see a way to make the function simpler. So for the sake of readability and robustness, I think we shouldn't use `insert_check` / `insert_commit` here and just use `find` and `insert`.

### DIFF
--- a/src/input/cache/Manager.cxx
+++ b/src/input/cache/Manager.cxx
@@ -81,11 +81,9 @@ InputCacheManager::Get(const char *uri, bool create)
 	if (!PathTraitsUTF8::IsAbsolute(uri))
 		return {};
 
-	UriMap::insert_commit_data hint;
-	auto result = items_by_uri.insert_check(uri, items_by_uri.key_comp(),
-						hint);
-	if (!result.second) {
-		auto &item = *result.first;
+	auto iter = items_by_uri.find(uri, items_by_uri.key_comp());
+	if (iter != items_by_uri.end()) {
+		auto &item = *iter;
 
 		/* refresh */
 		items_by_time.erase(items_by_time.iterator_to(item));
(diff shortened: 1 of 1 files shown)


src/input/cache/Manager.cxx
```diff
@@ -81,11 +81,9 @@ InputCacheManager::Get(const char *uri, bool create)
...
 		return {};
 
-	UriMap::insert_commit_data hint;
-	auto result = items_by_uri.insert_check(uri, items_by_uri.key_comp(),
-						hint);
-	if (!result.second) {
-		auto &item = *result.first;
+	auto iter = items_by_uri.find(uri, items_by_uri.key_comp());
+	if (iter != items_by_uri.end()) {
+		auto &item = *iter;
```
jacobvosmaer: I am not an experienced C++ programmer. This patch is the result of me bashing the keyboard until the build succeeds.

jacobvosmaer: I'm also using this binary to listen to music, with the input cache enabled. But I was never able to trigger the bug this should solve, so I don't consider that very thorough testing.

jacobvosmaer:
I edited the PR description to make my reasoning clearer.

PR #697, opened 2019-12-25
StringBuffer: Include cstddef for size_t

kraj:
Fixes
a.cpp:3:1: error: 'size_t' does not name a type
    3 | size_t s;
      | ^~~~~~
a.cpp:2:1: note: 'size_t' is defined in header '<cstddef>'; did you forget to '#include <cstddef>'?

### DIFF
--- a/src/util/StringBuffer.hxx
+++ b/src/util/StringBuffer.hxx
@@ -31,6 +31,7 @@
 #define STRING_BUFFER_HXX
 
 #include <array>
+#include <cstddef>
 
 /**
  * A statically allocated string buffer.


MaxKellermann:
Which compiler and C++ library are you using? Including `<array>` should make `std::size_t` available, because `std::size_t` is a template parameter type of `std::array`.

kraj:
Gcc trunk and this won’t be the case in gcc10+

MaxKellermann:
So this is a compiler/library bug, and your patch is a workaround for the bug?
It is important for me to know whether MPD's code is correct or not. If it's buggy, and the include is necessary, we'll have many more headers which need it.
But if MPD is correct, then we shouldn't bother having workarounds for unreleased compiler bugs.

kraj:
> So this is a compiler/library bug, and your patch is a workaround for the bug?
> It is important for me to know whether MPD's code is correct or not. If it's buggy, and the include is necessary, we'll have many more headers which need it.
> But if MPD is correct, then we shouldn't bother having workarounds for unreleased compiler bugs.

no, on the contrary, older compilers indirectly include cstddef was a bug which compiler has fixed now. Applications should explicitly ask for standard headers which they depend on instead of relying on compiler behavior.

MaxKellermann:
> Applications should explicitly ask for standard headers which they depend on instead of relying on compiler behavior.

I understand that. But did you consider what I wrote why I believe that including `<cstddef>` for `size_t` should not be necessary?

Let me repeat with more detail: the C++ standard section 23.3.1 "Class template array" says:
```
namespace std {
  template <class T, size_t N >
  struct array {
```

This implies that including `<array>` must make `std::size_t` available, doesn't it?
An implementation of `<array>` which doesn't make `std::size_t` available violates the C++ standard and is buggy, in my opinion.

What is your interpretation of this part of the C++ standard? Why do you believe that the type name used in the C++ standard can be undefined?

(Yes, this is language-lawyering. I want to know the nature of the problem. This goes deeper, as I noted before - and let me repeat one more thing: if you're correct, we need many more fixes in many other source files.)

kraj:
> > Applications should explicitly ask for standard headers which they depend on instead of relying on compiler behavior.
> 
> I understand that. But did you consider what I wrote why I believe that including `<cstddef>` for `size_t` should not be necessary?

I did. You seem to be conflating `size_t` and `std::size_t`

> 
> Let me repeat with more detail: the C++ standard section 23.3.1 "Class template array" says:
> 
> ```
> namespace std {
>   template <class T, size_t N >
>   struct array {
> ```
> 
> This implies that including `<array>` must make `std::size_t` available, doesn't it?

yes and that all is fine as long as it is `std::size_t`. 

> An implementation of `<array>` which doesn't make `std::size_t` available violates the C++ standard and is buggy, in my opinion.
> 
> What is your interpretation of this part of the C++ standard? Why do you believe that the type name used in the C++ standard can be undefined?
> 
> (Yes, this is language-lawyering. I want to know the nature of the problem. This goes deeper, as I noted before - and let me repeat one more thing: if you're correct, we need many more fixes in many other source files.)

The key is the use of `size_t` vs `std::size_t`, if one was using `std::size_t` then including `<array>` would pull in needed bits for that. However, if `size_t` is used then its not in std namespace but anonymous and would require `<cstddef>` to be explicitly included. Earlier that happened via `<stdexcept>` indirectly.

MaxKellermann:
> I did. You seem to be conflating `size_t` and `std::size_t`

No, I don't. `StringBuffer.hxx` uses `std::size_t` and not `size_t`.

> yes and that all is fine as long as it is `std::size_t`.

Which it is!

> The key is the use of `size_t` vs `std::size_t`, if one was using `std::size_t` then including `<array>` would pull in needed bits for that.

Exactly my point!

Here's the source code of the commit id in your pull request:

https://github.com/MusicPlayerDaemon/MPD/blob/03521c4406511260262fb33331840d3d91e7763c/src/util/StringBuffer.hxx#L34-L46

That's `std::size_t` here!

kraj:
ah, now I see, I am looking at older codebase on v0.20.x ( from OpenEmbedded recipe ) where its missing 2e8f42c6ad, so I guess you have already taken care of size_t to std::size_t migration. so this patch can be considered obsolete for master.

Issue #698, opened 2019-12-30
Compilation error (fall through)


Wang-Yue:
```
[421/560] Compiling C object 'src/util/ed6d25d@@util@sta/format.c.o'.
FAILED: src/util/ed6d25d@@util@sta/format.c.o 
cc -Isrc/util/ed6d25d@@util@sta -Isrc/util -I../../src/util -Isrc -I../../src -I. -I../../ -I/usr/local/opt/expat/include -Xclang -fcolor-diagnostics -DNDEBUG -pipe -Wall -Winvalid-pch -std=c99 -O2 -g -Wall -Wextra -Wvla -fvisibility=hidden -ffast-math -ftree-vectorize -Wmissing-prototypes -Wshadow -Wpointer-arith -Wstrict-prototypes -Wcast-qual -Wwrite-strings -pedantic -ffunction-sections -fdata-sections -D_GNU_SOURCE -MD -MQ 'src/util/ed6d25d@@util@sta/format.c.o' -MF 'src/util/ed6d25d@@util@sta/format.c.o.d' -o 'src/util/ed6d25d@@util@sta/format.c.o' -c ../../src/util/format.c
../../src/util/format.c:242:4: error: expected expression
                        gcc_fallthrough;
                        ^
../../src/util/Compiler.h:149:25: note: expanded from macro 'gcc_fallthrough'
#define gcc_fallthrough [[fallthrough]]
                        ^
1 error generated.
[426/560] Compiling C++ object 'mpd@exe/src_db_update_Archive.cxx.o'.
ninja: build stopped: subcommand failed.


$ clang --version
Apple clang version 11.0.0 (clang-1100.0.33.16)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin


$ uname  -a
Darwin wangyue-macbookpro.roam.corp.google.com 18.7.0 Darwin Kernel Version 18.7.0: Sun Dec  1 18:59:03 PST 2019; root:xnu-4903.278.19~1/RELEASE_X86_64 x86_64
```
latest xcode version.

jacobvosmaer:
I have the same problem.

```
$ cc --version
Apple clang version 11.0.0 (clang-1100.0.33.16)
Target: x86_64-apple-darwin18.6.0
Thread model: posix
[...]

PR #699, opened 2019-12-30
util: do not use (C++) fallthrough in C file

jacobvosmaer:
Fixes https://github.com/MusicPlayerDaemon/MPD/issues/698

It appears `[[fallthrough]]` is specific to C++. In a1afe9afc6879960389491b6fb587f14ae1ae5df we added this attribute to `src/util/format.c`, which is not a C++ file. Clang 11 on macOS does not like this: it causes compilation to fail with:

```
../src/util/format.c:242:4: error: expected expression
                        gcc_fallthrough;
                        ^
../src/util/Compiler.h:149:25: note: expanded from macro 'gcc_fallthrough'
#define gcc_fallthrough [[fallthrough]]
```

### DIFF
--- a/src/util/format.c
+++ b/src/util/format.c
@@ -19,7 +19,6 @@
  */
 
 #include "format.h"
-#include "util/Compiler.h"
 
 #include <stdbool.h>
 #include <stdio.h>
@@ -239,7 +238,6 @@ format_object2(const char *format, const char **last, const void *object,
 			}
 
 			/* fall through */
-			gcc_fallthrough;
 
 		default:
 			/* pass-through non-escaped portions of the format string */


jacobvosmaer:
I also tried using `__attribute__ ((fallthrough))`. That does not look right either:

```
../src/util/format.c:242:4: warning: declaration does not declare anything
      [-Wmissing-declarations]
                        __attribute__ ((fallthrough));
                        ^
```

jacobvosmaer:
OK this does not work because now GCC is not happy. (the travis build failed)

jacobvosmaer:
I think I see an ifdef-workaround...

jacobvosmaer:
Closing in favor of https://github.com/MusicPlayerDaemon/MPD/pull/700

Issue #673, opened 2019-10-28
Failed to bind socket


talentlessguy:
## Question

I start an `mpd` service with `mpd.socket` but both start with errors. I need to fix it but I don't know what is wrong.

scanning my local machine ports doesn't show that 6600 is in use:

```
Starting Nmap 7.80 ( https://nmap.org ) at 2019-10-28 18:37 MSK
Nmap scan report for localhost (127.0.0.1)
Host is up (0.000088s latency).
Other addresses for localhost (not scanned): ::1
Not shown: 998 closed ports
PORT     STATE SERVICE
631/tcp  open  ipp
5432/tcp open  postgresql
```

`mpd.socket` errors:

```
Oct 28 18:31:42 v1rtl-pc systemd[7074]: mpd.socket: Failed to create listening socket ([::]:6600): Address already in use
Oct 28 18:31:42 v1rtl-pc systemd[1]: mpd.socket: Failed to receive listening socket ([::]:6600): Input/output error
Oct 28 18:31:42 v1rtl-pc systemd[1]: mpd.socket: Failed to listen on sockets: Input/output error
Oct 28 18:31:42 v1rtl-pc systemd[1]: mpd.socket: Failed with result 'resources'.
```

`mpd` errors:

```
Oct 28 18:29:39 v1rtl-pc mpd[6835]: exception: bind to '0.0.0.0:6600' failed (continuing anyway, because binding to '[::]:6600' succeeded): Failed to bind socket: Address already in use
Oct 28 18:29:39 v1rtl-pc libsmbclient[6835]: exception: RTIOThread could not get realtime scheduling, continuing anyway: sched_setscheduler failed: Operation not permitted
Oct 28 18:29:39 v1rtl-pc systemd[1015]: Started Music Player Daemon.
```

Here's my `mpd.conf`:

```
audio_output {
        type            "pulse"
        name            "pulse audio"
}

# QOBUZ input plugin
input {
        enabled    "no"
        plugin     "qobuz"
#        app_id     "ID"
#        app_secret "SECRET"
#        username   "USERNAME"
#        password   "PASSWORD"
#        format_id  "N"
}

# TIDAL input plugin
input {
        enabled      "no"
        plugin       "tidal"
#        token        "TOKEN"
#        username     "USERNAME"
#        password     "PASSWORD"
#        audioquality "Q"
}

# Decoder #####################################################################
#

decoder {
        plugin                  "hybrid_dsd"
        enabled                 "no"
[...]

Issue #557, opened 2019-05-16
If bind_to_address is any, mpd tries to bind to the same address twice.


crocket:
mpd-0.21.8

MPD configuration

```
music_directory "/data/music"
playlist_directory "/srv/mpd-pi/playlists"
db_file "/srv/mpd-pi/database"
log_file "/var/log/pi/mpd-pi.log"
log_level "default"
bind_to_address "any"
bind_to_address "@mpd"
auto_update "no"
audio_output {
  type "jack"
  name "jack"
  mixer_type "software"
  client_name "mpd"
  server_name "default"
  source_ports "out_1,out_2"
  destination_ports "system:playback_1,system:playback_2"
  autostart "no"
  always_on "yes"
}
filesystem_charset "UTF-8"
restore_paused "yes"
metadata_to_use "+comment"

default_permissions "read"
local_permissions "read,add,control,admin"
password "****@read,add,control,admin"
```

Error

```
May 16 20:09 : exception: bind to '0.0.0.0:6600' failed (continuing anyway, because binding to '[::]:6600' succeeded): Failed to bind socket: Address already in use
```

If I replace `any` with `::`, `::` acts like `any`.
After binding to `[::]:6600`, binding to `0.0.0.0:6600` fails.

MaxKellermann:
> If bind_to_address is any, mpd tries to bind to the same address twice.

That's not the same address. `0.0.0.0` is the IPv4 wildcard, and `::` is the IPv6 wildcard. These are different addresses, even different address families.

> If I replace `any` with `::`, `::` acts like `any`.

Not exactly correct. `::` acts only like `any` if your kernel decides to bind the IPv6 wildcard to IPv4 as well. For example, the Linux kernel does this if you set `/proc/sys/net/ipv6/bindv6only` to 0. If this option is enabled or if your Linux kernel was compiled without IPv6 support, the behavior is very different. On other (non-Linux) kernels, it may be a completely different story as well. What Linux does here is non-standard.

crocket:
Right, `/proc/sys/net/ipv6/bindv6only` is already 0.

crocket:
Perhaps, you want to enable `IPV6_V6ONLY` in socket. It seems that the default value for `/proc/sys/net/ipv6/bindv6only` is 0 according to `man ipv6`.

MaxKellermann:
[...]

PR #712, opened 2020-01-16
SocketUtil: explicitly set V6ONLY socket option

nmeum:
Currently, MPD emits the following warning on Linux when
`net.ipv6.bindv6only` is set to the default value zero:

	exception: bind to '0.0.0.0:6600' failed (continuing anyway, because binding to '[::]:6600' succeeded): Failed to bind socket: Address already in use

This is a known issues (#673, #557, …) caused by the fact that MPD, by
default, creates separate sockets for `AF_INET` and `AF_INET6`. With
this architecture it is not necessary to use the `AF_INET6` socket for
both IPv6 and IPv4. For this reason, explicitly enable the V6ONLY
socket option for `AF_INET6` sockets, thereby preventing the error
causing the warning to be emitted.

### DIFF
--- a/src/net/SocketDescriptor.cxx
+++ b/src/net/SocketDescriptor.cxx
@@ -322,12 +322,6 @@ SocketDescriptor::SetTcpUserTimeout(const unsigned &milliseconds) noexcept
 }
 
 bool
-SocketDescriptor::SetV6Only(bool value) noexcept
-{
-	return SetBoolOption(IPPROTO_IPV6, IPV6_V6ONLY, value);
-}
-
-bool
 SocketDescriptor::SetBindToDevice(const char *name) noexcept
 {
 	return SetOption(SOL_SOCKET, SO_BINDTODEVICE, name, strlen(name));
@@ -377,6 +371,12 @@ SocketDescriptor::AddMembership(SocketAddress address) noexcept
 }
 
 #endif
+
+bool
+SocketDescriptor::SetV6Only(bool value) noexcept
+{
+	return SetBoolOption(IPPROTO_IPV6, IPV6_V6ONLY, value);
+}
 
 bool
 SocketDescriptor::Bind(SocketAddress address) noexcept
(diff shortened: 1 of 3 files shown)


MaxKellermann:
I thought of doing this a few times, but didn't because I didn't want to override the admin's global setting for this. By default, MPD should be a good citizen and should follow global configuration.
And don't forget: your proposed code change breaks configurations where admins have disabled "v6only" globally and made MPD bind to "::"; with your change, they can't use IPv4 anymore.
Do you have a good justification for this breaking change?

nmeum:
> And don't forget: your proposed code change breaks configurations where admins have disabled "v6only" globally and made MPD bind to "::"; with your change, they can't use IPv4 anymore.

I think the amount of people doing that is close to zero. Besides, if you bind a socket to a specific IPv6 address through a software configuration option I would not expect the software to create an IPv4 socket as well. Also keep in mind that this only works for `in6addr_any`. If you bind explicitly to some other IPv6 address it will only create an IPv6 socket, even on current git head. If the intention of the admin is to bind MPD to both IPv6 and IPv4 he can simply use `bind_to_address "any"` instead of `bind_to_address "::"` in the first place.

> I thought of doing this a few times, but didn't because I didn't want to override the admin's global setting for this. By default, MPD should be a good citizen and should follow global configuration.

I don't think that this is a popular interpretation of `net.ipv6.bindv6only`. In my experience most software sets this socket option explicitly, either creating a single IPv6 socket and disabling `V6ONLY` or creating two sockets and enabling `V6ONLY`.

neheb:
This needs to be rebased to pass Travis.

nmeum:
Rebased and fixed the build on Mac OS.

MaxKellermann:
> I don't think that this is a popular interpretation of `net.ipv6.bindv6only`.

I do think so, because that's how this sysctl is defined.

I'm not convinced that this change is a good idea. It can potentially break user setups and overrides global configuration for no reason.

Issue #567, opened 2019-05-29
status of replay_gain


mistepien:
## Feature request
It would be great to have status-like command in protocol to find out current status of replay_gain, especially what is actually applied to current song.

MaxKellermann:
There's `replay_gain_status`: https://www.musicpd.org/doc/html/protocol.html#playback-options

mistepien:
> replay_gain_status
>     Prints replay gain options. Currently, only the variable replay_gain_mode is returned.
I think about more details, what is ie. volume change of current song.

cataldor:
This pull request is related: https://github.com/MusicPlayerDaemon/MPD/pull/715

If accepted, i will add support for these tags in `libmpdclient`; then, it is easy to access it with, for instance, `mpc`

breversa:
I’m looking for a bs1770gain-supporting (http://bs1770gain.sourceforge.net/) music player. Is it a planned feature for MPD ?

Issue #219, opened 2018-02-12
Weird volume behavior when replay_gain_hander is specified as "mixer"


estshorter:
## Issues
1. A value from `mpc volume` varies by itself after playing a song
2. A dB value written in replaygain tag is quite different from that obtained by `amixer get Digital`

mpd: 0.20.17
outputs: alsa

## Detail
I prepared a test file (nosound, replaygain: -10db).
[01 no_sound rg-10.zip](https://github.com/MusicPlayerDaemon/MPD/files/1715561/01.no_sound.rg-10.zip)

Command log when playing the test file is as follows.
```
$ mpc volume 100
volume:100%   repeat: off   random: off   single: off   consume: off
$ mpc play
no_sound rg-10  
[playing] #1/1   0:00/5:00 (0%)
volume:100%   repeat: off   random: off   single: off   consume: off
$ mpc volume
volume: 30%
$ amixer get Digital
Simple mixer control 'Digital',0
  Capabilities: pvolume pswitch
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 207
  Mono:
  Front Left: Playback 145 [70%] [-31.00dB] [on]
  Front Right: Playback 145 [70%] [-31.00dB] [on]
```

As you can see above, initial volume (100%) changed to 30% after `mpc play`.
This is not the case when using only software volume; the volume is constant.

Moreover, the dB value written in the replaygain tag (-10dB) is quite different from that obtained by `amixer get Digital` (-31dB).
The latter value should be around -10dB.

## MPD verbose log
```
Feb 12 13:41 : state_file: Loading state file /tmp/mpd_state
Feb 12 13:41 : inotify: initializing inotify
Feb 12 13:41 : inotify: watching music directory
Feb 12 13:41 : client: [0] opened from [::1]:44438
Feb 12 13:41 : client: [0] process command "setvol "100""
Feb 12 13:41 : client: [0] command returned 0
Feb 12 13:41 : client: [0] process command list
Feb 12 13:41 : client: process command "status"
Feb 12 13:41 : client: command returned 0
Feb 12 13:41 : client: process command "currentsong"
Feb 12 13:41 : client: command returned 0
Feb 12 13:41 : client: [0] process command list returned 0
Feb 12 13:41 : client: [0] closed
Feb 12 13:41 : client: [1] opened from [::1]:44440
Feb 12 13:41 : client: [1] process command "play"
[...]

PR #715, opened 2020-01-21
Add support for ReplayGain tags

cataldor:
This adds support for seven ReplayGain tags.
------
Hello, this is my first commit for MPD.
I have separated this patch into two commits; the second one (https://github.com/MusicPlayerDaemon/MPD/commit/2fdbaf17dbbd5541b3460a258dbf5699b8303344) identify the MPD version for the protocol documentation; since no other MPD version is identified for tags, this may be unnecessary.

Tests were made using the loudgain program for FLAC (VorbisComments) and MP3 (Idv3) files. Unfortunately, i don't have an APE/Wavepak file to test ApeTag.


MaxKellermann:
What's the point of adding these tags and exposing them to clients?

cataldor:
Hello Max Kellerman,
the point is to provide the user the opportunity to fetch the replay gain values being applied on the current song.
I would point out two motivations:
1. (from the client point-of-view) currently, you are able to choose if you prefer applying album or track ReplayGain values by using the `replay_gain_mode` command. You can also fetch the current status through the `replay_gain_status` command. However, you are unable to fetch the ReplayGain values used by the current song (if any). By exporting these tags, you will be able to [1].

2. Fetching ReplayGain values were flagged as a feature for MPD in the issue https://github.com/MusicPlayerDaemon/MPD/issues/567.

As far as i know, you are unable to find out the current ReplayGain values from MPD; this patch solves this problem indirectly as it allows clients to provide this information.

[1] There may be songs that have ReplayGain applied to them, but no tags are present. The reason is that MPD parses replaygain data from Id3v2 by using case-insensitive comparison:
https://github.com/MusicPlayerDaemon/MPD/blob/1e07d1542856d63f026448eb37e1276576d8381b/src/tag/ReplayGain.cxx#L63
While tag searching is done with case-sensitive comparison: 
https://github.com/MusicPlayerDaemon/MPD/blob/1e07d1542856d63f026448eb37e1276576d8381b/src/tag/Table.cxx#L34. I would like to change that, but this would affect *all tags*. Would that be okay?
I noticed this because the user from the issue https://github.com/MusicPlayerDaemon/MPD/issues/219 generated an mp3 with ReplayGain values that will not generate the tag information with this patch because its Id3v2 uses lowercase, and MusicBrainz expects uppercase. I don't know if he added the ReplayGain data by hand or used another tool.

MaxKellermann:
>     1. [...] you are unable to fetch the ReplayGain values used by the current song (if any). By exporting these tags, you will be able to [1].

"You are not able to, but with this feature you are able to." Sure, that's deducted with simple logic, but it doesn't explain why one would want or need to.

>     2. Fetching ReplayGain values were flagged as a feature for MPD in the issue #567.

No. #667 is about exposing the currently applied value to the song currently being played. That's not what your code is about.

And it's not a reason either, even if this feature request was identical with your code; you are referring to a feature request which doesn't answer the "why?" question.

If it's so impossible to explain the reasons for this code change, it may not be a good idea after all.

PR #723, opened 2020-01-31
treewide: math fixes

neheb:
std::lround is C++11 and requires _GLIBCXX_USE_C99_MATH_TR1 to be
supported. This is not the case with uClibc-ng.

### DIFF
--- a/src/mixer/plugins/volume_mapping.c
+++ b/src/mixer/plugins/volume_mapping.c
@@ -34,11 +34,6 @@
 #include <stdbool.h>
 #include "volume_mapping.h"
 
-#ifdef __UCLIBC__
-/* 10^x = 10^(log e^x) = (e^x)^log10 = e^(x * log 10) */
-#define exp10(x) (exp((x) * log(10)))
-#endif /* __UCLIBC__ */
-
 #define MAX_LINEAR_DB_SCALE	24
 
 static inline bool use_linear_dB_scale(long dBmin, long dBmax)
@@ -111,9 +106,9 @@ static double get_normalized_volume(snd_mixer_elem_t *elem,
 	if (use_linear_dB_scale(min, max))
 		return (value - min) / (double)(max - min);
 
-	normalized = exp10((value - max) / 6000.0);
+	normalized = pow(10, (value - max) / 6000.0);
 	if (min != SND_CTL_TLV_DB_GAIN_MUTE) {
-		min_norm = exp10((min - max) / 6000.0);
+		min_norm = pow(10, (min - max) / 6000.0);
 		normalized = (normalized - min_norm) / (1 - min_norm);
 	}
 
@@ -159,7 +154,7 @@ static int set_normalized_volume(snd_mixer_elem_t *elem,
 	}
 
 	if (min != SND_CTL_TLV_DB_GAIN_MUTE) {
-		min_norm = exp10((min - max) / 6000.0);
+		min_norm = pow(10, (min - max) / 6000.0);
 		volume = volume * (1 - min_norm) + min_norm;
 	}
 	value = lrint_dir(6000.0 * log10(volume), dir) + max;
--- a/src/ReplayGainGlobal.cxx
+++ b/src/ReplayGainGlobal.cxx
@@ -20,12 +20,12 @@
 #include "ReplayGainGlobal.hxx"
 #include "ReplayGainConfig.hxx"
 #include "config/Data.hxx"
+#include "util/Math.hxx"
 
 #include <cassert>
 #include <stdexcept>
 
 #include <stdlib.h>
-#include <math.h>
 
 static float
 ParsePreamp(const char *s)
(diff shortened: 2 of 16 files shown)


MaxKellermann:
I don't quite understand yet. MPD requires a C++17 compliant compiler, and the C++17 standard mandates that `std::lround()` is available. So your build environment does not meet those requirements, not even C++11, right?

neheb:
It's not that. The build environment most definitely supports C++17 (2a even). It's just that the math functions are not exposed to the libstdc++ by the libc.

Here's the solution is boost: https://github.com/boostorg/math/blob/0f7b469ccb16efbb134edb70e3a55788d1a8ed87/include/boost/math/tools/roots.hpp#L800

MaxKellermann:
> to the libstdc++ by the libc.

For C++ standards compliance, I don't care what's exposed from libc to libstdc++. I care only if libstdc++ exposes all the standard C++ API to my application. If one API is not available, for whatever reason, then this build environment is not C++ compliant.

neheb:
This is the only thing needed to get it to build though...

neheb:
Note that there are several alreas already where lround is used (not the std variant).

MaxKellermann:
I don't accept changes like that if I don't understand the problem.
You insist that your build environment is C++11 compliant, and now that I'm trying to nail you down on it, you change the topic.

neheb:
My build environment is GCC9 with its corresponding libstdc++. It has support for C++20 stuff even.

The only difference is the libc. uClibc-ng instead of glibc. No other difference.

MaxKellermann:
I don't care what libc you use. I don't care if your toolchain has "C++20 stuff". This has nothing to do with this problem. Don't change the topic all the time, this is just wasting both of our time.

neheb:
You asked for my build environment...

MaxKellermann:
No, this was my only question here:
> So your build environment does not meet those requirements, not even C++11, right?

neheb:
All C++ functionality is present. The C++11 math functions are not. The reason is that the libc does not expose them to std::.

The reason is that the libc does not support the proper C99 TR1 math functionality as the define implies.

I have tested this with both libstdc++ (GCC) and libc++ (clang) and both show the same result.

neheb:
Now that I think about it, using the boost rounding functions might be appropriate as iround for example returns an int, which is the correct type to return in several places.

Also, boost's lround is available even when. std::lround is not.

neheb:
I did just that. Compilation with boost's rounding functions has been tested to work.

MaxKellermann:
I don't like Boost's implementation which adds some bloat to this trivial function call for error handling which I don't want or need.
Why don't you just admit that your build environment isn't fully C++11 compliant, because standard library functions are not available? Is that so hard to admit? Once you admit that, we could talk about working around this deficiency in your build environment, and call your orignal PR what it really is: a workaround for that deficiency.

neheb:
I'm all ears as to how that could be approached. The boost guys decided to add their own overloads.

neheb:
Third attempt. Added a fallback to boost::lround when std::lround is unavailable.

This also seems to work better. Compiled size went down. I'm guessing std::lround avoids several type promotions.

MaxKellermann:
That looks like an acceptable workaround, but I don't like the code duplication. Can you please move this ifdef orgy to, say, `src/util/Math.hxx`?

neheb:
You want a new file?

neheb:
I did just that. Not sure about the copyright/license of the new file. I just copied from a random file in the same directory.

src/mixer/plugins/volume_mapping.c
```diff
@@ -111,9 +106,9 @@ static double get_normalized_volume(snd_mixer_elem_t *elem,
 	if (use_linear_dB_scale(min, max))
 		return (value - min) / (double)(max - min);
 
-	normalized = exp10((value - max) / 6000.0);
+	normalized = pow(10, (value - max) / 6000.0);
```
MaxKellermann: Why that?

neheb: There was a patch like this for alsa-utils that was accepted upstream. I figure it’s the same.

MaxKellermann: That doesn't answer my question.

neheb: exp10 is a GNU extension, not part of C++, and not available everywhere.

neheb: Reworded commit message.

neheb:
Added the lround commit here.

neheb:
Rebased. Passes Travis.

neheb:
A small note: the std:: math functions under gcc are constexpr. No idea if it makes a meaningful difference.

src/mixer/plugins/volume_mapping.c
neheb: Did so again.

neheb:
Rebased. Anything else missing?

PR #724, opened 2020-01-31
MadDecoderPlugin: fix bad printf format

neheb:
max_frames is size_t, not unsigned long. Fixes GCC warning.


neheb:
Btw GCC cannot check most usages of formatted strings in the codebase. Most of the functions are using some weird std::forward thing with Args&&.

Try compiling with -Wformat=2 to see what I mean.

MaxKellermann:
Cherry-picked to v0.21.x 80a0cf694f152f94488b9c5cf2bd35518f149f73

PR #729, opened 2020-02-02
[clang-tidy] replace std::bind with lambdas

neheb:
Found with modernize-avoid-bind


MaxKellermann:
I don't get it, what's the point? The new code is much longer and more complicated.

neheb:
Funny enough, the clang-tidy page says:

std::bind can be hard to read and can result in larger object files and binaries due to type information that will not be produced by equivalent lambdas.

PR #737, opened 2020-02-02
[clang-tidy] remove needless std::move

neheb:
As a result of the last commit.

Found with performance-move-const-arg

### DIFF
--- a/src/config/File.cxx
+++ b/src/config/File.cxx
@@ -72,7 +72,7 @@ config_read_name_value(ConfigBlock &block, char *input, unsigned line)
 		throw FormatRuntimeError("\"%s\" is duplicate, first defined on line %i",
 					 name, bp->line);
 
-	block.AddBlockParam(name, std::move(value), line);
+	block.AddBlockParam(name, value, line);
 }
 
 static ConfigBlock
--- a/src/storage/CompositeStorage.cxx
+++ b/src/storage/CompositeStorage.cxx
@@ -118,7 +118,7 @@ CompositeStorage::Directory::Make(const char *uri)
 	Directory *directory = this;
 	while (*uri != 0) {
 		const std::string name = NextSegment(uri);
-		auto i = directory->children.emplace(std::move(name),
+		auto i = directory->children.emplace(name,
 						     Directory());
 		directory = &i.first->second;
 	}
--- a/src/storage/plugins/UdisksStorage.cxx
+++ b/src/storage/plugins/UdisksStorage.cxx
@@ -168,7 +168,7 @@ UdisksStorage::OnListReply(ODBus::Message reply) noexcept
 			if (!o.IsId(id))
 				return;
 
-			dbus_path = std::move(o.path);
+			dbus_path = o.path;
 			mount_point = std::move(o.mount_point);
 		});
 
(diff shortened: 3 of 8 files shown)


src/storage/plugins/UdisksStorage.cxx
```diff
@@ -168,7 +168,7 @@ UdisksStorage::OnListReply(ODBus::Message reply) noexcept
 			if (!o.IsId(id))
 				return;
 
-			dbus_path = std::move(o.path);
+			dbus_path = o.path;
```
MaxKellermann: This change is bad. You have let clang-tidy do its job without understanding the real problem.

src/storage/CompositeStorage.cxx
```diff
@@ -118,7 +118,7 @@ CompositeStorage::Directory::Make(const char *uri)
 	Directory *directory = this;
 	while (*uri != 0) {
 		const std::string name = NextSegment(uri);
-		auto i = directory->children.emplace(std::move(name),
+		auto i = directory->children.emplace(name,
```
MaxKellermann: Same here. The problem is the `const`, not the `std::move()`.

src/config/File.cxx
```diff
@@ -72,7 +72,7 @@ config_read_name_value(ConfigBlock &block, char *input, unsigned line)
 		throw FormatRuntimeError("\"%s\" is duplicate, first defined on line %i",
 					 name, bp->line);
 
-	block.AddBlockParam(name, std::move(value), line);
+	block.AddBlockParam(name, value, line);
```
MaxKellermann: Why remote this `std::move()`?

neheb: warning: std::move of the variable 'value' of the trivially-copyable type 'char *' has no effect; remove std::move() [hicpp-move-const-arg]

src/storage/CompositeStorage.cxx
neheb: Warning: std::move of the const variable 'name' has no effect; remove std::move() or make the variable non-const [hicpp-move-const-arg]

I guess I  can remove the const.

src/storage/plugins/UdisksStorage.cxx
neheb: warning: std::move of the const expression has no effect; remove std::move() [hicpp-move-const-arg]

src/storage/CompositeStorage.cxx
neheb: Fixed.

neheb:
clang-tidy claims to remove std::move under three conditions:

if std::move() is called with a constant argument,
if std::move() is called with an argument of a trivially-copyable type,
if the result of std::move() is passed as a const reference argument.

All are shown when running the tool.

src/storage/plugins/UdisksStorage.cxx
MaxKellermann: The warning is correct, but not the solution. That's what I mean: you use clang-tidy blindly, without understanding why the problem is there, and let clang-tidy choose a solution which isn't good here.

MaxKellermann:
ac50bb5d2bbe1dad81eac6a6fd84d1f0d25eccab

I removed several wrong changes.

PR #738, opened 2020-02-02
[clang-tidy] use auto where appropriate

neheb:
Found with modernize-use-auto

My understanding is the following:

auto c = (const char *) -> lhs is const char *
const auto *c = (const char *) -> lhs is const char *
const auto c = (const char *) -> lhs is const char * const

Removing the * makes the type clearer IMO.

### DIFF
--- a/src/IcyMetaDataParser.cxx
+++ b/src/IcyMetaDataParser.cxx
@@ -194,7 +194,7 @@ icy_parse_tag(
 size_t
 IcyMetaDataParser::Meta(const void *data, size_t length) noexcept
 {
-	const auto *p = (const unsigned char *)data;
+	auto p = (const unsigned char *)data;
 
 	assert(IsDefined());
 	assert(data_rest == 0);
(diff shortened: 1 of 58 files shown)


MaxKellermann:
Build breakage:
```
../../src/util/UriUtil.cxx: In function ‘std::__cxx11::string uri_remove_auth(const char*)’:
../../src/util/UriUtil.cxx:93:16: error: invalid conversion from ‘const char*’ to ‘char*’ [-fpermissive]
   slash = auth + strlen(auth);
           ~~~~~^~~~~~~~~~~~~~
```

neheb:
weird. I don't get that locally. Fixed anyways.

neheb:
Oh I see the problem. strchr returns char *. std::strchr returns const char *.

MaxKellermann:
```
../../src/decoder/plugins/AudiofileDecoderPlugin.cxx: In function ‘AFvirtualfile* setup_virtual_fops(AudioFileInputStream&)’:
../../src/decoder/plugins/AudiofileDecoderPlugin.cxx:132:44: error: use of ‘vf’ before deduction of ‘auto’
  auto vf = (AFvirtualfile *)malloc(sizeof(*vf));
                                            ^~
```
Do you compile before you submit a PR?

neheb:
I remember encountering this error. I thought I fixed it. Maybe I did so in a different branch. I'll update this when I can.

neheb:
Fixed and rebased. Travis passes.

MaxKellermann:
I don't think any of this is really an improvement.

PR #739, opened 2020-02-03
revert to {} to work around gcc bug

neheb:

### DIFF
--- a/src/storage/CompositeStorage.cxx
+++ b/src/storage/CompositeStorage.cxx
@@ -179,7 +179,10 @@ CompositeStorage::Directory::MapToRelativeUTF8(std::string &buffer,
 	return false;
 }
 
-CompositeStorage::CompositeStorage() noexcept = default;
+CompositeStorage::CompositeStorage() noexcept
+{
+}
+
 CompositeStorage::~CompositeStorage() = default;
 
 Storage *


MaxKellermann:
Bad commit message. Which bug?

neheb:
Travis is failing on it actually. With GCC 6.

neheb:
Updated.

MaxKellermann:
But is this really a bug? Did you get any confirmation from the GCC project?
I have a feeling that the compiler error is correct, so I'm curious for your analysis.

neheb:
It's most definitely a bug, either with older GCC or older libstdc++ since it does not happen with newer versions. I think it compiles with clang + libc++ as well.

MaxKellermann:
Do you understand the error message? Do you know why the compiler complains?
The complaint is based on the C++ standard.
Differences between compiler versions are legal/possible, because some compilers (and their standard libraries) may choose to make methods/constructors `noexcept` even if the standard does not mandate it. And maybe a newer libstdc++ version has made certain constructors `noexcept`, which are called implicitly here. Did you consider this?

neheb:
Yeah I've seen that with the math functions. They're constexpr in GCC even though the standard doesn't mention anything.

All I know is that it's failing on the buildbots currently.

PR #746, opened 2020-02-04
[clang-tidy] add const to parameter

neheb:
Found with readability-non-const-parameter


MaxKellermann:
Why? Having it `const` implies that there is no relation to the first parameter, but it is really the end of the range. Even if nobody actually writes to it, it's a semantic indication.

neheb:
Hrm OK

PR #749, opened 2020-02-04
[clang-tidy] fix mismatching declarations

neheb:
Found with readability-inconsistent-declaration-parameter-name


MaxKellermann:
I don't like that. The paramter names did differ, but sometimes the one in the implementation was worse, but your tool doesn't even try to decide which one is better, and you didn't review it either.

neheb:
Sure, the tool modifies the headers as that's the least error prone way of making this change.

MaxKellermann:
True, but this is not about how to be most lazy about writing a tool. This isn't even about the tool, it's only about improving MPD.

PR #751, opened 2020-02-04
[clang-tidy] remove misplaced const

neheb:
The rhs evaluates to void *const, not const void*.

Found with misc-misplaced-const


MaxKellermann:
Which `const` was misplaced? Those you removed looked useful to me, so why do you want to remove them?

neheb:
The pointer itself is const but is not of type const void *. The const can be added before the auto. I typically don't like doing so as the compiler is usually smarted about it. That is, it tends to optimize better without const.

PR #753, opened 2020-02-05
[clang-tidy] fix incorrect rounding

neheb:
Found with bugprone-incorrect-roundings


MaxKellermann:
That should be `std::lround()`. Or some compatibility wrapper for broken build environment like yours.

neheb:
Noted. Although it would probably make sense for that commit to be merged first.

neheb:
Actually just moved to the other PR instead.

PR #755, opened 2020-02-05
[clang-tidy] remove duplicate branches

neheb:
Found with bugprone-branch-clone


MaxKellermann:
Some of the changes may be acceptable, but some make the code fragile. I'd like to keep some of those `case`s separate, so it's obvious what it does. Imagine somebody some day adds code to the last `case`, and it'll be hard to see what is wrong.
And I'd like to preserve the order on `switch(enum)`, for readability.
The existing code is not "bugprone". The patched code is.

neheb:
Noted.

PR #764, opened 2020-02-17
[clang-tidy] use dynamic_cast for derived classes

neheb:
Found with cppcoreguidelines-pro-type-static-cast-downcast


MaxKellermann:
No, this is wrong. Use your brain, you're not a stupid monkey pressing the `clang-tidy` button, and so you shouldn't behave like one!
If this were necessary, it would crash.

PR #768, opened 2020-02-18
gme: check for empty metadata strings instead of nullptr

jprjr:
Using libgme 0.6.2 on macOS, it appears that `gme_info_t` strings can be empty, which creates weird track titles: ` (001/050)`

This adds an additional check for an empty string.

### DIFF
--- a/src/decoder/plugins/GmeDecoderPlugin.cxx
+++ b/src/decoder/plugins/GmeDecoderPlugin.cxx
@@ -223,7 +223,7 @@ ScanGmeInfo(const gme_info_t &info, unsigned song_num, int track_count,
 	if (track_count > 1)
 		handler.OnTag(TAG_TRACK, StringFormat<16>("%u", song_num + 1).c_str());
 
-	if (info.song != nullptr) {
+	if (info.song[0] != '\0') {
 		if (track_count > 1) {
 			/* start numbering subtunes from 1 */
 			const auto tag_title =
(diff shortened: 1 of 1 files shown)


MaxKellermann:
This rather sounds like a libgme bug. Did you consult that library project?

jprjr:
It looks like it's intentional - libgme is using [static buffers for the string data:](https://bitbucket.org/mpyne/game-music-emu/src/3408a3365bcd0318888e470482fefd7b36b7e49b/gme/Gme_File.h#lines-36), with comments that it's an empty string if not available.

`gme_track_info` just updates the pointers to these internal static buffers.

In fact, re-reading the public c-interface header, it states "defaults to empty string if not available" - so the `nullptr` check can be removed. I'll update, squash, and push.

jprjr:
Done - this replaces the nullptr checks entirely and checks for empty strings instead.

Verified this works with libgme 0.6.2 as well as the latest code from the `master` branch.

MaxKellermann:
> It looks like it's intentional - libgme is using [static buffers for the string data:](https://bitbucket.org/mpyne/game-music-emu/src/3408a3365bcd0318888e470482fefd7b36b7e49b/gme/Gme_File.h#lines-36), with comments that it's an empty string if not available.

Oh, that's an API change! The libgme version I have has `const char*` pointers.

MaxKellermann:
Are you sure your code still works with older libgme versions? It may dereference the null pointer... do all libgme versions guarantee that the pointer is never null?

jprjr:
I think it should - [Game_Music_Emu 0.5.2](http://www.slack.net/~ant/libs/audio.html#Game_Music_Emu) actually declares them as `char[256]` in the public `gme.h` header.

Version 0.5.5, 0.6.2, and the latest master all declare them as `const char*` in the public `gme.h`, but `Gme_File.h` has a separate `track_info_t` type with`char[256]`.

`gme.cpp` winds up allocating the `track_info_t`, then updates the pointers in the `gme_info_t` to the `track_info_t`'s `char[256]` entries.

I just wrote a quick metadata-dumping program, and it looks like with 0.5.5, 0.6.2, and the latest git, the strings are empty, not null.

MaxKellermann:
Cherry-picked to v0.21.x 976372ff6334db52e2e2beef24794cd4549de592
.. and copied text from this PR to the commit message. If you add text to a PR, it means your commit message is bad!

PR #769, opened 2020-02-20
[clang-tidy] do not use else after return

neheb:
Found with readability-else-after-return


MaxKellermann:
I don't see how this improves anything.

Issue #758, opened 2020-02-09
Specifying relative paths in playlist (m3u,xspf) doesn't work


smithjd15:
This seems to be the same issue as report here: https://github.com/MusicPlayerDaemon/MPD/issues/200

### Describe the bug
Playlists opened from files with relative paths to the playlist tracks don't work.

## Expected Behavior
MPD should resolve these playlists relative to the path the playlists are in, and playback without error.

## Actual Behavior
I have a daemon acting as a master for a remote daemon with smb and the proxy db. The error for trying to play back a track from a playlist loaded from a file in M.A.L.P. for both differ. mpc won't load any tracks into the playlist (via mpc load playlists/xxxx.xspf) on either daemon. Further, I get a "Bad song index" error when trying to play back the entire xspf playlist file from M.A.L.P.. This error occurs on both daemons.

#### Master daemon:
Music Player Daemon 0.21.5 (0.21.5)

Copyright 2003-2007 Warren Dukes <warren.dukes@gmail.com>
Copyright 2008-2018 Max Kellermann <max.kellermann@gmail.com>
This is free software; see the source for copying conditions.  There is NO
warranty; not even MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Database plugins:
 simple proxy upnp

Storage plugins:
 local smbclient udisks nfs curl

Neighbor plugins:
 smbclient upnp udisks

Decoders plugins:
 [mad] mp3 mp2
 [mpg123] mp3
 [vorbis] ogg oga
 [oggflac] ogg oga
 [flac] flac
 [opus] opus ogg oga
 [sndfile] wav aiff aif au snd paf iff svx sf voc w64 pvf xi htk caf sd2
 [audiofile] wav au aiff aif
 [dsdiff] dff
 [dsf] dsf
 [hybrid_dsd] m4a
 [faad] aac
 [mpcdec] mpc
 [wavpack] wv
 [modplug] 669 amf ams dbm dfm dsm far it med mdl mod mtm mt2 okt s3m stm ult umx xm
 [mikmod] amf dsm far gdm imf it med mod mtm s3m stm stx ult uni xm
 [sidplay] sid mus str prg P00
 [wildmidi] mid
 [fluidsynth] mid
 [adplug] amd d00 hsc laa rad raw sa2
[...]

Issue #200, opened 2018-01-18
m3u relative path of music file is relative to music_directory, not the playlist


gamescom15:
Given that you have a structure like this:
```
Music
├─ Genre1
├─Genre2
├─etc...
└─Playlists
```

or like this:
```
.
├─ Playlists
│   ├─ Playlist1
│   └─ Playlist2
└─ Music
    ├ Genre1
    ├ Genre2
    └ etc ...
```

with "Music" as the `music_directory`.
Even though there is no official standard, most music players, for example VLC, implement it so that the relative filepaths in the m3u-playlist should be relative to the playlist. This is the correct way, according to Wikipedia ([source](https://en.wikipedia.org/wiki/M3U)).
However, in MPD the filenames have to be relative to the `music_directory`. This means that you can't use a folder structure like this across multiple different music players and operating systems with MPD.

MaxKellermann:
Hmmm:

    $ mpc -q clear; mpc load bugs/200/foo.m3u
    loading: bugs/200/foo.m3u
    $ mpc playlist -f '%file%'
    bugs/200/bar.mp3
    $ echo listplaylistinfo bugs/200/foo.m3u |nc localhost 6601
    OK MPD 0.20.0
    file: bugs/200/bar.mp3
    Last-Modified: 2018-01-18T20:48:33Z
    Artist: Deftones
    Title: Feiticeira
    Album: White Pony
    Track: 1
    Time: 190
    duration: 189.944
    OK
    ^C
    $ cat ~/mpd/music/bugs/200/foo.m3u
    bar.mp3

This very much looks like MPD already does what you want it to do, doesn't it?

gamescom15:
I meant if the file is not in the same folder as the playlist, sorry i wasn't specific enough.
Folder structure:
```
Music
├── playlists
│   ├── to_musicdir.m3u
│   └── to_playlist.m3u
└── songs
    └── bar.mp3
```


```
sudo egrep '^music_directory|^playlist_directory' /etc/mpd.conf
music_directory			"~/Music"
playlist_directory		"~/Music/playlists"
```

```
$ mpc -q clear; mpc load to_musicdir
loading: to_musicdir
l$ mpc playlist -f '%file%'
songs/bar.mp3
$ mpc -q clear; mpc load to_playlist
loading: to_playlist
$ mpc playlist -f '%file%'
$ tail playlists/*
==> playlists/to_musicdir.m3u <==
songs/bar.mp3

[...]

PR #772, opened 2020-02-26
Fix loading of playlists with song uri's containing relative path elements

smithjd15:
Addresses issues https://github.com/MusicPlayerDaemon/MPD/issues/760, https://github.com/MusicPlayerDaemon/MPD/issues/758, https://github.com/MusicPlayerDaemon/MPD/issues/200


MaxKellermann:
If a PR contains text, it means that your commit message is bad. And indeed, it is bad, because it doesn't explain anything.
This code looks fishy. I don't understand why it does that and why it fixes the problem.
And I don't like the excessive use of `std::string`.

smithjd15:
Basically, all uris in the database should be relative to the mpd music dir. That's not the problem, the problem is that relative path elements won't be in the database, so they have to be removed from the lookup uri or the lookup fails because the path isn't found. So the patch removes a number of relative path elements and leaves the path relative to the mpd music dir, essentially dropping the playlist part of the path (and how to get to the track file from it).

So for example 'playlists/xxx/../../music/track.mp3' becomes 'music/track.mp3', allowing the track to be found during the add.

MaxKellermann:
> Basically, all uris in the database should be relative to the mpd music dir. That's not the problem, the problem is that relative path elements won't be in the database

What is a "relative path element"? And why is it not in the database?

> So for example 'playlists/xxx/../../music/track.mp3' becomes 'music/track.mp3', allowing the track to be found during the add.

So this implements path normalization, but it is a *very* lousy implementation. It is slow, bloated and buggy. For example:
```
"playlists/abc/def/ghi/../track.mp3" -> "track.mp3"
"playlists/abc/def/ghi/./track.mp3" -> "track.mp3"
"abc/def./track.mp3" -> "track.mp3"
```
All of that is wrong.

smithjd15:
> > Basically, all uris in the database should be relative to the mpd music dir. That's not the problem, the problem is that relative path elements won't be in the database
> 
> What is a "relative path element"? And why is it not in the database?

../ is not in the file path probably because the directory indexer doesn't add found tracks relative to a directory other than the music directory.

> 
> > So for example 'playlists/xxx/../../music/track.mp3' becomes 'music/track.mp3', allowing the track to be found during the add.
> 
> So this implements path normalization, but it is a _very_ lousy implementation. It is slow, bloated and buggy. For example:
> 
> 
> "playlists/abc/def/ghi/../track.mp3" -> "track.mp3"
> "playlists/abc/def/ghi/./track.mp3" -> "track.mp3"
> "abc/def./track.mp3" -> "track.mp3"
> 
> 
> All of that is wrong.

It's very primitive, yes. I'm working out a more robust version.

smithjd15:
I created a helper function in UriRelative to normalize relative paths, and now normalize each uri for LightSong creation, resolving the playlist relative path problem.

There is another unrelated bug that I've run into, this one in the Xspf playlist parser, that chops the front part of the retrieved path off. I doesn't seem to do this in any logical way. Some tracks (an obscure minority) are affected, most aren't. It doesn't seem to be a result of special characters in the path. If the path is modified with a further path element at the start of the uri, the path is parsed correctly.

MaxKellermann:
Your new code still uses `std::string` excessively. It allocates memory and creates copies of the string just to compare it! And the API requires the parameter to be `std::string`, creating a temporary copy of the parameter.

And I still don't understand the problem this is trying to solve. File names inside playlists are relative to the playlist file. If this file doesn't exist, why does it suddenly exist when you normalize the path?

smithjd15:
> Your new code still uses `std::string` excessively. It allocates memory and creates copies of the string just to compare it! And the API requires the parameter to be `std::string`, creating a temporary copy of the parameter.
> 
The normalization requires several abilities of std::string, so the input should be a std::string. And returning a std::string fits very well within UriRelative. The token string could not be a char without breaking the output.

> And I still don't understand the problem this is trying to solve. File names inside playlists are relative to the playlist file. If this file doesn't exist, why does it suddenly exist when you normalize the path?

Files are looked up by path. Looking up "../xxx/track.mp3" isn't the same as looking up "track.mp3" and currently the path used for lookup is not normalized, so while the track uri "track.mp3" may exist in the database, the uri "../xxx/track.mp3" from the playlist will not match it. The aim is firstly to make otherwise valid playlists work correctly in mpd and secondly to enable mpd to play existing playlists that correctly work across a large number of other players without modification.

smithjd15:
I've gotten around the temporary string usage by creating a char strictly for compare, and applying the substring equivalent to the result string instead of appending the char.

smithjd15:
Re-written using the SplitString utility class. It seems to be as stable, and is much simpler. The string issues are now resolved.

smithjd15:
- Rename, move normalize function to util/UriUtil.
- Normalize the path in playlist/PlaylistSong, closer to the path's origin.

smithjd15:
Revisit nomenclature, adapt to string_view.

xeruf:
@MaxKellermann if it is so lousy, why don't you help address it? Supporting relative paths in playlists is a basic feature and I do not understand how the highly touted unix-philosophy-following `mpd` can refuse to support it for years...

MaxKellermann:
> @MaxKellermann if it is so lousy, why don't you help address it?

What a useless question to a volunteer working for you on MPD in his spare time.
I could ask: why don't **you**?

> Supporting relative paths in playlists is a basic feature and I do not understand how the highly touted unix-philosophy-following `mpd` can refuse to support it for years...

MPD does support it.

xeruf:
I am considering to help, but my lack of skills in C++ would result in quite slow progress, especially since I have many other projects in consideration. Is there any documentation on playlist support in MPD? The manual is rather sparse about it.

MaxKellermann:
Your help is very much welcome, but the discussion is off-topic here. Let's discuss this on IRC (`#mpd` on Liberachat) or in a GitHub discussion (see https://github.com/MusicPlayerDaemon/MPD/discussions).
So far, I don't know what issue you want to help with; the issues linked above in this PR are all closed. There is no open issue about relative path support in MPD playlists.

PR #777, opened 2020-03-04
new command "tagstats"

geneticdrift:
Added command `tagstats` to get counts of all enabled tag types, instead of only albums, artists and songs that `stats` returns.


MaxKellermann:
This is a very heavy command. What use is it?

geneticdrift:
Let the client know which tags are actually used without listing each one and counting.
The client can adjust the UI and the tagtypes mask accordingly.
I didn't measure how heavy it is but from testing with large db it didn't fill different from the normal stats. The results are cached anyway so it's once per db update.

MaxKellermann:
I don't understand your explanation.
What kind of UI changes would a client want to do based on the number of items in certain tags?
What kind of tag mask adjustments would a client want to do based on the number of items in certain tags?
What's the story?

> Let the client know which tags are actually used

That means the client doesn't need to know the numbers.

geneticdrift:
UI changes for example:
1. Not showing menu item for `Conductors` if there are zero `Conductors` tag values.
2. Not showing option to group albums by `OriginalDate` if the are are zero `OriginalDate` tag values.
3. In a navigation menu for tag types, show only the tags that are relevant.

Essentially like `tagtypes` tell the client which tags are enabled, `tagstats` tell the client which tags are actually used in the music collection.

The more information and stats the client has about the music collection the more features and information it can offer to the user.

I think having MPD offer these stats to all clients is preferable to having every client implement them on its own, and it makes all clients and thus MPD more valuable.

There are all kinds of use cases. Some users don't care about more than album and artist, others care about performer, conductor, work, and label too.

MaxKellermann:
OK, so it is useful for a client to know which tags are present in the database. I understand.
But you did not explain why the numbers are necessary.

geneticdrift:
> But you did not explain why the numbers are necessary.
Not necessary, but nice to have. 

If as a design principal you prefer to leave it to clients to handle stats other than the minimum necessary then I understand. It's just a proposal I thought was beneficial but you have the full perspective of the project. It's your call.

If the concern is just for performance, maybe it would be better to process stats during the db update process and save them in the db file (it could become a place for other stats in the future such as tag clouds).
Or instead of tag stats for all tags at once, a command to count individual tags e.g. `tag_count performers`. That shouldn't be worse than other commands that do `CollectUniqueTags`.

MaxKellermann:
> Not necessary, but nice to have.

Nice to have for what reason?

Without the numbers, the whole process of collecting those tags needs just a bit mask (`TagMask`) and nothing else. And it's stored in that same bit mask.

With numbers, you need to copy all unique tag values of all tags and store them in a hash table. That kind of complexity and overhead needs to be justified with more than "it's nice".

For some tag types, it doesn't even make any sense to count the unique values - for example, the MusicBrainz ID, or the Date tag.

geneticdrift:
Understood.
Thank you for taking the time to review the PR.

Issue #220, opened 2018-02-12
[Feature request] selective choosing frequency for resampling


mistepien:
It would be great to resample taking into account multiplication of 44100 and 48000.
Eg. sound card supports 44100 48000 88200 96000.
If input file is higher than 96000 than it will resampled to 96000, notwithstanding it is 192000 or 352800.

Resampling from 352800 to 88200 is noticeably faster than from 352800 to 96000.
It matters in case of slower machines. Thus is suggest solution according to which eg. 352800 is resampled to 44100, 88200 or 176400 (depending what is supported by sound card) and 384000 is resampled to 48000, 96000 or 192000.

RafaPolit:
I second the request.  I would have hopped to get SoX kind of 'built into' this feature and have the following behaviour:

We currently get (in SoX) 16 bit / *, 24 bit / * and 32 bit / *, which is nice to zero pad a signal and keep the native sample rate.

Ideally, upsampling or downsampling is better within same clocks: 44.1 and 48.  So, if I am upsampling a 44.1 file to the 3xx kHz, it is better served with a 352.8 frequency.  If I am doing the same for a 48 file, 384 makes more sense.

Could we get 'intelligent clock-wise' modes like:
16 bit : 44.1 / 48
24 bit: 176.4 / 192
32 bit: 352.8 / 384

I read somewhere that this was, precisely, the way SoX handled these scenarios, but I have done some testing in MPD and there is no way to achieve this (with SoX). 384 is 384 no matter the source clock.

Can this be worked in? Thanks,
Rafa.

alexbo1:
I would be interested in this feature too. 
Ideally it should be possible to configure resampling for each combination of input and output sampling rate/bits, but I would also be very happy with a simple config option to deactivate sampling rate conversion for DSD material.

At the moment everything is converted to the specified sampling rate/bit resolution, if resampling is activated. For PCM this is generally ok, aside from the abovementioned performance issues.
The conversion of DSD to PCM (and vice versa) is more problematic and definitely should not be done if a DAC supports DSD natively, such as my RME ADI2 Pro Fs.
I would be able  to playback DSD data in its native form, but have 44.1/16 and other low resolution PCM data resampled to >= 192/32, which definitely sounds better - maybe due to possible higher filter cutoff frequencies ...

At the moment I have to reconfigure (turn on/off resampling) und restart mpd to achieve the desired behavior for both DSD and PCM data.

[...]

PR #782, opened 2020-03-10
support selective resample based on 44k and 48k clocks

bitkeeper:
Implements #220


MaxKellermann:
" For example if the \fBaudio_output\fP is set to 96k, a 44.1 song will be resampled to 88.2."

I don't understand that description. How does one set it to 96k, and why 88.2?

I think this should be a per-output setting.

New documentation should be added to the sphinx files, not to the manpage.

bitkeeper:
@mistepien in  #220 explains it in the top message.

Lets says that the audio_output_format is set "96000:24:2" for upsampling (with Sox).

and the base sample rates:
44.1 based: 44.1 , 88.2, 176.4, 352.8 kHz  
48 base: 48, 96, 192, 384 kHz

If you play a 48k song it is just upsampled with a ratio of 2.0. But if you play a song of 44.1 now the ratio becomes 2.17 and the upsampling is a lot complexer. For example on raspberry the load is increased (almost twice as much). I don't dare to say something about the audio quality, but ratio of 2.0 seems easier to me.

With the new option selective_44k_resample, you can choose if you want to resample to nearest (but lower) 44k1 clock based. A source of 44.1 is then upsampled to 88.2 and source of 88.2 just stays 88.2.

The audio play distributions that I tried, had only a single configuration point for upsampling. Which related to the global setting of the global audio_output_format.

MaxKellermann:
The explanation doesn't belong on GitHub. The commit needs to contain everything the user (and the reviewer) needs to know. Text on GitHub is lost eventually, but MPD should last longer.

MaxKellermann:
> For example on raspberry the load is increased (almost twice as much).

Please add benchmark results to the commit message. "Load" is a bad metric; the best metrics can be logged with `perf stat`. You can run the resampler on the command line with the `run_convert` debug program.

bitkeeper:
To sure; summarize what is missing:
- [ ] Add documentation to rst files and be more clear,
Where in user.rst should I add the documentation ?
- [ ] Add benchmark tp to the documentation.
Pi 2 running Raspbian, so no perf-state. I only have one development pi, so I can no give a broad overview.

mpd % cpu usage during playback:
- 0.7  - 4.0   44k1 played
- 6.7  - 14.0   48k > 96k played
- 15.9 -21.8    44k1 > 96k played

MaxKellermann:
Raspbian has perf. I use it all the time.
(I said add benchmark results **to the commit message**)

btw. `top` is worse than load for measuring the CPU usage. CPU usages from `top` are mostly worthless.

bitkeeper:
No apt-get  linux-perf-4.19 (known issue https://github.com/RPi-Distro/repo/issues/159), but you can get it as manually download.

Can you tell me howto test + options required, perhaps you have a test document for such cases ?

bitkeeper:
Added docu and perf stat results to commit message

doc/mpd.conf.5
```diff
@@ -108,6 +108,11 @@ for the format of this parameter.  Multiple audio_output sections may be
 specified.  If no audio_output section is specified, then MPD will scan for a
 usable audio output.
 .TP
+.B selective_44k_resample <yes or no>
```
MaxKellermann: I'm not so happy with this option's name (and definition). There is no "44k" sample rate; it's "44.1k". And the documentation says it's not even about "44.1k" but about "44.1k or 48k".

doc/mpd.conf.5
```diff
@@ -108,6 +108,11 @@ for the format of this parameter.  Multiple audio_output sections may be
...
+When enabled adhere the 44.1k or 48k clock base if a song requires resampling.
+For example if the \fBaudio_output\fP is set to 96k, a 44.1 song will be resampled to 88.2.
+The default is no.
+.TP
```
MaxKellermann: This is still the old description which is hard to understand. The commit message has explanatory text which is missing here. Users will never read commit messages - those are for developers only.

src/Main.cxx
```diff
@@ -316,11 +316,14 @@ initialize_decoder_and_player(Instance &instance,
...
 	});
 
+	bool selective_44k_resample = config.GetBool(ConfigOption::SELECTIVE_44K_RESAMPLE, false);
+
 	instance.partitions.emplace_back(instance,
 					 "default",
 					 max_length,
 					 buffered_chunks,
 					 configured_audio_format,
+					 selective_44k_resample,
```
MaxKellermann: Why is this a global option, and not a per-output option?

src/Partition.cxx
```diff
@@ -36,6 +36,7 @@ Partition::Partition(Instance &_instance,
 		     unsigned max_length,
 		     unsigned buffer_chunks,
 		     AudioFormat configured_audio_format,
+			 bool selective_44k_resample,
```
MaxKellermann: Bad indent

src/Partition.cxx
```diff
@@ -47,7 +48,9 @@ Partition::Partition(Instance &_instance,
 	 pc(*this, outputs,
 	    instance.input_cache.get(),
 	    buffer_chunks,
-	    configured_audio_format, replay_gain_config)
+	    configured_audio_format, 
+		selective_44k_resample,
```
MaxKellermann: Bad indent

src/decoder/Control.cxx
```diff
@@ -28,11 +28,13 @@
 DecoderControl::DecoderControl(Mutex &_mutex, Cond &_client_cond,
 			       InputCacheManager *_input_cache,
 			       const AudioFormat _configured_audio_format,
+				   const bool _selective_44k_resample,
```
MaxKellermann: Bad indent

src/decoder/Control.cxx
```diff
@@ -51,7 +53,12 @@ DecoderControl::SetReady(const AudioFormat audio_format,
 	assert(audio_format.IsValid());
 
 	in_audio_format = audio_format;
-	out_audio_format = audio_format.WithMask(configured_audio_format);
+
+	if(selective_44k_resample) {
```
MaxKellermann: Bad coding style (missing space after `if`)

src/decoder/Control.hxx
```diff
@@ -121,6 +121,11 @@ private:
 	 */
 	const AudioFormat configured_audio_format;
 
+	/**
+	 * The "selective_44k_resample" setting.
```
MaxKellermann: This is boilerplate text without any content.

src/decoder/Control.hxx
```diff
@@ -179,6 +184,7 @@ public:
 	DecoderControl(Mutex &_mutex, Cond &_client_cond,
 		       InputCacheManager *_input_cache,
 		       const AudioFormat _configured_audio_format,
+			   const bool _selective_44k_resample,
```
MaxKellermann: Bad indent, again

src/pcm/AudioFormat.cxx
```diff
@@ -22,15 +22,46 @@
...
+ * For exmple if the source is 44.1k and the target is 96k it will return 88.2k
+ */
+unsigned
+determine_selective_resample_rate(unsigned source_rate, unsigned target_rate) noexcept;
+
+unsigned
+determine_selective_resample_rate(unsigned source_rate, unsigned target_rate) noexcept
+{
+	unsigned out_sample_rate = source_rate;
+	const std::map<unsigned, unsigned> lut48to41 = {
```
MaxKellermann: This `std::map` gets initialized on each call. This is unnecessary bloat - it's the wrong data structure here. And the name of this variable is cryptic.
Why is this even a hard-coded list?

src/pcm/AudioFormat.cxx
```diff
@@ -22,15 +22,46 @@
 
 #include <assert.h>
 #include <stdio.h>
+#include <map>
+
+/**
+ * If source is based on 44.1k get nearest valid 44.1k variant from the target sample rate.
+ * For exmple if the source is 44.1k and the target is 96k it will return 88.2k
+ */
+unsigned
```
MaxKellermann: Why have this global function's prototype inside the C++ file? Why is this even global?
I don't understand the documentation.

src/pcm/AudioFormat.cxx
```diff
@@ -22,15 +22,46 @@
...
+{
+	unsigned out_sample_rate = source_rate;
+	const std::map<unsigned, unsigned> lut48to41 = {
+		{384000, 352800 },
+		{192000, 176400 },
+	 	 {96000,  88200 },
+		 {48000,  44100 }
+	};	
+
+	if( source_rate % 44100 == 0 && lut48to41.find(target_rate) != lut48to41.end() )
```
MaxKellermann: Why hard-code 44100 here? Where does it consider 48k? Why hard-code anything?

src/player/Control.cxx
```diff
@@ -31,11 +31,13 @@ PlayerControl::PlayerControl(PlayerListener &_listener,
 			     InputCacheManager *_input_cache,
 			     unsigned _buffer_chunks,
 			     AudioFormat _configured_audio_format,
+				 bool _selective_44k_resample,
```
MaxKellermann: Bad indent

src/player/Control.hxx
```diff
@@ -240,6 +245,7 @@ public:
 		      InputCacheManager *_input_cache,
 		      unsigned buffer_chunks,
 		      AudioFormat _configured_audio_format,
+			  bool selective_44k_resample,
```
MaxKellermann: Bad indent

src/Partition.cxx
bitkeeper: Oeps didn't see the difference in the IDE, IDE setting was still tabs instead of spaces. Corrected all.

doc/mpd.conf.5
bitkeeper: Which name would you suggest ?

doc/mpd.conf.5
bitkeeper: Made it the same as the documentation in user.rst

src/decoder/Control.hxx
bitkeeper: Just copying style from the private member above it. I that the "setting" context here meant that is really coming from a value from the setting file (mpd.conf).

src/pcm/AudioFormat.cxx
bitkeeper: Context of each call is if an other stream (like as in another song) is selected.
Should I make it a switch with cases ?

bitkeeper:
Some explaintation:
In `Main.cxx` `initialize_decoder_and_player` the config value is read.

In `control.cxx` `DecoderControl::SetReady` this config value is needed to decide to use it or not (there is the choice mad of the sample rate of the song should be changed or not).

Between `initialize_decoder_and_player` and `DecoderControl::SetReady` are some classes needs to pass the setting to finally DecoderControl. Just following the exact same pattern as with `configured_audio_format`.

Till now all samplerate related code in located in `AudioFormat.cxx`. So I added it here and I agree it isn't clean.

1
Possible solution direction for removing the passing passing along the setting code over multiple classes, would be make it part of AudioFormat. Then the ApplyMask can sort it out.

Only the `initialize_decoder_and_player` and `AudioFormat ` are changed.

2
Concerning the implementation it self I can replace it by switch with cases.

Are prever that I implement 1 and 2 ?

src/pcm/AudioFormat.cxx
MaxKellermann: I don't understand anything in your reply. Conext, what context, and what does that have to do with my review? And how does switch/case solve the issue?

doc/mpd.conf.5
MaxKellermann: The problem is that everything about this is badly named, and it's hard to understand what this is all about. I mean, what is "selective resample" supposed to mean? And how can it be "based on 44k and 48k clocks"? I have no idea!
Once you rephrase the meaning and goal of this feature in a way normal people can understand easily, you can think about how to name the setting.

MaxKellermann:
> Some explaintation:

I don't understand any of your explanations. What does this even explain?

There are fundamental flaws in your concept, and your code works only under certain very narrow circumstances.

For example, there is no feedback from the ALSA device which sample rates are really supported by the DAC. You put this into the MPD core, but the MPD core has no idea of that - that's why I suggested make this a per-output setting.

Let's say you have a 44.1kHz song (ripped from CD) and a 96kHz DAC. Now your code makes MPD convert everything to 88.2kHz. But when your DAC happes to not support 88.2kHz, the next layer of MPD will convert 88.2kHz to 96kHz. Which means you have doubled the CPU usage! With a feature that is supposed to cut the CPU usage in half. And your code is too dumb to notice that.

But let's say your DAC supports 88.2kHz - then most likely it also supports 44.1kHz - are there really DACs which support 96, 88.2 but not 44.1? If it supports 44.1, then what's the point of upsampling to 88.2?

bitkeeper:
Okay, let's start all over again and forget about the current PR.
New explanation and requirements for the implementation below.
If you approve it, I will make a new implementation with a new PR.

People use upsampling on the side of MPD to improve the performance of the DAC filter.
MPD can use Soxr for up and downsampling, which is much better than the default resampler.

The samplerate of most regular used digital audio is based on two different base clocks ranges; 44.1kHz and 48 kHz:
* The 44.1 kHz based range samplerates are 44.1 , 88.2, 176.4 and 352.8 kHz. 
* The 48 kHz based range samplerates are 48, 96, 192 and 384 kHz.

Resampling to a target samplerate that isn't the same base clock range as the source samplerate, is far more complex. With Sox this result in almost twice the needed cpu resources. 

Examples:
* 48 kHz to 96 kHz: ok
* 44.1 kHz to 96 kHz: nok
* 44.1 kHz to 88.2 kHz: ok

The proposed change is about make sure that with resampling you keep yourself in the same clock base:
* you indicate you want to adhere the baseclock (requires a new setting)
* you set the resampling to the maxium allowed/wanted samplerate. In practice this will be expressed in samplerate of the 48kHz clock range, like 192 kHz. In this ways the 44.1 kHz based ones always fit nice in.

Requirements for the implementation:
* The name of the new setting is `resample_rounded_ratio`. Its value is `yes `or `no`. Default `no`.
* The setting is output based and by this should respect the supported samplerates of the output.
* The maxium samplerate can already be specified with the existing output `format` setting.
* Should also work for non standard audio samplerates.

Note:
Audio Distributions like Moode and VolumeIO support the resampling. Only both take a different approach. Moode uses the global setting audio_output_format while VolumeIO is using the output specific setting format.

MaxKellermann:
> People use upsampling on the side of MPD to improve the performance of the DAC filter.

I don't understand this part. How does this improve "performance", and what does "performance" mean in this context?
What objective instrument can be used to measure this "performance" improvement?

bitkeeper:
Changed _People use_ to _People can use_.

Context is the, digital or analog, anti-aliasing filter of the digital analog converter. The performance improved is a less and or different type of filter is possible due higher nyquist frequency. Which like everything with audio is to the listener to like or not. Nothing to do with MPD it self.

MaxKellermann:
You did not answer my questions. I'm not convinced that this is a real thing - this very much sounds like voodoo and bullshit. If it were a real thing, it would be possible for you to explain it and answer my questions.
If you do believe this thing exists, create a new PR and include an explanation.

bitkeeper:
The upsampling is already implemented by MPD. The option I propose is ONLY about if people use up or down sampling to use a more resource friendly form (50% of the current used CPU resources)  which adhere round ration factor for multiply or divide a samplerate.

Sorry I thought that I answered your question, but I wrongly asumed that you know also about audio processing.  [Nyquist ](https://en.wikipedia.org/wiki/Nyquist%E2%80%93Shannon_sampling_theorem) and [anti-aliasing filter](https://en.wikipedia.org/wiki/Anti-aliasing_filter) are very real.

MaxKellermann:
Yes, I know about these things, but your sentence mentioning these didn't explain anything. That sentence didn't make any sense at all. The whole post was incoherent.

PR #784, opened 2020-03-12
replace assert.h with cassert

neheb:
These were deprecated in C++14.

### DIFF
--- a/src/IcyMetaDataParser.cxx
+++ b/src/IcyMetaDataParser.cxx
@@ -23,8 +23,8 @@
 #include "util/StringView.hxx"
 
 #include <algorithm>
+#include <cassert>
 
-#include <assert.h>
 #include <string.h>
 
 #ifdef HAVE_ICU_CONVERTER
--- a/src/IcyMetaDataParser.hxx
+++ b/src/IcyMetaDataParser.hxx
@@ -24,9 +24,8 @@
 #include "tag/Tag.hxx"
 #include "config.h"
 
+#include <cstddef>
 #include <memory>
-
-#include <stddef.h>
 
 class IcyMetaDataParser {
 	size_t data_size = 0, data_rest;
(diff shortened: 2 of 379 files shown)


MaxKellermann:
Does the C++14 standard guarantee that the top-level namespace symbols are made available by the C++ headers?
I made the same complaint already last month: https://github.com/MusicPlayerDaemon/MPD/pull/753#issuecomment-582548250 but you did not reply to that, and instead closed the PR.

neheb:
This seems to be the standard: http://eel.is/c++draft/depr.c.headers#other

Modified quote:

```
Every C header, each of which has a name of the form <name.h>, behaves as if each name placed in the standard library namespace by the corresponding <cname> header is placed within the global namespace scope,
```

In that other PR, I changed it to std::lround. Speaking of which, I should probably change the title...

MaxKellermann:
From this standard text, I read the exact opposite, invalidating your PR!
The text you cited says that the C header (which MPD currently uses without this PR) defines symbols in the global namespace. But it does **not** say that the C++ variants (which MPD uses *with* this PR) populate the global namespace.
Quite the opposite:
> The header <cstdlib> assuredly provides its declarations and definitions within the namespace std.
It may also provide these names within the global namespace

Note the wording: "It may", but that's not a guarantee - that's merely telling the C++ developer that those symbols may already be defined, so they're not available; but it does not legalize actually using those symbols, because they're an implementation detail.

neheb:
Hrm OK I'll partially change the PR.

assert.h and cassert are the same, according to the C++11 and C++17 standards: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0063r3.html (19.3). I'll see about the others.

neheb:
Changed to only replace assert.h.

neheb:
Changed several more.

stdarg, stddef, and stdint place several types in the std:: namespace, but I don't think it makes sense to replace all of them. Changes would be massive:

```
git grep size_t | wc -l
1402
```

MaxKellermann:
Thanks, those commit messages are good work and help readers understand the change!

PR #785, opened 2020-03-12
[clang-tidy] use nodiscard

neheb:
Introduced in C++17. It replaces gcc's warn_unused_result.

Found with modernize-use-nodiscard.

### DIFF
--- a/src/archive/plugins/Bzip2ArchivePlugin.cxx
+++ b/src/archive/plugins/Bzip2ArchivePlugin.cxx
@@ -72,7 +72,7 @@ public:
 	~Bzip2InputStream() override;
 
 	/* virtual methods from InputStream */
-	bool IsEOF() const noexcept override;
+	[[nodiscard]] bool IsEOF() const noexcept override;
 	size_t Read(std::unique_lock<Mutex> &lock,
 		    void *ptr, size_t size) override;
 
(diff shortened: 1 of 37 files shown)


MaxKellermann:
Is there any value in repeating the attribute in overrides?

neheb:
AFAIK, g++ and clang do not inherit nodiscard in overrides.

In other news, it seems g++6 does not like nodiscard. Although if I remember correctly, g++6 doesn't support C++17 fully.

MaxKellermann:
I looked it up - indeed, the C++ standard doesn't say that any function attributes are inherited. This is unfortunate, because doing so would make a lot of sense!

GCC6 is indeed a problem, but I think it would be ok to drop GCC6 support now. I always try to keep MPD compatible with the default compiler in Debian Stable and Ubuntu LTS. Debian Stable has GCC8 (used to be GCC6, this is why MPD requires compatibility with 6), and Ubuntu LTS has GCC 7. Support for `nodiscard` was added in version 7, so this PR is ok.

neheb:
Hrm maybe this was not too correct.

As I see, src/util/Compiler.h has several defines for gcc attributes. Maybe a better solution would have been to replace

```
gcc_warn_unused_result __attribute__((warn_unused_result))
```

with

```
gcc_warn_unused_result [[nodiscard]]
```

But then again that's just pandering to old compilers.

PR #787, opened 2020-03-16
fix missing overrides

neheb:
Found with clang's -Winconsistent-missing-destructor-override


MaxKellermann:
I'd welcome adding this warning flag. I love warning flags!

PR #794, opened 2020-03-18
remove ending commas from enumerator lists

neheb:
Found with -Wpedantic


MaxKellermann:
I don't like that, because it requires changing one existing line when adding a new enum value. Is the trailing comma disallowed?

neheb:
It's only a pedantic warning. This can be closed.

PR #795, opened 2020-03-18
more clang-tidy

neheb:
This was done on Windows with Ubuntu 18.04 with clang-11 and more of MPD being compiled.

### DIFF
--- a/src/fs/io/FileOutputStream.cxx
+++ b/src/fs/io/FileOutputStream.cxx
@@ -270,12 +270,7 @@ FileOutputStream::Commit()
 #endif
 
 	if (!Close()) {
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Making those methods `const` is technically correct, but semantically wrong. Those methods modify pointed-to members, and having the method `const` suggests that nothing is modified.
I don't understand the "false ifdef" commit, please explain.
I've cherry-picked the other 4 commits.

neheb:
```
#ifdef _WIN32
#else
#ifdef _WIN32 <-- will never be true.
#endif
#endif
```

neheb:
Rebased, leaving only one commit.

PR #801, opened 2020-03-22
Add "archive_path" and "delete_after_record" options (feature request 800)

fesc2000:
delete_after_record: Save and discard current song, unless a child plugin instance signals to store it persistently in archive_path when title ends.
Implemented via new Signal() method on output controller.

### DIFF
--- a/src/output/plugins/RecorderOutputPlugin.cxx
+++ b/src/output/plugins/RecorderOutputPlugin.cxx
@@ -29,10 +29,20 @@
 #include "fs/io/FileOutputStream.hxx"
 #include "util/Domain.hxx"
 #include "util/ScopeExit.hxx"
+#include "Partition.hxx"
+#include "Instance.hxx"
+#include "Main.hxx"
+#include "thread/Thread.hxx"
+#include "thread/Name.hxx"
+
+#include "output/MultipleOutputs.hxx"
+#include "output/State.hxx"
+#include "output/Filtered.hxx"
 
 #include <cassert>
 #include <memory>
 #include <stdexcept>
+#include <fstream>
 
 #include <stdlib.h>
 
@@ -55,6 +65,8 @@ class RecorderOutput final : AudioOutput {
 	 * destination path.
 	 */
 	std::string format_path;
+
+	AllocatedPath archive_path = nullptr;
 
 	/**
 	 * The #AudioFormat that is currently active.  This is used
@@ -87,6 +99,7 @@ private:
 
 	size_t Play(const void *chunk, size_t size) override;
 
+	bool Signal (intptr_t info);
 private:
 	[[nodiscard]] gcc_pure
 	bool HasDynamicPath() const noexcept {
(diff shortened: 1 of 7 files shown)


src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -29,10 +29,20 @@
 #include "fs/io/FileOutputStream.hxx"
 #include "util/Domain.hxx"
 #include "util/ScopeExit.hxx"
+#include "Partition.hxx"
+#include "Instance.hxx"
+#include "Main.hxx"
```
MaxKellermann: Layering violation: a plugin shouldn't access MPD's global variables.

src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -29,10 +29,20 @@
...
+#include "thread/Name.hxx"
+
+#include "output/MultipleOutputs.hxx"
+#include "output/State.hxx"
+#include "output/Filtered.hxx"
 
 #include <cassert>
 #include <memory>
 #include <stdexcept>
+#include <fstream>
```
MaxKellermann: Don't use C++ iostreams, they add a *lot* of bloat to the MPD binary.

src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -56,6 +66,8 @@ class RecorderOutput final : AudioOutput {
 	 */
 	std::string format_path;
 
+	AllocatedPath archive_path = nullptr;
```
MaxKellermann: `AllocatedPath` is initialized implicitly, unlike `Path`, which is a dumb pointer.

src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -102,14 +115,47 @@ class RecorderOutput final : AudioOutput {
...
 	void ReopenFormat(AllocatedPath &&new_path);
+
+	void ArchiveTask();
+
+	/* parent (true) recorder to send archive message to */
+	AudioOutputControl *parent;
+	int is_archive_recorder = 0;
+	int archive_requested = 0;
+	int can_archive = 0;
+	int delete_after_record = 0;
```
MaxKellermann: Are these boolean flags?

src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -102,14 +115,47 @@ class RecorderOutput final : AudioOutput {
...
+	std::string archive_format_path;
+	Thread archive_thread;
+	char *archive_source = NULL;
+	char *archive_dest = NULL;
```
MaxKellermann: Who owns the pointed-to memory?

src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -200,9 +268,78 @@ RecorderOutput::Commit()
...
+	}
+
+	if (rc == 0) {
+		FormatDebug(recorder_domain, "%s \"%s\" to \"%s\"",
+			delete_after_record ? "Moved" : "Copied",
+			archive_source, archive_dest);
+	} else {
+		FormatError(recorder_domain, "Failed to %s \"%s\" to \"%s\", rc=%d",
+			delete_after_record ? "move" : "copy",
+			archive_source, archive_dest, errno);
```
MaxKellermann: `errno` isn't portable. Maybe log the exception you caught, instead of using errno?

src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -200,9 +268,78 @@ RecorderOutput::Commit()
...
+	}
+
+
+	free (archive_source);
+	free (archive_dest);
```
MaxKellermann: Don't use plain C memory allocators. They're unsafe and will leak.

src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -200,9 +268,78 @@ RecorderOutput::Commit()
...
+		    unlink(path.c_str());
+	    }
+	}
+
 	delete file;
 }
 
+void
+RecorderOutput::ArchiveTask() {
+	int rc;
```
MaxKellermann: What's the point of this flag? Instead of moving the logger code out of the try/catch, why not log events where they happen?

src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -200,9 +268,78 @@ RecorderOutput::Commit()
...
+
+	SetThreadName("archive_file");
+
+	if (!archive_source || !archive_dest)
+		return;
```
MaxKellermann: Who frees the memory if only one of the two is `nullptr`?

PR #802, opened 2020-03-23
Android: add gdb support

tguillem:
Hello,

I don't know if it is mergeable as it is (cf. WIP) but I wanted to share this way of using gdb on a MPD android build.


MaxKellermann:
I never used gdb on Android, but maybe the script is useful for somebody else - I cherry-picked it to v0.23.x: 4260e788611f0d37542c313b5a271a1f6c9189e4

PR #803, opened 2020-03-23
Android: add AudioManager Mixer

tguillem:
This mixer is the global android one. Setting the volume will change the volume of all applications.
It is also possible to add a SLES mixer.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -311,6 +311,7 @@ if not is_android
 else
   sources += [
     'src/android/Context.cxx',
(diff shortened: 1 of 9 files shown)


MaxKellermann:
Cherry-picked to v0.21.x: 81c16273c543822fa8b87522cb1146a7859edd7a
That's the stable branch, but since all Android features are still experimental, it's ok to mess with the stable branch; this way, users get those improvements faster.

PR #804, opened 2020-03-24
cppcheck fixes + manual stuff

neheb:


neheb:
With this PR, I've reduced the UCLIBC hacks to only the lround function. I don't think it makes sense to try replacing lround with ceil/floor, which both return floats.

You mentioned that the problem is the build environment. The problem is in gcc's libstdc++. The macro there that controls the availability of these functions is _GLIBCXX_USE_C99_MATH_TR1.

I have not gotten LLVM's libc++ to compile with uClibc-ng as it's missing locale_t + some other stuff.

neheb:
Did just that. What a stupid problem.

MaxKellermann:
I dropped the "weird c_str()" patch because I don't understand what's wrong or dangerous about it. If you believe cppcheck has a point, resubmit with explanation.

PR #807, opened 2020-03-25
[0.21.x]treewide: use boost::lround when std::round is unavailable

neheb:
This is the case with uClibc-ng currently.

(cherry picked from commit 769cd0ee9f0cf8ceb026aa751b5d4a390bb5dbdc)


neheb:
Never mind.

PR #809, opened 2020-03-25
Android tv support

tguillem:
I'm using MPD from an Android TV (a Shield) and the MPD UI settings was hidden in system preferences. These commits make the MPD launcher available from TV apps and fix the foreground permission that is needed on recent devices.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -7,6 +7,11 @@
 
   <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="28"/>
 
+  <uses-feature android:name="android.software.leanback"
+                android:required="false" />
+  <uses-feature android:name="android.hardware.touchscreen"
+                android:required="false" />
+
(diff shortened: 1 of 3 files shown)


MaxKellermann:
Why use Java reflection? You can compile directly with this API, and as long as the code is not called, it will still work on older Android versions. Though it may be necessary to move this into a separate class, I'm not sure.

MaxKellermann:
I've cherry-picked this to v0.21.x, and we can refactor this to eliminate reflection later.

tguillem:
> Why use Java reflection? You can compile directly with this API, and as long as the code is not called, it will still work on older Android versions. Though it may be necessary to move this into a separate class, I'm not sure.

Maybe it is possible indeed, but how you handle the "android.app.NotificationChannel" import if missing ? Is that why we need to use a separate class ?

MaxKellermann:
The `import` statement is just syntactic sugar. The JVM resolves those classes only when they're really used, not when the application is loaded. Instead of `import`, you can just use absolute class names where you use the class, without `import` at the beginning of the file.

PR #817, opened 2020-04-01
replace time.h with ctime and switch to std:: variants

neheb:
Made similar std adjustments in edited files.

### DIFF
--- a/src/LogBackend.cxx
+++ b/src/LogBackend.cxx
@@ -24,10 +24,9 @@
 #include "config.h"
 
 #include <cassert>
-
-#include <stdio.h>
-#include <string.h>
-#include <time.h>
+#include <cstdio>
+#include <cstring>
+#include <ctime>
 
 #ifdef HAVE_SYSLOG
 #include <syslog.h>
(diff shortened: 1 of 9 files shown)


src/LogBackend.cxx
```diff
@@ -90,10 +89,10 @@ EnableLogTimestamp() noexcept
 static const char *
 log_date() noexcept
 {
-	static constexpr size_t LOG_DATE_BUF_SIZE = 16;
+	static constexpr std::size_t LOG_DATE_BUF_SIZE = 16;
 	static char buf[LOG_DATE_BUF_SIZE];
-	time_t t = time(nullptr);
-	strftime(buf, LOG_DATE_BUF_SIZE, "%b %d %H:%M : ", localtime(&t));
+	std::time_t t = std::time(nullptr);
+	std::strftime(buf, LOG_DATE_BUF_SIZE, "%b %d %H:%M : ", localtime(&t));
```
MaxKellermann, line 95: This should be `std::localtime`

src/time/Convert.cxx
```diff
@@ -57,7 +54,7 @@ LocalTime(std::chrono::system_clock::time_point tp)
 {
 	const time_t t = std::chrono::system_clock::to_time_t(tp);
 #ifdef _WIN32
-	const struct tm *tm = localtime(&t);
+	const struct tm *tm = std::localtime(&t);
 #else
 	struct tm buffer, *tm = localtime_r(&t, &buffer);
```
MaxKellermann, line 59: Which header provides `localtime_r` now?

src/time/Convert.cxx
```diff
@@ -80,7 +77,7 @@ GetTimeZoneOffset() noexcept
 	struct tm tm;
```
MaxKellermann, line 77: Which header provides `struct tm` now?

neheb: There's std::tm. I'll convert.

src/time/Convert.cxx
neheb: I believe ctime includes time.h. I'll try to find a reference from the standard.

src/LogBackend.cxx
neheb: I'll fix it.

neheb:
On second thought, never mind. After doing some research, it seems there are no good alternatives to localtime and gmtime untill C++20. I'll defer it to when this project moves to that.

PR #819, opened 2020-04-01
[cppcheck] convert several functions to use std::any_of

neheb:
std::any_of becomes constexpr in C++20. I'm not sure it results in better
performance.

Found with useStlAlgorithm

### DIFF
--- a/src/command/FileCommands.cxx
+++ b/src/command/FileCommands.cxx
@@ -127,12 +127,7 @@ gcc_pure
 static bool
 IsValidValue(const StringView s) noexcept
 {
-	for (const char ch : s) {
-		if ((unsigned char)ch < 0x20)
-			return false;
-	}
-
-	return true;
+	return std::none_of(s.begin(), s.end(), [](const auto &ch) { return (unsigned char)ch < 0x20; });
 }
 
 class PrintCommentHandler final : public NullTagHandler {
--- a/src/input/InputPlugin.cxx
+++ b/src/input/InputPlugin.cxx
@@ -33,11 +33,8 @@ InputPlugin::SupportsUri(const char *uri) const noexcept
 			if (StringStartsWithIgnoreCase(uri, *i))
 				return true;
 	} else {
-		for (const auto& schema : protocols()) {
-			if (StringStartsWithIgnoreCase(uri, schema.c_str())){
-				return true;
-			}
-		}
+		return std::any_of(protocols().begin(), protocols().end(), [uri](const auto &schema)
+			{ return StringStartsWithIgnoreCase(uri, schema.c_str()); } );
 	}
 	return false;
 }
--- a/src/lib/nfs/Cancellable.hxx
+++ b/src/lib/nfs/Cancellable.hxx
@@ -113,11 +113,7 @@ public:
 #ifndef NDEBUG
 	gcc_pure
 	bool IsEmpty() const noexcept {
-		for (const auto &c : list)
-			if (!c.IsCancelled())
-				return false;
-
-		return true;
+		return std::all_of(list.begin(), list.end(), [](const auto &c) { return c.IsCancelled(); });
 	}
 #endif
 
--- a/src/output/MultipleOutputs.cxx
+++ b/src/output/MultipleOutputs.cxx
@@ -258,11 +258,8 @@ MultipleOutputs::Open(const AudioFormat audio_format)
 bool
 MultipleOutputs::IsChunkConsumed(const MusicChunk *chunk) const noexcept
 {
-	for (const auto &ao : outputs)
-		if (!ao->LockIsChunkConsumed(*chunk))
-			return false;
-
-	return true;
+	return std::all_of(outputs.begin(), outputs.end(), [chunk](const auto &ao) {
+		return ao->LockIsChunkConsumed(*chunk); });
 }
 
 unsigned
--- a/src/output/MultipleOutputs.hxx
+++ b/src/output/MultipleOutputs.hxx
@@ -28,6 +28,7 @@
 #include "Chrono.hxx"
 #include "util/Compiler.h"
 
+#include <algorithm>
 #include <cassert>
 #include <memory>
 #include <vector>
@@ -106,11 +107,7 @@ public:
 	 */
 	gcc_pure
 	bool IsDummy() const noexcept {
-		for (const auto &i : outputs)
-			if (!i->IsDummy())
-				return false;
-
-		return true;
+		return std::all_of(outputs.begin(), outputs.end(), [](const auto &i) { return i->IsDummy(); });
 	}
 
 	/**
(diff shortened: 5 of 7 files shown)


src/command/FileCommands.cxx
```diff
@@ -127,12 +127,7 @@ gcc_pure
 static bool
 IsValidValue(const StringView s) noexcept
 {
-	for (const char ch : s) {
-		if ((unsigned char)ch < 0x20)
-			return false;
-	}
-
-	return true;
+	return std::any_of(s.begin(), s.end(), [](const auto &ch) { return ch >= 0x20; });
```
MaxKellermann, line 130: This omits the `unsigned char` cast and thus breaks the test.

MaxKellermann: Also, this check is inverted. This checks if there is at least one valid character; but it is really important that *all* characters are valid.

src/lib/nfs/Cancellable.hxx
```diff
@@ -113,11 +113,7 @@ public:
 #ifndef NDEBUG
 	gcc_pure
 	bool IsEmpty() const noexcept {
-		for (const auto &c : list)
-			if (!c.IsCancelled())
-				return false;
-
-		return true;
+		return std::any_of(list.begin(), list.end(), [](const auto &c) { return c.IsCancelled(); });
```
MaxKellermann, line 116: Another inverted check.

src/output/MultipleOutputs.cxx
```diff
@@ -258,11 +258,8 @@ MultipleOutputs::Open(const AudioFormat audio_format)
 bool
 MultipleOutputs::IsChunkConsumed(const MusicChunk *chunk) const noexcept
 {
-	for (const auto &ao : outputs)
-		if (!ao->LockIsChunkConsumed(*chunk))
-			return false;
-
-	return true;
+	return std::any_of(outputs.begin(), outputs.end(), [chunk](const auto &ao) {
+		return ao->LockIsChunkConsumed(*chunk); });
```
MaxKellermann, line 262: Yet another inverted check.

src/output/MultipleOutputs.hxx
```diff
@@ -106,11 +107,7 @@ public:
 	 */
 	gcc_pure
 	bool IsDummy() const noexcept {
-		for (const auto &i : outputs)
-			if (!i->IsDummy())
-				return false;
-
-		return true;
+		return std::any_of(outputs.begin(), outputs.end(), [](const auto &i) { return i->IsDummy(); });
```
MaxKellermann, line 110: Yet another inverted check.

src/song/AndSongFilter.cxx
```diff
@@ -54,9 +56,5 @@ AndSongFilter::ToExpression() const noexcept
 bool
 AndSongFilter::Match(const LightSong &song) const noexcept
 {
-	for (const auto &i : items)
-		if (!i->Match(song))
-			return false;
-
-	return true;
+	return std::any_of(items.begin(), items.end(), [&song](const auto &i) { return i->Match(song); });
```
MaxKellermann, line 59: Yet another inverted check!

src/tag/Builder.cxx
```diff
@@ -154,11 +154,7 @@ TagBuilder::CommitNew() noexcept
 bool
 TagBuilder::HasType(TagType type) const noexcept
 {
-	for (auto i : items)
-		if (i->type == type)
-			return true;
-
-	return false;
+	return std::any_of(items.begin(), items.end(), [type](const auto &i) { return i->type == type; });
```
MaxKellermann, line 157: Hooray, this one looks correct!

neheb:
Not sure I understand these inverted check comments. With the last one, the if condition in the for loop is identical to the any_of lamba. It's also return true else false. The others are return false else true. That's why I inverted them.

src/song/AndSongFilter.cxx
MaxKellermann: Previously, the method returns true when there are exactly zero mismatches. (All items match)
With your code, the method returns true when there is at least one match. (Allows `n-1` mismatches)
See the difference?

A correct implementation could be:
```
return std::all_of(items.begin(), items.end(), [&song](const auto &i) { return i->Match(song); });
```
(i.e. `all_of` instead of `any_of`).

MaxKellermann:
> That's why I inverted them.

You can't just invert a single lambda. I hope the comment I added helps you understand the fallacy.

neheb:
Maybe this will work.

src/command/FileCommands.cxx
```diff
@@ -127,12 +127,7 @@ gcc_pure
 static bool
 IsValidValue(const StringView s) noexcept
 {
-	for (const char ch : s) {
-		if ((unsigned char)ch < 0x20)
-			return false;
-	}
-
-	return true;
+	return std::all_of(s.begin(), s.end(), [](const auto &ch) { return (unsigned char)ch < 0x20; });
```
MaxKellermann, line 130: This is inverted, just in a different way than before....
Previously, it required that zero elements are below `0x20`.
Now it requires that *all* elements are below `0x20`.

neheb:
I think I fixed everything now.

src/input/InputPlugin.cxx
```diff
@@ -32,14 +32,10 @@ InputPlugin::SupportsUri(const char *uri) const noexcept
...
-		for (const auto& schema : protocols()) {
-			if (StringStartsWithIgnoreCase(uri, schema.c_str())){
-				return true;
-			}
-		}
 	}
-	return false;
+
+	return std::any_of(protocols().begin(), protocols().end(), [uri](const auto &schema)
+			{ return StringStartsWithIgnoreCase(uri, schema.c_str()); } );
```
MaxKellermann, line 38: You've moved this out of the `else` branch, why?

neheb: It's basically doing else after return originally. This simplifies it.

MaxKellermann: Before, it never looped over `protocols` if `prefixes!=nullptr`.
With your patch, it always loops over `protocols`, even if `prefixes` has already been checked.
This changes the code's behavior, and I asked "why?" - you may have had reasons to change the code (but it would be a bad idea to do so in a pure refactoring patch) - but you didn't answer my question. I think you weren'd aware of this change.

neheb: Added a return false statement.

neheb:
That was quick.

MaxKellermann:
Too quick. Your `InputPlugin` change caused a crash bug, because it used `begin()` and `end()` of different temporaries.

MaxKellermann:
Regression fixed in 9f8dc31b50aca20db24bcea32fea249b2ac658ba

neheb:
My thinking was that if statement. I figure the return false can just be moved up to avoid the problem.

MaxKellermann:
Avoid what problem?

neheb:
/s/problem/ielse

PR #826, opened 2020-04-14
SolarisOutputPlugin: Various improvements

niacat:
- Use AUDIO_INITINFO rather than AUDIO_GETINFO for initialization
  This is recommended by Solaris and NetBSD documentation.
- Support S8 and S32 audio formats.

### DIFF
--- a/src/output/plugins/SolarisOutputPlugin.cxx
+++ b/src/output/plugins/SolarisOutputPlugin.cxx
@@ -44,6 +44,7 @@
 #define I_FLUSH 0
 #endif
 
+#define AUDIO_INITINFO(v)
 #define AUDIO_GETINFO 0
 #define AUDIO_SETINFO 0
 #define AUDIO_ENCODING_LINEAR 0
@@ -94,6 +95,8 @@ SolarisOutput::Open(AudioFormat &audio_format)
 	struct audio_info info;
 	int ret;
 
+	AUDIO_INITINFO(&info);
+
 	/* support only 16 bit mono/stereo for now; nothing else has
 	   been tested */
 	audio_format.format = SampleFormat::S16;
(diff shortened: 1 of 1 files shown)


MaxKellermann:
I don't like that this PR mixes three completely different things in one commit, which makes it hard to read. Reformatting the code adds a fourth component.

> it hasn't been necessary for a long time.

What is "a long time"? I have no idea who uses MPD on Solaris or NetBSD - which operating system versions does this problem affect, and can we assume that all MPD users have newer versions?

niacat:
At the very least I can confirm that this workaround isn't needed on Solaris 11 (2011), all versions of illumos, and the two most recent supported NetBSD releases (8 and 9).

I honestly have no idea where it's needed, I'm only familiar with this from seeing a similar workaround in SDL (which was removed last year).

Sorry about the format - this is me copying a patch from our packaging system. I'll split it up.

MaxKellermann:
> At the very least I can confirm that this workaround isn't needed on Solaris 11 (2011), all versions of illumos, and the two most recent supported NetBSD releases (8 and 9).

But is this really a workaround? This implements expected and documented behavior:

https://docs.oracle.com/cd/E19253-01/816-5177/6mbbc4g2o/index.html
> if neither the O_NDELAY nor the O_NONBLOCK flag are set, then open() hangs until the device is available

If another process uses `/dev/audio` already, I don't want MPD to block. I want an error condition, immediately.

Has this API changed meanwhile in an incompatible way? Is the documentation above outdated?

niacat:
This is the most recent version of that document (can't find it on Oracle's website) for Solaris 11. I believe your version is from Solaris 10. I've just been paying attention to the illumos docs since they're easy to find. https://docs.huihoo.com/solaris/11.1/english/html/E29044/audio-7i.html

(It mentions that the blocking behaviour is only observed if a single process opens the device twice.)

I'll drop that part of the patch, since I guess Solaris 10 is still technically supported in some form.

MaxKellermann:
> > Has this API changed meanwhile in an incompatible way? Is the documentation above outdated?
> 
> The change in Solaris 11 appears to be that that the blocking behaviour is only observed if a single process opens the device twice, which I don't think is a problem for MPD.

This is not what I understand from the changed wording.
- Solaris 10 allows only one file handle to `/dev/audio`
- Solaris 11 allows multiple file handles to `/dev/audio`, but only one per process, but does not specify what happens when one process opens it twice; blocking however wouldn't make sense, because this would wait forever, because the blocked process can't ever close the first handle, deadlocking itself (of course, it could be closed by another thread, but that would be rather weird)
- Solaris 10 blocks if a second handle is going to be opened, and Solaris 11 blocks if all "channels" are busy (but this has nothing to do with same-process)

PR #828, opened 2020-04-14
v0.21.x: storage/curl: fix href path comparison

tguillem:


MaxKellermann:
Can you please write a better commit message for the second commit? "fix X" is never enough.

tguillem:
> Can you please write a better commit message for the second commit? "fix X" is never enough.

Indeed, I dropped the first commit message/title I wrote unintentionally (vim...).

MaxKellermann:
b18074f8998f7aefc326e09932db37aae4f13062
(Cherry-picked with NEWS line)

Issue #494, opened 2019-02-25
Support for case insensitive list command


ferdkuh:
I think it would be nice to nice to have the ability to have to ability to do case-insensitive queries for tags using the `list `command. Currently this only seems possible when searching for songs using.

After playing around a bit I added a `listi` command that does that, analogous to  `find` / `search`. Is there interest in adding this feature? Should I create a "pull request" (it's really just ~15 lines)?
(Sorry, I am not really familiar with this process)

Best regards
Jonas

MaxKellermann:
Yes, please create a PR.
I don't think a proper implementation is possible with just 15 lines because you'd need to extend the `Database` interface plus implementations for two database plugins, with compile-time and run-time checks in the `proxy` plugin. This isn't trivial, but I welcome your effort. Tell me if you need help.

asktoby:
I'm controlling mpc with voice (via Google Home and IFTTT) and am having trouble with the
`mpc find`
command being case sensitive.

@ferdkuh in your first message you seem to imply that you know a way to do this, but you didn't finish the sentence?

> Currently this only seems possible when searching for songs using.   ...?

What was your method please?

MaxKellermann:
@asktoby your question has nothing to do with this feature request. What you want is `mpc search`.

asktoby:
Thanks, that's great. Sorry for the noise on-thread.

buzuck:
Currently `mpc search` requires the right accentuation (eg `Mànran` is not found when searching for `manran`)

This may be harder to fix, so I will gladly open another issue for this if needed :-)

PR #829, opened 2020-04-15
Client traits to sort lists by case sensitive or insensitive order (M…

geneticdrift:
…usicPlayerDaemon/MPD#494)


MaxKellermann:
There's no explanation and no documentation, I can't review this.
But what smells very bad is the non-constant global variables in ClientTraits.cxx, and the use of `std::map` for runtime-once-initialized global variables (also not const).

geneticdrift:
> no explanation
What do mean?
> no documentation
Do you mean code documentation or protocol documentation? If it's the protocol then it will be added if the PR is accepted.

Anyway, this addresses the feature request that lists be optionally sorted in case-insensitive order instead of the current case sensitive order.

MaxKellermann:
Both code documentation and protocol documentation. Those global variables look obscure to me. And without knowing what kind of protocol extension you suggest to do, it's hard to understand your code.

geneticdrift:
It's simple.

The heart of the change is in `PrintUniqueTags`: it selects to call a cs (case sensitive) or ci (case insensitive) sort order based on a new client property `ClientTraits` lists sort order.

The next significant change is in `RecursiveMap`: instead of one there are now two variants `RecursiveStringMapCS` and `RecursiveStringMapCI`. The difference between them being only the `key_compare` template arg passed to the recursive map.

The rest is the addition of the client traits that holds the lists-sort-type, the `traits` command to list, get, and set traits, and the configuration for the default traits.

I hope this clarifies the commit.
Thanks.

MaxKellermann:
What does the `traits` command do, and what does the word `traits` even mean here?

geneticdrift:
>what does the word traits even mean here

It means a distinguishing property. But I'm not native English speaker so maybe there's a better term.
Basically I just borrowed it from templating usages like `char_traits` and `type_traits`.

>What does the traits command do

The `traits` command lets the client query traits values or change them.
Currently there's only one trait `ListsSortType`.
Example:
<pre>
traits list
ListsSortType: CaseInsensitive
OK
traits set ListsSortType CaseSensitive      
OK
traits get ListsSortType
ListsSortType: CaseSensitive
OK
</pre>

MaxKellermann:
So all this `std::map` stuff is just to have a compile-time constant list of per-client settings? This is way too complicated, complex and too bloated.
And a client needs special privileges to change his own setting. Why that?
Can't we make this a per-command parameter?

geneticdrift:
>Can't we make this a per-command parameter?

Advantages of adding a client trait vs. adding an argument to the `list` command are:

1. Not complicating existing command by more arguments.
2. Trait can affect other commands too, existing or future ones.
3. Compatibility with existing clients that are not aware of the new option: by changing the MPD trait configuration the user can get the desired behaviour without every client having to directly support the new command arg.

>a client needs special privileges to change his own setting

Not sure I understand. Do mean the `PERMISSION_CONTROL` on the command?
I didn't give it too much thought but I don't see a good reason why it can't be `PERMISSION_READ`.

>all this std::map stuff is just to have a compile-time constant list of per-client settings?

The `std::map`s are just a convenience for translating between enum and string.

>This is way too complicated, complex and too bloated.

In what sense is it complicated and bloated? I think it's localized and idiomatic. A one time boilerplate which makes it easy to later add, in a consistent way, additional traits if needed.
If by bloated you mean the template instantiation size, I'm sure it has some size cost but wouldn't most of it be removed by the linker?
Anyway, I think it's a minor implementation detail localized in ` ClientTraits.cxx` and can easily be replaced without any effect outside that file.

MaxKellermann:
What's the point of compatibility with existing client? The response to the `list` command isn't documented to be sorted. See https://www.musicpd.org/doc/html/protocol.html#command-list which says nothing abourt odering. Note that what currently looks like sorted (dumb ASCII sorting) reply is just an internal implementation detail which no client can rely on. Existing clients can never expect any particular order. Existing clients must do client-side sorting.

So making this protocol-level behavior a global configuration option makes no sense at all. This feature is only useful for new clients which are explicitly aware of this feature.

And even for them, I doubt the usefulness of this feature, because there's no advantage of sorting server-side - if the client wants to have it sorted, just do it client-side with the client's locale (different languages have different sorting rules) and with the client's taste. No need for adding this complexity to the server, this is a bottomless pit.

We do already have server-side sorting though, and it is useful: https://www.musicpd.org/doc/html/protocol.html#command-search ... but only for predictable/reliable incremental result downloads using `window`. If it weren't for `window`, then server-side sorting wouldn't be useful for this command either. (What's still missing here is locale-specific sorting.)

> The std::maps are just a convenience for translating between enum and string.

The wrong tool for the job. `std::map` is all about runtime-dynamic mappings, but this here is compile-time constant.

Unfortunately, C++ has no compile-time reflection, so translating between enums and strings is somewhat cumbersome.

> A one time boilerplate which makes it easy to later add, in a consistent way, additional traits if needed.

I have no idea what kind of "traits" you want to add later. But I also have no idea why you want to make sorting a special case of this new pseudo-generic "traits" system. It's pseudo-generic because your way of doing it isn't really generic - every code location really is a special case. Look here:
https://github.com/MusicPlayerDaemon/MPD/blob/a5a368c8e423c7126126cc0e8f1910b657c1981a/src/client/ClientTraits.cxx#L94-L97
Nothing generic here. You pay the bloaty price of being generic by doing a full `std::map` lookup here (even though this `trait_name` is already known at compile time), but you also pay the price for writing this non-generic boilerplate code. Lose-lose.

Everywhere I look, I see no generic code, just lots of `switch/case` with specific code, one more example:
https://github.com/MusicPlayerDaemon/MPD/blob/a5a368c8e423c7126126cc0e8f1910b657c1981a/src/client/ClientTraits.cxx#L147-L153

The whole `Traits.cxx` is a complicated mess. What could be a small 20-line file with some specific code for parsing strings to enum, you made it a bloated pseudo-generic monster with 140 lines of code.

>  If by bloated you mean the template instantiation size, I'm sure it has some size cost but wouldn't most of it be removed by the linker?

No. Your code generates bloated runtime-initialization code and bloated runtime-lookup code. Nothing of that will be optimized away by anybody.

geneticdrift:
> if the client wants to have it sorted, just do it client-side with the client's locale

The client can as well send command to the server for the preferred locale for sort.

> No need for adding this complexity to the server, this is a bottomless pit.
> 
> We do already have server-side sorting though, and it is useful... If it weren't for window, then server-side sorting wouldn't be useful for this command either.

Wouldn't `window` and `sort` be just as useful for `list` command as they are to `find` and `search`?
Doesn't look like one is more complex than the other.
It's not a bottomless pit, just completing the existing pit. It can even be as simple as redefining `RecursiveMap` to use the `IcuCollate` for key comparison.

> What's still missing here is locale-specific sorting

Agree. We can make this a configuration and also a command. Just not call it trait and make a completely new PR addressing the implementation issues.

> there's no advantage of sorting server-side

I disagree. The advantage is consistency across clients and processing close to the database.
Same as `group` is better on the server side. I think `group` added a lot of benefit to clients as opposed to having them each implement its own.
The more data processing clients have to do on the raw data is an encouragement for local caching which, I agree, is best to avoid as much as possible. But clients have constraint too.

Also I'm a bit confused because in this issue https://github.com/MusicPlayerDaemon/MPD/issues/494 you welcomed a PR for `listi` command.

MaxKellermann:
> Wouldn't `window` and `sort` be just as useful for `list` command as they are to `find` and `search`?

Maybe, but so far nobody has requested this feature, and without it, sorting server-side isn't useful.

> > there's no advantage of sorting server-side
> 
> I disagree. The advantage is consistency across clients and processing close to the database.

What kind of consistency? Why close to the database?

> Same as `group` is better on the server side.

No, not the same as all. `group` is server-side because it reduces the amount of data transferred to the client. Sorting doesn't reduce the data transfer.

> Also I'm a bit confused because in this issue #494 you welcomed a PR for `listi` command.

#494 is about "case insensitive list command", and has nothing to do with sorting!

MaxKellermann:
I see no value in server-side `list` sorting, and apparently this PR is based on a misunderstanding that this feature has something to do with #494. Closing.

PR #834, opened 2020-04-20
Update manual pages

mxjeff:
Here are updated versions of mpd.conf.5 and mpd.1 manpages.

The current versions are incomplete and mention old config path (`~/.mpdconf` and `~/.mpd/mpd.conf`). It also mention `log_file` as required, which is no longer exact I believe.

I propose to strip all parameters and refer to HTML doc instead.

Both manpage are built from reStructuredText source.  
I tried to include the rst sources to the build system but I did not manage to have each manual built in its own section (`man1/mpd.`1 and `man5/mpd.conf.5`).

What do you think ?


MaxKellermann:
We should never have generated files in the git repository. But I welcome converting existing raw manpages to RST.

MaxKellermann:
... and when you do convert to RST, the conversion commit should not have text changes. That should be another commit.

mxjeff:
Here are both manpages source in rst format, using rst2man it builds pretty similar manpages as originals.

As i said before I did not manage to integrate manpage build to sphinx. It could instead be integrated to meson as a custom target using rst2man (the tool comes from python3-docutils which in turn is a dependency of sphinx).

MaxKellermann:
> As i said before I did not manage to integrate manpage build to sphinx.

Here's how ncmpc does it: https://github.com/MusicPlayerDaemon/ncmpc/blob/c21af66fa636ac61642c2b434a467aa1ec0ca9be/doc/meson.build#L31-L39

mxjeff:
Indeed Max, this is what I tried to use, but then I cannot have each manual built in its own section (man1/mpd.1 and man5/mpd.conf.5).

With the latest code I pushed, manual are built in `<OUPUT>/doc/man/{mpd.1,mpd.conf.5}` and then both installed in `/usr/local/share/man/man` instead of having `mpd.1` in `/usr/local/share/man/man1` and `mpd.conf.5` in `/usr/local/share/man/man5`.

I've tried several configuration of both meson and sphinx to workaround this, but got no solution so far. I'm no export in meson/ninja, I'm afraid I can't go any further now :|

MaxKellermann:
5f63ffd86cdd27ee98b4451f6445889e1efb3fb5

PR #835, opened 2020-04-21
clang fixes

neheb:


MaxKellermann:
A file called `src/encoder/plugins/.OpusEncoderPlugin.cxx.swp` sneaked in!

MaxKellermann:
What about "Fix -Wunreachable code" - I don't understand the problem and how this commit fixes the problem. What code is unreachable?

MaxKellermann:
"Fix unsafe float comparison." cherry-picked to v0.21.x, a few others cherry-picked to master.

neheb:
Rebased against master.

I have no idea about the unreachable code commit. It seems fairly legit. The variables were moved down to their scope.

MaxKellermann:
But that's still the same scope. Just below a different jump label. The scope is the same, it's the `switch{}`. I don't get it.

neheb:
I haven't seen any codebases (I'm sure they exist) where variables are declared in the switch but outside the case statement. Maybe there's something weird that happens when this is done.

neheb:
Rebased.

src/Permission.cxx
```diff
@@ -70,11 +71,9 @@ static unsigned parsePermissions(const char *string)
...
 	assert(string != nullptr);
 
-	unsigned permission = 0;
-
-	for (const auto i : IterableSplitString(string, PERMISSION_SEPARATOR))
-		if (!i.empty())
-			permission |= ParsePermission(i);
+	auto perm = IterableSplitString(string, PERMISSION_SEPARATOR);
+	unsigned permission = std::accumulate(perm.begin(), perm.end(), 0, [](const auto &pm, const auto &i)
+		{ return pm | ParsePermission(i);});
```
MaxKellermann, line 76: This looks more complicated than before, so what's the point... and you forgot the `empty()` check.

src/tag/Pool.cxx
```diff
@@ -69,12 +70,8 @@ static TagPoolSlot *slots[NUM_SLOTS];
...
 calc_hash(TagType type, StringView p) noexcept
 {
-	unsigned hash = 5381;
-
-	for (auto ch : p)
-		hash = (hash << 5) + hash + ch;
-
-	return hash ^ type;
+	return std::accumulate(p.begin(), p.end(), 5381, [](auto h, auto ch)
+		{ return (h << 5) + ch; }) ^ type;
```
MaxKellermann, line 74: This also looks more complicated than before, and you forgot to add `h`.

neheb:
https://github.com/MusicPlayerDaemon/MPD/pull/878 is probably better.

PR #849, opened 2020-04-29
decoder/sidplay: support new song length format with libsidplayfp 2.0

skidoo23:
High Voltage SID Collection introduces new song length database. PR adds support for new libsidplayfp 2.0 feature while being compatible to older versions of library and format. No need to convert the database any longer :)


MaxKellermann:
This PR has text which is not present in the commit message, so it'll get lost.
Is this a candidate for backporting to v0.21.x?

skidoo23:
No backporting to 0.21 sidplay_global is introduced in 0.22 / git master.  If there is interest I can prepare a PR for 0.21. Maybe useful for Arch Linux users.

PR #853, opened 2020-05-01
use std chr functions

neheb:
The ones in std have overloads for const char/char.

### DIFF
--- a/src/Permission.cxx
+++ b/src/Permission.cxx
@@ -27,11 +27,10 @@
 #include "util/StringView.hxx"
 
 #include <cassert>
+#include <cstring>
 #include <map>
 #include <string>
 #include <utility>
-
-#include <string.h>
 
 static constexpr char PERMISSION_PASSWORD_CHAR = '@';
 static constexpr char PERMISSION_SEPARATOR = ',';
@@ -89,7 +88,7 @@ initPermissions(const ConfigData &config)
 		permission_default = 0;
 
 		param.With([](const char *value){
-			const char *separator = strchr(value,
+			const char *separator = std::strchr(value,
 						       PERMISSION_PASSWORD_CHAR);
 
 			if (separator == nullptr)
--- a/src/PlaylistDatabase.cxx
+++ b/src/PlaylistDatabase.cxx
@@ -25,7 +25,8 @@
 #include "util/StringStrip.hxx"
 #include "util/RuntimeError.hxx"
 
-#include <string.h>
+#include <cstring>
+
 #include <stdlib.h>
 
 void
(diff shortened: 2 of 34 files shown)


src/Permission.cxx
```diff
@@ -89,7 +89,7 @@ initPermissions(const ConfigData &config)
 		permission_default = 0;
 
 		param.With([](const char *value){
-			const char *separator = strchr(value,
+			const char *separator = std::strchr(value,
```
MaxKellermann, line 92: Missing include

src/PlaylistDatabase.cxx
```diff
@@ -25,7 +25,8 @@
 #include "util/StringStrip.hxx"
 #include "util/RuntimeError.hxx"
 
-#include <string.h>
```
MaxKellermann, line 28: With `<string.h>` removed, who provides `strcmp()`?

src/PlaylistFile.cxx
```diff
@@ -81,9 +81,9 @@ spl_valid_name(const char *name_utf8)
 	 * filenames isn't going to happen, either.
 	 */
 
-	return strchr(name_utf8, '/') == nullptr &&
-		strchr(name_utf8, '\n') == nullptr &&
-		strchr(name_utf8, '\r') == nullptr;
+	return std::strchr(name_utf8, '/') == nullptr &&
+		std::strchr(name_utf8, '\n') == nullptr &&
+		std::strchr(name_utf8, '\r') == nullptr;
 }
```
MaxKellermann, line 87: Include missing

src/PlaylistDatabase.cxx
neheb: Page 705 of the standard says:

The contents and meaning of the header<cstring>are the same as the C standard library header<string.h>

MaxKellermann: Aye. Then you still need to convert some `<string.h> to `<cstring>`, because only the latter provides the `std::` namespace.

src/Permission.cxx
neheb: It's included in string_view, coming from util/StringView.hxx

Speaking of which, that file is funny. Why is there a C++ version check? MPD does not compile without C++17.

src/PlaylistDatabase.cxx
neheb: Some other header provides it.

neheb:
This totally breaks under libcxx. Will fix.

neheb:
Added partial libcxx fix.

src/Permission.cxx
MaxKellermann: Does the standard guarantee that `<string_view>` implies `<cstring>` or is that an implementation detail?

But we shouldn't rely on indirect includes by our own headers; these are certainly implementation details.

`StringView.hxx` has a version check because I shared this files with various projects, some of which must be buildable with GCC 6, but for "pure" C++17 projects like MPD, it is completely obsolete and is fading out.

src/PlaylistDatabase.cxx
MaxKellermann: Which other header? Don't rely on indirect includes, unless they are guaranteed.

neheb:
Fixed issues.

MaxKellermann:
```
../../src/util/WStringAPI.hxx:55:14: error: ‘wmemchr’ is not a member of ‘std’; did you mean ‘memchr’?
```
Cherry-picked the second commit to v0.21.x

MaxKellermann:
Oh wait, the second commit was a fixup for the first one...
e4dad42ca123b9f0f987218c68c42422d6735c0a

PR #858, opened 2020-05-03
Add tags relevant to classical music.

SimonPersson:
Hi!

In this PR, I add 5 tags that are interesting/relevant for classical music. They all show up in the [musicbrainz mapping](https://picard.musicbrainz.org/docs/mappings/) or in the [Xiph suggested tags](https://wiki.xiph.org/VorbisComment). I realise that the use of these tags is not obvious to listeners of contemporary/popular music, so let me try to give a justification for each:

### Ensemble
This is an ensemble that is playing the music, such as Wiener Philharmoniker. One could put the ensemble in an `ARTIST` tag, but that tag is generally very ambiguous for classical music, as it can refer to any of the composer, conductor, soloist, or ensemble.

### Movement
Most classical music collections contain `TITLE` tags like "Sonata for Piano No. 8 in C minor, Op. 13 "Pathétique": I. Grave - Allegro di molto e con brio", i.e. the work, opus, movement number, and movement name are all squeezed into a single tag. This is necessary as most music players only cater only to popular music. Thus, if we were to tag a track with an `ALBUM` tag "Klaviersonaten Nos. 30 & 31" and a `TITLE` tag "II. Allegro Molto", we cannot tell if we are currently listening to the second movement of the 30th or 31st piano sonata, as the player does not display the `WORK` tag. Hence, the `TITLE` tag needs to contain also the work, or our music collection will be useless in less sophisticated players, which would be a pity. As MPD already supports the `WORK` tag, adding tags for movement would be great! The two new tags, `MovementNumber` and `Movement` could correspond to "II" and "Allegro", for example.

### Composer sort
I want to look for Beethoven's 9th under B, for Beethoven, not L for Ludwig.

### Location
This is the location of the recording, e.g. "Wiener Musikverein". This is not limited to classical music, but in my opinion it is a nice piece of data to have.


MaxKellermann:
This PR contains lots of text which is neither in the commit message nor in the code. This text gets lost after merging.

Why do we need "Ensemble" when we already have "Performer"?
"Artist" is somehwat unspecific; sometimes, it's the composer, and sometimes, it's the performer. But there are both "composer" and "performer" tags, so why yet another performer-like tag?

I don't yet grasp the "movement" thing, but that's because I'm not used to classical music. But looking at MusicBrainz, it has "movement" and "movementnumber". There's no such thing as "movementname". This confuses me, and probably others.

SimonPersson:
Hi,

That's a good point. I have amended my commit to contain the same justification that I give in this PR.

Before I respond to your concerns, I'd just like to point out that I see your perspective as well. Adding new metadata like this is a slippery slope, and it could quickly grow out of control. I would love for there to be even more tags available ("lyricist", "engineer", "release type",  you get the point), but I narrowed this PR down to what I consider to be the most "essential" tags that are currently missing, particularly for classical.

> Why do we need "Ensemble" when we already have "Performer"?

We don't strictly need it, just like we don't strictly need the "Conductor" tag. We could also tag our tracks with `PERFORMER=Herbert von Karajan (Conductor)` and `PERFORMER=Berliner Philharmoniker (Orchestra)`. However, there is already precedence for using the "Ensemble" tag (as seen in the metabrainz wiki), and it does add valuable information. If we don't add it, we are essentially asking users to re-tag their music, and giving developers a harder time implementing a user interface with rich metadata.

> I don't yet grasp the "movement" thing, but that's because I'm not used to classical music

From a functionality perspective, the most important part to me is that the movements needs to be played in order. For popular music you might put a playlist on shuffle, which would shuffle tracks in random order. For classical, you would instead play the works in random order, but the movements in their normal order. That, and the visual aspect I mentioned in the original comment. It's just plain ugly to have such a long "title" tag, and it would be nicer to use a shorter "movement" tag in combination with "work".

> But looking at MusicBrainz, it has "movement" and "movementnumber". There's no such thing as "movementname". This confuses me, and probably others.

This is utterly confusing, and a stupid convention. In my opinion, these tags should be called "Movementname" and "Movementnumber", to avoid confusion. I went with "movement" and "movementname", as those are already in use for Vorbis comments (and presumably other plain key-val metadata formats). On the musicbrainz wiki, you can see that "movementnumber" and "movement" are the internal names in Picard, whereas "movementname" and "movement" are the tags used for Vorbis comments. By using this naming, we get support for Flac/Vorbis/Opus (I guess, I only tested on FLAC and I'm not too familiar with the MPD code base) "for free", no tag-mapping needed. It seemed to me that the existing tags were pretty standard vorbis comment tags, so I assumed that was the current convention in MPD? If not, we can have a different internal name in MPD, with a mapping to the format-specific names.

SimonPersson:
> I don't understand that, because I don't understand the difference between "performer" and "ensemble". Is there a definition that everybody agrees on? Your PR documentation doesn't contain a definition.

In the ensemble tag we store the group that is playing. This can be a band or an orchestra. In the performer tag we store individual performers, and what instrument they play. This is more or less according to the dictionary definition of those words.

Kodi supports the ensemble tag: https://trac.kodi.tv/ticket/17151
Roon supports ensemble: https://kb.roonlabs.com/File_Tag_Best_Practice

> So "movement" is like "disc" for modern music. An "album" (="work") contains several "discs" (="movements"), and each has a number of "tracks".

No, that analogy does not work.

Let's keep in mind that classical music predates recorded music, so maybe thinking in terms of live performances help. If a classical concert is advertised as a performance of some work (and that is generally how they are advertised), say Beethoven's 9th symphony, then every concert goer will expect to hear the whole symphony, all four movements, in order. Anything else would be strange. For example:

* The orchestra plays the movements in reverse order - this would be super strange.
* The orchestra plays the movements in random order - also super strange.
* The orchestra plays the 1st movement from Beethoven's 9th Symphony, followed by the 3rd movement of Mahler's 3rd Symphony, finishing with the 3rd movement of Mozart's 40th Symphony - fucking strange!

I can't think of a good analogy to popular music. Pretty much anything goes in a rock concert:

* Bruce Springsteen plays random songs from his albums - totally normal.
* Springsteen plays a song by Chuck Berry instead of one of his own - still totally normal.
* Springsteen plays a mix of songs, his own and others', in a medley - still pretty normal!

A better analogy can be made to literature. A novel is generally split into chapters. The chapters are not little pieces that can be enjoyed on their own though. To enjoy the book, you need all the chapters. And you certainly need the chapters in the right order, or the whole thing would not make any sense.

> This doesn't sound like it adds another dimension, and needs a new tag. Forgive my ignorance - I know little about how classical music is organized. I just don't see usable definitions here, and my analogies try to tell you what information I need to understand it

Let's look at an example. This is the back side of a CD that contains 2 Beethoven symphonies:
:format(jpeg):mode_rgb():quality(90)/discogs-images/R-1089282-1215190023.jpeg.jpg)

We can see that the disc is split into two works, Beethoven's 5th and 7th symphony. The CD begins with the 5th symphony, thus the movement numbers (the numbers followed by a `.`) and the track numbers (the numbers in a little square) are the same for the first 4 tracks. On track 5 though, the 7th symphony begins, and the movement number starts over from 1, so track and movement numbers are no longer the same.

As far as I can tell, we have one more "dimension" here. In popular music, we have: album -> track. In classical music, we can have: disc -> work -> movement.

> And anyway, why is there "movementnumber" and "movementname"? What do these things mean?

As seen in the picture above, each movement is numbered and "named"/"described" ("Allegro", "Presto").

dyenemonster (changes requested):
I think that the modification suggested by SimonPersson could be very beneficial for many people who listen to classical music, including myself. However, regarding the "MOVEMENT" and "MOVEMENTNAME" tags, I would rather choose "MOVEMENT" for the name of the movement and "MOVEMENTNUMBER" for the movement number (in roman numerals), just like MusicBrainz does (although this is absolutely not a standard).

SimonPersson:
@dyenemonster - I'm happy to have some support here, maybe the two of us can convince @MaxKellermann that these tags are actually useful.

I used the names that Musicbrainz suggested for Vorbis comments/tags, it looks like the other fields that MPD supports does that as well. Anyhow, I can switch to MOVEMENT/MOVEMENTNUMBER if that makes it more likely to get this PR merged!

Emilot:
+1 for this PR. I agree with SimonPersson and it would be just fine for the MOVEMENT/MOVEMENTNUMBER suggestion, to keep it inline with MusicBrainz suggestions.

SimonPersson:
@MaxKellermann - any chance you can revisit this PR? I have updated the PR to replace `MovementName` and `Movement` with `Movement` and `MovementNumber` based on the comments from @dyenemonster and @Emilot , and added a custom mapping for the vorbis comment style tags according to the [MusicBrainz mapping](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#movement-4).

SimonPersson:
I realized the changes to the `NEWS` file were for an older release. I have updated my commit to change the (unreleased) 0.23 release notes.

PR #861, opened 2020-05-07
move ffmpeg include below the C++ ones

neheb:
Fixes the following error with libcxx:

/usr/include/ffmpeg/libavutil/common.h:30:2: error: missing -D__STDC_CONSTANT_MACROS / #define __STDC_CONSTANT_MACROS

### DIFF
--- a/src/lib/ffmpeg/Filter.hxx
+++ b/src/lib/ffmpeg/Filter.hxx
@@ -22,12 +22,12 @@
 
 #include "Error.hxx"
 
+#include <new>
+#include <utility>
+
 extern "C" {
 #include <libavfilter/avfilter.h>
 }
-
-#include <new>
-#include <utility>
 
 struct AudioFormat;
 


MaxKellermann:
This looks like a FFmpeg but, doesn't it?

neheb:
libcxx defines this macro internally.

In any case, It's good practice for C++ headers to come before C ones.

MaxKellermann:
I'm not quite satisified with this. If it were ultimately necessary (and not just "good practice" for whatever reason I don't understand), then we'd need stricter measures to ensure this. For example, if one header includes a C header, how do we ensure that all necessary C++ headers are included before that header, even if none of the headers need those C++ headers for themselves?
I have no idea what `__STDC_CONSTANT_MACROS` is, but having a two underscore prefix, this means it's a compiler implementation detail which no application should care about. All of this seems so wrong, and it sounds like we're playing whack-a-mole with badly designed code.

neheb:
https://github.com/FFmpeg/FFmpeg/blob/62d92a878d89e35695e839f40065f2e936032f89/libavutil/common.h#L29

Seems that this is needed by C++ programs.

MaxKellermann:
But what does this comment mean, and what is this obscure macro?
Does FFmpeg document this fact somewhere?
What exactly does FFmpeg require C++ to include before including something? Which FFmpeg headers require that?
Your change may be necessary, but I want to fix this properly and everywhere in MPD, but to do that, we need to know the exact nature of the problem.

neheb:
I see https://github.com/FFmpeg/FFmpeg/commit/66872161cbb7187b0c20267653d415ef700aca35

That description doesn't seem too useful.

MaxKellermann:
This commit is as bad as it can get. It does something obscure, fiddling with implementation-defined macros which happen to be present in some version of the expected C++ compiler, but may not be present on others, without any documentation/explanation and a useless commit message repeating what the code does, but not why.

neheb:
Any progress on this? It's a simple bugfix.

MaxKellermann:
It's not a MPD bug. There's no progress because we still don't understand the problem.

neheb:
I did find something interesting: https://github.com/MusicPlayerDaemon/MPD/blob/master/src/decoder/plugins/FfmpegDecoderPlugin.cxx#L20

Looks like this problem was well known:

```
~/MPD (master)> git grep __STDC_CONSTANT_MACROS
src/decoder/plugins/FfmpegDecoderPlugin.cxx:#define __STDC_CONSTANT_MACROS
src/decoder/plugins/FfmpegIo.cxx:#define __STDC_CONSTANT_MACROS
src/decoder/plugins/FfmpegMetaData.cxx:#define __STDC_CONSTANT_MACROS
src/input/plugins/FfmpegInputPlugin.cxx:#define __STDC_CONSTANT_MACROS
src/lib/ffmpeg/Init.cxx:#define __STDC_CONSTANT_MACROS
src/lib/ffmpeg/LogCallback.cxx:#define __STDC_CONSTANT_MACROS
```

neheb:
Rebased.

joerg-krause:
The issue is fixed in the C11 and C++11 standards. Best would be to get rid of `__STDC_CONSTANT_MACROS` and support only C11 or C++11-compliant toolchains.

neheb:
> The issue is fixed in the C11 and C++11 standards. Best would be to get rid of `__STDC_CONSTANT_MACROS` and support only C11 or C++11-compliant toolchains.

FFMPEG is c99.

MaxKellermann:
> FFMPEG is c99.

Actually, even worse, FFmpeg is C90: https://github.com/FFmpeg/FFmpeg/blob/62a0368fb469277ccd6d9209ebdc39b03196eb59/doc/developer.texi#L115)
Living in the past. It's been 30 years.

MaxKellermann:
92bfdffa42304fb11663c30c2870cc0dfa205a3e

PR #880, opened 2020-05-29
src/db: fitting libmpdclient interface

ibmibmibm:
`mpd_connection_get_server_version` returns array of integers, not string.


MaxKellermann:
Damn, and this wasn't caught by the printf warning because that warning isn't supported for variadic templates... maybe it's about time MPD uses a type-safe string formatting library!

ibmibmibm:
Would you like to cherry-pick this commit into master branch?

PR #882, opened 2020-05-30
lib/icu: add null terminate in win32 string and wstring

ibmibmibm:
This prevent for strange extra char occur after any file path, while MPD tried to create or open any file on my windows 10 computer,

Also, using `std::make_unique` for creating `std::unique_ptr`, but `std::make__unique` is for c++14 and newer, should we prepare for compatibility to c++11?


MaxKellermann:
> is for c++14 and newer, should we prepare for compatibility to c++11?

MPD is C++17, no need for C++11 compatibility.

PR #884, opened 2020-05-31
misc: replace c header name with c++ header name, and sort in alphabet order

ibmibmibm:


MaxKellermann:
Did you verify that all of those C++ headers are guaranteed to provide all symbols in the root namespace, and not just in `std::`?

ibmibmibm:
I checked this goes fine with gcc, clang, and msvc. but seems that behavior doesn't exists on at least one compiler.
http://www.keil.com/support/man/docs/armcc/armcc_chr1359124948721.htm

I'll check those c functions, and make sure they use std namespace.

MaxKellermann:
Submitter did not reply.

PR #885, opened 2020-05-31
misc: replace /bin/tar call with python builtin library

ibmibmibm:


MaxKellermann:
This means we can't switch to `pbzip2`/`pigz`/etc. for parallel decompression. This isn't implemented currently, but would be easy to do with `subprocess`. What's your opinion on this?
What's the performance? Right now, `/usr/bin/tar` spawns `bzip2` as a different process, meaning that two CPU cores can be used (one for `tar` and one for `bzip2`). How does the pure Python implementation compare?

ibmibmibm:
Python use pure-python implement tar file parser and compiled binary decompress function, support gzip, bzip2 and xz.
Tar parsing should not consume too much time, because file writing is bounded with decompress speed much more.

ibmibmibm:
Maybe I can make python version a fallback when /bin/tar is not presented?

MaxKellermann:
> Maybe I can make python version a fallback when /bin/tar is not presented?

That sounds like a good idea.
I'd really like to default to tar/pbzip2 for best speed. Decompressing Boost can take a *lot* of time!

PR #924, opened 2020-07-21
lib/icu: fix build without libc iconv support

neheb:
Need to check for it in iconv.h. Otherwise meson prefixes a __builtin variant in the check.


MaxKellermann:
Cherry-picked to v0.21.x c876d6a51cddfb5853fde65a32da4920799b8942

PR #927, opened 2020-07-29
Support soxr custom recipes.

bitkeeper:
MPD uses soxr with prefined resample recipes. Soxr also support defining a recipe your self.
This commit will support a custom recipe by changing the existing quality setting to "custom".

The same structs as the predefined recipes uses can now be set by hand.

This will make the following settings available:

- precision 16|20|24|28|32 bits, example "28"
- phase_response - 0-50, example "45"
- passband_end - used bandwidth of source 80-99.7%, example "99.7.0"
- stopband_begin - anti aliasing 100.0+%, example "100".
- attenuation - signal reduciton in dB's, 0-30. example "3.0".
- flags "0" - additional bitmask with extra settings

The data is set in the structs soxr_quality_spec and soxr_io_spec (found in soxr.h).

### DIFF
--- a/src/pcm/SoxrResampler.cxx
+++ b/src/pcm/SoxrResampler.cxx
@@ -28,6 +28,7 @@
 
 #include <cassert>
 
+#include <cmath>
 #include <string.h>
 
 static constexpr Domain soxr_domain("soxr");
@@ -39,8 +40,16 @@ static constexpr unsigned long SOXR_DEFAULT_RECIPE = SOXR_HQ;
  */
 static constexpr unsigned long SOXR_INVALID_RECIPE = -1;
 
+/**
+ * Special value for the recipe selection for custom recipe.
+ */
+static constexpr unsigned long SOXR_CUSTOM_RECIPE = -2;
+
+static soxr_io_spec_t soxr_io_custom_recipe;
 static soxr_quality_spec_t soxr_quality;
 static soxr_runtime_spec_t soxr_runtime;
+static bool soxr_use_custom_recipe;
+
 
 static constexpr struct {
 	unsigned long recipe;
@@ -51,6 +60,7 @@ static constexpr struct {
 	{ SOXR_MQ, "medium" },
 	{ SOXR_LQ, "low" },
 	{ SOXR_QQ, "quick" },
+	{ SOXR_CUSTOM_RECIPE, "custom" },
 	{ SOXR_INVALID_RECIPE, nullptr }
 };
 
@@ -80,20 +90,117 @@ soxr_parse_quality(const char *quality) noexcept
 	return SOXR_INVALID_RECIPE;
 }
 
+static unsigned
+SoxrParsePrecision(unsigned value) {
+	switch (value) {
+	case 16:
+	case 20:
+	case 24:
+	case 28:
+	case 32:
+		break;
+	default:
+		throw FormatInvalidArgument(
+			"soxr converter invalid precision : %d [16|20|24|28|32]", value);
+	}
+	return value;
+}
+
+static double
+SoxrParsePhaseResponse(unsigned value) {
+	if (value > 100) {
+		throw FormatInvalidArgument(
+			"soxr converter invalid phase_respons : %d (0-100)", value);
+	}
+
+	return double(value);
+}
+
+static double
+SoxrParsePassbandEnd(const char *svalue) {
+	char *endptr;
+	double value = strtod(svalue, &endptr);
+	if (svalue == endptr || *endptr != 0) {
+		throw FormatInvalidArgument(
+			"soxr converter passband_end value not a number: %s", svalue);
+	}
+
+	if (value < 1 || value > 100) {
+		throw FormatInvalidArgument(
+			"soxr converter invalid passband_end : %s (1-100%%)", svalue);
+	}
+
+	return value / 100.0;
+}
+
+static double
+SoxrParseStopbandBegin(const char *svalue) {
+	char *endptr;
+	double value = strtod(svalue, &endptr);
+	if (svalue == endptr || *endptr != 0) {
+		throw FormatInvalidArgument(
+			"soxr converter stopband_begin value not a number: %s", svalue);
+	}
+
+	if (value < 100 || value > 199) {
+		throw FormatInvalidArgument(
+			"soxr converter invalid stopband_begin : %s (100-150%%)", svalue);
+	}
+
+	return value / 100.0;
+}
+
+static double
+SoxrParseAttenuation(const char *svalue) {
+	char *endptr;
+	double value = strtod(svalue, &endptr);
+	if (svalue == endptr || *endptr != 0) {
+		throw FormatInvalidArgument(
+			"soxr converter attenuation value not a number: %s", svalue);
+	}
+
+	if (value < 0 || value > 30) {
+		throw FormatInvalidArgument(
+			"soxr converter invalid attenuation : %s (0-30dB))", svalue);
+	}
+
+	return 1 / std::pow(10, value / 10.0);
+}
+
 void
 pcm_resample_soxr_global_init(const ConfigBlock &block)
 {
 	const char *quality_string = block.GetBlockValue("quality");
 	unsigned long recipe = soxr_parse_quality(quality_string);
+	soxr_use_custom_recipe = recipe == SOXR_CUSTOM_RECIPE;
+
 	if (recipe == SOXR_INVALID_RECIPE) {
 		assert(quality_string != nullptr);
-
 		throw FormatRuntimeError("unknown quality setting '%s' in line %d",
 					 quality_string, block.line);
-	}
+	} else if (recipe == SOXR_CUSTOM_RECIPE) {
+		// used to preset possible internal flags, like SOXR_RESET_ON_CLEAR
+		soxr_quality = soxr_quality_spec(SOXR_DEFAULT_RECIPE, 0);
+		soxr_io_custom_recipe = soxr_io_spec(SOXR_FLOAT32_I, SOXR_FLOAT32_I);
+		
+		soxr_quality.precision =
+			SoxrParsePrecision(block.GetBlockValue("precision", SOXR_HQ));
+		soxr_quality.phase_response =
+			SoxrParsePhaseResponse(block.GetBlockValue("phase_response", 50));
+		soxr_quality.passband_end = 
+			SoxrParsePassbandEnd(block.GetBlockValue("passband_end", "95.0"));
+		soxr_quality.stopband_begin = SoxrParseStopbandBegin(
+			block.GetBlockValue("stopband_begin", "100.0"));
+		// see soxr.h soxr_quality_spec.flags
+		soxr_quality.flags = (soxr_quality.flags & 0xFFFFFFC0) |
+			(block.GetBlockValue("flags", 0) & 0x3F);
+		soxr_io_custom_recipe.scale = 
+			SoxrParseAttenuation(block.GetBlockValue("attenuation", "0")); 
+	} else {
+		soxr_quality = soxr_quality_spec(recipe, 0);
+	};
 
-	soxr_quality = soxr_quality_spec(recipe, 0);
-
+	
 	FormatDebug(soxr_domain,
 		    "soxr converter '%s'",
 		    soxr_quality_name(recipe));
@@ -109,14 +216,32 @@ SoxrPcmResampler::Open(AudioFormat &af, unsigned new_sample_rate)
 	assert(audio_valid_sample_rate(new_sample_rate));
 
 	soxr_error_t e;
+	soxr_io_spec_t* p_soxr_io = nullptr;
+	if(soxr_use_custom_recipe) {
+		p_soxr_io = & soxr_io_custom_recipe;
+	}
 	soxr = soxr_create(af.sample_rate, new_sample_rate,
 			   af.channels, &e,
-			   nullptr, &soxr_quality, &soxr_runtime);
+			   p_soxr_io, &soxr_quality, &soxr_runtime);
 	if (soxr == nullptr)
 		throw FormatRuntimeError("soxr initialization has failed: %s",
 					 e);
 
 	FormatDebug(soxr_domain, "soxr engine '%s'", soxr_engine(soxr));
+	if (soxr_use_custom_recipe)
+		FormatDebug(soxr_domain,
+			    "soxr precision=%0.0f, phase_response=%0.2f, "
+				"passband_end=%0.2f, stopband_begin=%0.2f scale=%0.2f",
+				soxr_quality.precision, soxr_quality.phase_response,
+				soxr_quality.passband_end, soxr_quality.stopband_begin, 
+				soxr_io_custom_recipe.scale);
+	else {
+		FormatDebug(soxr_domain,
+			    "soxr precision=%0.0f, phase_response=%0.2f, "
+				"passband_end=%0.2f, stopband_begin=%0.2f",
+				soxr_quality.precision, soxr_quality.phase_response,
+				soxr_quality.passband_end, soxr_quality.stopband_begin);
+	}
 
 	channels = af.channels;
 
--- a/doc/plugins.rst
+++ b/doc/plugins.rst
@@ -746,6 +746,25 @@ Valid quality values for libsoxr:
 * "medium"
 * "low"
 * "quick"
+* "custom"
+
+If the quality is set to custom also the following settings are available:
+
+   * - Name
+     - Description
+   * - **precision**
+     - The precision in bits. Valid values 16,20,24,28 and 32  bits.
+   * - **phase_response**
+     - Between the 0-100, Where 0=MINIMUM_PHASE and 50=LINEAR_PHASE.
+   * - **passband_end**
+     - The % of source bandwidth where to start filtering. Typical between the 90-99.7.
+   * - **stopband_begin**
+     - The % of the source bandwidth Where the anti aliasing filter start. Value 100+.
+   * - **attenuation**
+     - Reduction in dB's to prevent clipping from the resampling process.
+   * - **flags**
+     - Bitmask with additional option see soxr documentation for specific flags.
+
 
 .. _output_plugins:
 


src/pcm/SoxrResampler.cxx
```diff
@@ -29,7 +29,7 @@
 #include <cassert>
 
 #include <string.h>
-
+#include <math.h>
```
MaxKellermann, line 32: Use C++ headers for new code, i.e. `<cmath>`

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +88,110 @@ soxr_parse_quality(const char *quality) noexcept
 	return SOXR_INVALID_RECIPE;
 }
 
+static unsigned 
+soxr_parse_precision(unsigned value) {
+	switch(value) {
+		case 16:
```
MaxKellermann, line 94: Bad coding style, see https://www.musicpd.org/doc/html/developer.html#code-style
This problem affects most of your code!

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +88,110 @@ soxr_parse_quality(const char *quality) noexcept
...
+		default:
+			throw FormatInvalidArgument("soxr converter invalid precision : %d [16|20|24|28|32]",
+					    value);
+			break;
+	}
+	return value;
+}
+
+static double
+soxr_parse_phase_respons(signed value) {
```
MaxKellermann, line 109: Why make this `signed` if negative values are not allowed?
And why is this integer if it implicitly casts to floating point?

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +88,110 @@ soxr_parse_quality(const char *quality) noexcept
...
+	}
+
+	return value;
+}
+
+static double
+soxr_parse_passband_end(const char* svalue) {
+	char* endptr;
+	double value=strtod (svalue, &endptr);
+	if(*endptr) {
```
MaxKellermann, line 122: This allows parsing an empty string.

src/pcm/SoxrResampler.cxx
```diff
@@ -39,9 +39,16 @@ static constexpr unsigned long SOXR_DEFAULT_RECIPE = SOXR_HQ;
  */
 static constexpr unsigned long SOXR_INVALID_RECIPE = -1;
 
+/**
+ * Special value for custom recipe.
+ */
+static constexpr unsigned long SOXR_CUSTOM_RECIPE = -2;
+
+static soxr_io_spec_t soxr_io;
```
MaxKellermann, line 47: I don't like the variable name; this is for storing the custom settings and is unused otherwise, but the name doesn't tell that.

src/pcm/SoxrResampler.cxx
```diff
@@ -109,14 +207,19 @@ SoxrPcmResampler::Open(AudioFormat &af, unsigned new_sample_rate)
 	assert(audio_valid_sample_rate(new_sample_rate));
 
 	soxr_error_t e;
+	soxr_io_spec_t* p_soxr_io = nullptr;
+	if(soxr_io.scale != 0) {
```
MaxKellermann, line 211: This is a bad check for "using custom settings". This happens to work in practice, but depends on implicit zero initialization of a floating point in a global variable - which is obscure. Better be explicit here about checking `SOXR_CUSTOM_RECIPE`.

src/pcm/SoxrResampler.cxx
```diff
@@ -109,14 +207,19 @@ SoxrPcmResampler::Open(AudioFormat &af, unsigned new_sample_rate)
...
 	soxr = soxr_create(af.sample_rate, new_sample_rate,
 			   af.channels, &e,
-			   nullptr, &soxr_quality, &soxr_runtime);
+			   p_soxr_io, &soxr_quality, &soxr_runtime);
 	if (soxr == nullptr)
 		throw FormatRuntimeError("soxr initialization has failed: %s",
 					 e);
 
 	FormatDebug(soxr_domain, "soxr engine '%s'", soxr_engine(soxr));
+	FormatDebug(soxr_domain, "soxr precision=%0.2f, phase_response=%0.2f, passband_end=%0.2f, stopband_begin=%0.2f scale=%0.2f", soxr_quality.precision, soxr_quality.phase_response, soxr_quality.passband_end, soxr_quality.stopband_begin, soxr_io.scale  );
```
MaxKellermann, line 222: This is logged even if the global variable is never initialized.

src/pcm/SoxrResampler.cxx
bitkeeper: Was not sure to if I should follow the code-style or the style of the already existing code surrounding the code change. Ok clear always use the code-style.

src/pcm/SoxrResampler.cxx
bitkeeper: Is soxr_io_custom_recipe better ?

src/pcm/SoxrResampler.cxx
bitkeeper: I will make sure that the  struct is initialized, else an additional variable would be required.

src/pcm/SoxrResampler.cxx
bitkeeper: What do you prefer leaving the scale out or an if with two different  FormatDebug statements ?

neheb:
I suggest running through git clang-format

bitkeeper:
@neheb Thanks for the tip.

neheb:
You have 3 separate commits. Need to run it through all of them.

Actually...you should probably squash all of the commits.

src/pcm/SoxrResampler.cxx
```diff
@@ -28,20 +28,35 @@
...
 
 static constexpr unsigned long SOXR_DEFAULT_RECIPE = SOXR_HQ;
 
+/**
+ * The SOXR_RESET_ON_CLEAR flag makes it possible to jump in the position off the current 
+ * track. Applied when using a custom recipe.
+ */
+static constexpr unsigned long SOXR_RESET_ON_CLEAR =
+	2147483648; // 2^31 defined in internal.h of soxr as RESET_ON_CLEAR
+
```
MaxKellermann, line 44: This wasn't in your first version of the PR - and I don't like using undocumented features with internal flags. This can break any time.

src/pcm/SoxrResampler.cxx
```diff
@@ -51,8 +66,8 @@ static constexpr struct {
 	{ SOXR_MQ, "medium" },
 	{ SOXR_LQ, "low" },
 	{ SOXR_QQ, "quick" },
-	{ SOXR_INVALID_RECIPE, nullptr }
-};
+	{ SOXR_CUSTOM_RECIPE, "custom"},
+	{ SOXR_INVALID_RECIPE, nullptr}};
```
MaxKellermann, line 70: Don't reformat existing code in an unrelated commit

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +95,114 @@ soxr_parse_quality(const char *quality) noexcept
...
+	}
+
+	return double(value);
+}
+
+static double
+SoxrParsePassbandEnd(const char *svalue) {
+	char *endptr;
+	double value = strtod(svalue, &endptr);
+	if (strlen(svalue) == 0 || *endptr) {
```
MaxKellermann, line 129: This is a clumsy check - three things are wrong here:
- it use `strlen()` to check if a string is empty, but you could better use `*svalue == 0`
- if you check `svalue`, you better do it before calling `strtod()`, because this doesn't depend on `strtod()`
- ... but even that would be complicated; the canonical way to rule out a parsing error here is checking `svalue == endptr` because that will tell you whether anything has been parsed at all

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +95,114 @@ soxr_parse_quality(const char *quality) noexcept
...
+			throw FormatInvalidArgument(
+				"soxr converter passband_end value not a number: %s", svalue);
+	}
+
+	if (value < 1 || value > 100) {
+			throw FormatInvalidArgument(
+				"soxr converter invalid passband_end : %d (1-100%)", value);
```
MaxKellermann, line 136: Two bugs in this format string:
- using `%d` but passing a `double`
- unescaped `%`

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +95,114 @@ soxr_parse_quality(const char *quality) noexcept
...
+	}
+
+	return value / 100.0;
+}
+
+static double
+SoxrParseStopbandBegin(const char *svalue) {
+	char *endptr;
+	double value = strtod(svalue, &endptr);
+	if (strlen(svalue) == 0 || *endptr) {
```
MaxKellermann, line 146: See above

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +95,114 @@ soxr_parse_quality(const char *quality) noexcept
...
+			throw FormatInvalidArgument(
+				"soxr converter stopband_begin value not a number: %s", svalue);
+	}
+
+	if (value < 100 || value > 199) {
+			throw FormatInvalidArgument(
+				"soxr converter invalid stopband_begin : %d (100-150%)", value);
```
MaxKellermann, line 153: See above

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +95,114 @@ soxr_parse_quality(const char *quality) noexcept
...
+	}
+
+	return value / 100.0;
+}
+
+static double
+SoxrParseAttenuation(const char *svalue) {
+	char *endptr;
+	double value = strtod(svalue, &endptr);
+	if (*endptr) {
```
MaxKellermann, line 163: `endptr==svalue` missing

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +95,114 @@ soxr_parse_quality(const char *quality) noexcept
...
+			throw FormatInvalidArgument(
+				"soxr converter attenuation value not a number: %s", svalue);
+	}
+
+	if (value < 0 || value > 30) {
+			throw FormatInvalidArgument(
+				"soxr converter invalid attenuation : %d (0-30)%)", value);
```
MaxKellermann, line 170: See above

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +95,114 @@ soxr_parse_quality(const char *quality) noexcept
...
+SoxrParseAttenuation(const char *svalue) {
+	char *endptr;
+	double value = strtod(svalue, &endptr);
+	if (*endptr) {
+			throw FormatInvalidArgument(
+				"soxr converter attenuation value not a number: %s", svalue);
+	}
+
+	if (value < 0 || value > 30) {
+			throw FormatInvalidArgument(
```
MaxKellermann, line 169: Bad indent

src/pcm/SoxrResampler.cxx
```diff
@@ -109,14 +218,32 @@ SoxrPcmResampler::Open(AudioFormat &af, unsigned new_sample_rate)
 	assert(audio_valid_sample_rate(new_sample_rate));
 
 	soxr_error_t e;
+	soxr_io_spec_t* p_soxr_io = nullptr;
+	if(soxr_io_custom_recipe.scale != 0) {
```
MaxKellermann, line 222: This is still not an explicit check for "custom recipe".
You initialized the global variable, but it is not guaranteed per soxr API documentation. Looking and the soxr source code, this looks completely wrong, because `soxr_io_spec()` initializes `scale=1`. Did you ever test this?
And: comparing floating-point values with `==` is conceptually broken. You should avoid this.

src/pcm/SoxrResampler.cxx
bitkeeper: Ok then I hide it by call `soxr_quality_spec` to init `soxr_quality`.

src/pcm/SoxrResampler.cxx
bitkeeper: Yes you are right about the scale=1 vs 0 mixing up linear vs db.
Will add static variable soxr_use_custom_recipe for explit use.

bitkeeper:
Will squash the commits when no more changes are required.

src/pcm/SoxrResampler.cxx
```diff
@@ -51,7 +60,8 @@ static constexpr struct {
 	{ SOXR_MQ, "medium" },
 	{ SOXR_LQ, "low" },
 	{ SOXR_QQ, "quick" },
-	{ SOXR_INVALID_RECIPE, nullptr }
+	{ SOXR_CUSTOM_RECIPE, "custom"},
+	{ SOXR_INVALID_RECIPE, nullptr}
```
MaxKellermann, line 64: Whitespace change

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +90,118 @@ soxr_parse_quality(const char *quality) noexcept
...
 }
 
+static unsigned
+SoxrParsePrecision(unsigned value) {
+	switch (value) {
+		case 16:
+		case 20:
+		case 24:
+		case 28:
+		case 32:
```
MaxKellermann, line 100: Bad indent

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +90,118 @@ soxr_parse_quality(const char *quality) noexcept
...
+			break;
+		default:
+			throw FormatInvalidArgument(
+				"soxr converter invalid precision : %d [16|20|24|28|32]", value);
+			break;
```
MaxKellermann, line 105: Unreachable `break`

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +90,118 @@ soxr_parse_quality(const char *quality) noexcept
...
+	}
+	return value;
+}
+
+static double
+SoxrParsePhaseRespons(unsigned value) {
```
MaxKellermann, line 111: Typo in function name?

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +90,118 @@ soxr_parse_quality(const char *quality) noexcept
...
+	}
+
+	return double(value);
+}
+
+static double
+SoxrParsePassbandEnd(const char *svalue) {
+	char *endptr;
+	double value = strtod(svalue, &endptr);
+	if (svalue == endptr) {
```
MaxKellermann, line 124: This doesn't check for garbage after the parsed number

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +90,118 @@ soxr_parse_quality(const char *quality) noexcept
...
 	if (recipe == SOXR_INVALID_RECIPE) {
 		assert(quality_string != nullptr);
-
 		throw FormatRuntimeError("unknown quality setting '%s' in line %d",
-					 quality_string, block.line);
-	}
-
-	soxr_quality = soxr_quality_spec(recipe, 0);
-
+					quality_string, block.line);
```
MaxKellermann, line 182: Whitespace change

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +90,118 @@ soxr_parse_quality(const char *quality) noexcept
...
+	return 1 / pow(10, value / 10.0);
+}
+
 void
 pcm_resample_soxr_global_init(const ConfigBlock &block)
 {
 	const char *quality_string = block.GetBlockValue("quality");
 	unsigned long recipe = soxr_parse_quality(quality_string);
+	soxr_use_custom_recipe = recipe == SOXR_CUSTOM_RECIPE;
+	soxr_io_custom_recipe = soxr_io_spec(SOXR_FLOAT32_I, SOXR_FLOAT32_I);
```
MaxKellermann, line 177: Initialize this variable only if this is really a custom recipe - move the initialization to the `if (recipe == SOXR_CUSTOM_RECIPE) {` block.

MaxKellermann (changes requested):

src/pcm/SoxrResampler.cxx
bitkeeper: When no number could be parsed the a expection is rased and if the number isn't within the expected range also.
Can you clarify ? You mean garbage as like a string with "1.23 ABC" where only the first part is parsed ?

src/pcm/SoxrResampler.cxx
bitkeeper: The preview in github shows more indent then it really is in the file itself. Please look at the raw file.
The `case` has two times the indent as for `switch`.

src/pcm/SoxrResampler.cxx
MaxKellermann: > You mean garbage as like a string with "1.23 ABC" where only the first part is parsed ?

Yes. Such a string should raise an error, because MPD should not silently ignore such garbage.

Usually, after `strto*()`, you check for errors using `endptr == s || *endptr != 0`. This checks whether anything was parsed at all, and whether everything was parsed.

src/pcm/SoxrResampler.cxx
MaxKellermann: >  The `case` has two times the indent as for `switch`.

That's the problem. It should have the same indent.
Coding style is a matter of taste, of course - and your indentation isn't absolutely wrong, just doesn't match the coding style used by the rest of MPD.

src/pcm/SoxrResampler.cxx
```diff
@@ -80,20 +90,117 @@ soxr_parse_quality(const char *quality) noexcept
...
+		throw FormatInvalidArgument(
+			"soxr converter attenuation value not a number: %s", svalue);
+	}
+
+	if (value < 0 || value > 30) {
+		throw FormatInvalidArgument(
+			"soxr converter invalid attenuation : %s (0-30dB))", svalue);
+	}
+
+	return 1 / pow(10, value / 10.0);
```
MaxKellermann, line 167: You included `<cmath>`, and this header doesn't provide `pow()`, but only `std::pow()`.

MaxKellermann:
I think this looks good now. Please rebase & fold.

MaxKellermann:
Oh and please fix all those trailing whitespace. When I look at your code in my editor (and not GitHub), my screen turns red due to code style warnings...

src/pcm/SoxrResampler.cxx
bitkeeper: hmm strange that it did compile ... something else is pulling it in,  but I have changed it and added the namespace.

MaxKellermann: Usually, C++ library implementations implement `<cmath>` by including the good old C header `<math.h>` and then copy symbols over to `namespace std`, so this *usually* works. But strictly, it is a mistake to assume that the symbol without namespace qualification is available. This may break any time, and I want MPD to be as correct as possible.

bitkeeper:
> Oh and please fix all those trailing whitespace. When I look at your code in my editor (and not GitHub), my screen turns red due to code style warnings...

Hmm I don't see any trailing whitespace in my editors ?

bitkeeper:
> I think this looks good now. Please rebase & fold.

done

MaxKellermann:
9aa432c07838256536c39b00d14c04462862b4fc
I fixed the whitespace errors for you.
I did not fix the author name. It's not some obscure GitHub auto-generated address.

PR #928, opened 2020-07-30
icu/meson: link against iconv when used externally

neheb:
This basically adds -liconv to LDFLAGS, fixing a linking issue when
iconv is missing in the libc.

The previous commit fixed finding iconv but did not fix linking.

### DIFF
--- a/src/lib/icu/meson.build
+++ b/src/lib/icu/meson.build
@@ -18,8 +18,18 @@ if icu_dep.found()
     'Init.cxx',
   ]
 elif not get_option('iconv').disabled()
-  have_iconv = compiler.has_function('iconv', prefix : '#include <iconv.h>')
-  conf.set('HAVE_ICONV', have_iconv)
+  have_iconv = c_compiler.has_header_symbol('iconv.h', 'iconv')
+  if have_iconv
+    libiconv = c_compiler.find_library('iconv', required : false)
+    conf.set('HAVE_ICONV', have_iconv)
+    libiconv_dep = static_library('iconv', icu_sources, include_directories: inc, dependencies: libiconv)
+    icu_dep = declare_dependency(link_with: libiconv_dep, dependencies: util_dep)
+  endif
+  if not have_iconv and c_compiler.has_function('iconv')
+    libiconv = []
+    have_iconv = true
+    conf.set('HAVE_ICONV', have_iconv)
+  endif
   if not have_iconv and get_option('iconv').enabled()
     error('iconv() not available')
   endif


neheb:
Turns out it also fails since it uses a wrong variable type. Also fixed.

src/lib/icu/meson.build
```diff
@@ -18,8 +18,18 @@ if icu_dep.found()
     'Init.cxx',
   ]
 elif not get_option('iconv').disabled()
-  have_iconv = compiler.has_function('iconv', prefix : '#include <iconv.h>')
-  conf.set('HAVE_ICONV', have_iconv)
+  have_iconv = c_compiler.has_header_symbol('iconv.h', 'iconv')
+  if have_iconv
+    libiconv = c_compiler.find_library('iconv')
```
MaxKellermann, line 23: Does this work when there's no libiconv?

src/lib/icu/meson.build
```diff
@@ -18,8 +18,18 @@ if icu_dep.found()
     'Init.cxx',
   ]
 elif not get_option('iconv').disabled()
-  have_iconv = compiler.has_function('iconv', prefix : '#include <iconv.h>')
-  conf.set('HAVE_ICONV', have_iconv)
+  have_iconv = c_compiler.has_header_symbol('iconv.h', 'iconv')
```
MaxKellermann, line 21: Why switch from `has_function` to `has_header_symbol`?

MaxKellermann:
Using `const` in `iconv()` violates the POSIX specification. Why would MPD want to support POSIX-violating implementations?

src/lib/icu/meson.build
neheb: has_function was moved down.

src/lib/icu/meson.build
neheb: If there's no libiconv it fails and moves on to the libc check.

neheb:
According to https://stackoverflow.com/questions/45938990/why-does-the-iconv-function-need-a-non-const-inbuffer , it's a POSIX vs SUSv2 thing. The const version is used for platforms that do not have iconv in libc. The safe side in that case is to use the older standard, which is what GNU libiconv does.

MaxKellermann:
Your link points to the obsolete 1997 version. It has been superseded 19 years ago by SUSv3, and then SUSv4 (2008), SUSv4-2013, SUSv4-2016 and SUSv4-2018.
This is SUSv3 from 2004, without `const`: https://pubs.opengroup.org/onlinepubs/009695399/functions/iconv.html
This is SUSv4-2018 (the most recent edition), also without `const`: https://pubs.opengroup.org/onlinepubs/9699919799/functions/iconv.html

The MPD project was created in 2003. At this time, the document you linked to had already been obsolete for 2 years.

neheb:
GNU libiconv 1.16 was released April 26 2019. It's not old.

GNU libiconv is used in OpenWrt for people that want various languages supported (libc iconv is partially stripped in musl and completely unavailable with uClibc-ng).

Rebased and replaced the variable to use auto to avoid verbosity.

src/lib/icu/meson.build
MaxKellermann: Not here:
```
Header <iconv.h> has symbol "iconv" : YES 

../../src/lib/icu/meson.build:23:4: ERROR: C shared or static library 'iconv' not found
```

MaxKellermann:
From libiconv 1.16:
```
extern LIBICONV_DLL_EXPORTED size_t iconv (iconv_t cd,  char* * inbuf, size_t *inbytesleft, char* * outbuf, size_t *outbytesleft);
```
The truth is that libiconv attempts to imitate the system's `iconv()` prototype. There is an obscure check in `srcm4/iconv.m4` which sets `ICONV_CONST`, and then generates `include/iconv.h` from `include/iconv.h.in`:
```
extern size_t iconv (iconv_t cd, @ICONV_CONST@ char* * inbuf, size_t *inbytesleft, char* * outbuf, size_t *outbytesleft);
```
On my system, which complies with SUSv3+, libiconv also generates a compliant header.

So you want to support OpenWRT, which is not compliant with SUSv3+, and only complies with the 21 year old obsolete SUSv2?

src/lib/icu/meson.build
neheb: Hmm need to add required : false looks like.

MaxKellermann:
Btw. Musl is SUSv3+ compliant (no `const`): http://git.musl-libc.org/cgit/musl/tree/include/iconv.h

All this sounds like there are other problems which cause libiconv to generate the wrong declaration on your system. Why don't we fix those problems instead of making MPD support a 21-year old obsolete spec?

src/lib/icu/meson.build
```diff
@@ -18,8 +18,18 @@ if icu_dep.found()
...
+  if have_iconv
+    libiconv = c_compiler.find_library('iconv', required : false)
+    conf.set('HAVE_ICONV', have_iconv)
+    libiconv_dep = static_library('iconv', icu_sources, include_directories: inc, dependencies: libiconv)
```
MaxKellermann, line 25: This compiles `icu_sources` twice.

src/lib/icu/meson.build
MaxKellermann: OK, `has_function` is still there, but that doesn't answer why you used `has_header_symbol` now.
Also, your change uses `libiconv` even if the C library already provides `iconv()`. Why did you decide to do that?

neheb:
> From libiconv 1.16:
> 
> ```
> extern LIBICONV_DLL_EXPORTED size_t iconv (iconv_t cd,  char* * inbuf, size_t *inbytesleft, char* * outbuf, size_t *outbytesleft);
> ```
> 
> The truth is that libiconv attempts to imitate the system's `iconv()` prototype. There is an obscure check in `srcm4/iconv.m4` which sets `ICONV_CONST`, and then generates `include/iconv.h` from `include/iconv.h.in`:
> 
> ```
> extern size_t iconv (iconv_t cd, @ICONV_CONST@ char* * inbuf, size_t *inbytesleft, char* * outbuf, size_t *outbytesleft);
> ```
> 
> On my system, which complies with SUSv3+, libiconv also generates a compliant header.
> 
> So you want to support OpenWRT, which is not compliant with SUSv3+, and only complies with the 21 year old obsolete SUSv2?

That is...interesting.

From libiconv's configure.ac

```
if test "$am_cv_func_iconv" = yes -a -n "$am_cv_proto_iconv_arg1"; then
  ICONV_CONST="const"
else
  ICONV_CONST=""
fi
AC_SUBST([ICONV_CONST])
```

I have no idea what am_cv_proto_iconv_arg1 is.

Sounds like I can drop this PR if I can get it to configure without const.

src/lib/icu/meson.build
neheb: libiconv usually has more character maps. In OpenWrt specifically, some of those maps in musl are patched out.

neheb:
Got rid of the second commit. This no longer seems to be an issue...

I think the previous version in OpenWrt (libiconv 1.11.1) was buggy. There were also a bunch of patches, all of which get removed in the 1.16 update.

src/lib/icu/meson.build
neheb: I *think* I fixed.

src/lib/icu/meson.build
MaxKellermann: If you prefer the external libiconv, there should be a comment in the Meson source explaining it.
This change was not mentioned in your commit message, so finding this as surprising for me.

src/lib/icu/meson.build
MaxKellermann: Now the `icu_sources` are only compiled when an external libiconv is used. Not when libc's `iconv()` is used and not when libicu is used.

neheb:
superseded by https://github.com/MusicPlayerDaemon/MPD/pull/1515

PR #931, opened 2020-08-01
doc/plugins.rst: document libmpg123 limitations

naglis:


neheb:
Does libmad have the same limitation?

MaxKellermann:
> Does libmad have the same limitation?

No.

neheb:
Cool.

PR #951, opened 2020-08-27
use std::fill

neheb:
Changed C array to C++ array in order to have begin and end.

### DIFF
--- a/src/decoder/plugins/MadDecoderPlugin.cxx
+++ b/src/decoder/plugins/MadDecoderPlugin.cxx
@@ -436,11 +436,11 @@ static constexpr unsigned IN_MAGIC = (('I' << 8) | 'n');
 static constexpr unsigned FO_MAGIC = (('f' << 8) | 'o');
 
 struct xing {
-	long flags;             /* valid fields (see below) */
-	unsigned long frames;   /* total number of frames */
-	unsigned long bytes;    /* total number of bytes */
-	unsigned char toc[100]; /* 100-point seek table */
-	long scale;             /* VBR quality */
+	long flags;                         /* valid fields (see below) */
+	unsigned long frames;               /* total number of frames */
+	unsigned long bytes;                /* total number of bytes */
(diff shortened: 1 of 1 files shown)


src/decoder/plugins/MadDecoderPlugin.cxx
```diff
@@ -436,11 +436,11 @@ static constexpr unsigned IN_MAGIC = (('I' << 8) | 'n');
...
 
 struct xing {
-	long flags;             /* valid fields (see below) */
-	unsigned long frames;   /* total number of frames */
-	unsigned long bytes;    /* total number of bytes */
-	unsigned char toc[100]; /* 100-point seek table */
-	long scale;             /* VBR quality */
+	long flags;                         /* valid fields (see below) */
+	unsigned long frames;               /* total number of frames */
+	unsigned long bytes;                /* total number of bytes */
```
MaxKellermann, line 441: Don't reformat those lines. That's not relevant to your change.

src/decoder/plugins/MadDecoderPlugin.cxx
```diff
@@ -517,8 +517,7 @@ parse_xing(struct xing *xing, struct mad_bitptr *ptr, int *oldbitlen) noexcept
 	if (xing->flags & XING_TOC) {
 		if (bitlen < 800)
 			return false;
-		for (unsigned char & i : xing->toc)
-			i = mad_bit_read(ptr, 8);
+		std::fill(xing->toc.begin(), xing->toc.end(), mad_bit_read(ptr, 8));
```
MaxKellermann, line 520: This looks very broken. This reduces 100 `mad_bit_read()` calls to just one, by moving it out of the loop.

MaxKellermann (changes requested):

src/decoder/plugins/MadDecoderPlugin.cxx
neheb: Hrm good point.

PR #955, opened 2020-09-14
youtube-dl input/playlist plugin

Rio6:
This is my implementation of youtube-dl plugin, I made it after realizing #223 exists :roll_eyes: 

It provides basically the same functionalities as #223, so I'm fine if this get's closed, since it was here first.

Anyways, I'll open this PR for comparing, and get some critiques on my code if you'd like.

### DIFF
--- a/src/input/TaggedInputStream.hxx
+++ b/src/input/TaggedInputStream.hxx
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2003-2020 The Music Player Daemon Project
+ * http://www.musicpd.org
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef MPD_TAGGED_INPUTSTREAM
+#define MPD_TAGGED_INPUTSTREAM
+
+#include "../ProxyInputStream.hxx"
+#include "tag/Tag.hxx"
+
+#include <memory>
+
+class TaggedInputStream final : public ProxyInputStream {
+	std::unique_ptr<Tag> tag;
+
+public:
+	TaggedInputStream(InputStreamPtr _input, std::unique_ptr<Tag> &&_tag) :
+		ProxyInputStream(std::move(_input)), tag(std::move(_tag))
+	{}
+
+	std::unique_ptr<Tag> ReadTag() noexcept override {
+		return std::exchange(tag, nullptr);
+	}
+};
+
+#endif
--- a/src/input/TaggedInputStream.hxx
+++ b/src/input/TaggedInputStream.hxx
@@ -23,8 +23,11 @@
 #include "../ProxyInputStream.hxx"
 #include "tag/Tag.hxx"
 
-#include <memory>
-
+/**
+ * Wrapper stream that provides tags for the inner stream.  If the inner
+ * stream also has tags, it merges them.  The tag entries from the inner
+ * stream are overwrited by the outer tag entries.
+ */
 class TaggedInputStream final : public ProxyInputStream {
 	std::unique_ptr<Tag> tag;
 
@@ -34,7 +37,14 @@ public:
 	{}
 
 	std::unique_ptr<Tag> ReadTag() noexcept override {
-		return std::exchange(tag, nullptr);
+		auto inner_tag = ProxyInputStream::ReadTag();
+
+		if(!inner_tag)
+			return std::exchange(tag, nullptr);
+		else if(!tag)
+			return inner_tag;
+		else
+			return Tag::Merge(std::exchange(tag, nullptr), std::move(inner_tag));
 	}
 };
 
--- a/src/input/plugins/YoutubeInputPlugin.cxx
+++ b/src/input/plugins/YoutubeInputPlugin.cxx
@@ -17,9 +17,6 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-/* necessary because libavutil/common.h uses UINT64_C */
-#define __STDC_CONSTANT_MACROS
-
 #include "YoutubeInputPlugin.hxx"
 #include "CurlInputPlugin.hxx"
 #include "PluginUnavailable.hxx"
@@ -47,9 +44,7 @@ input_youtube_init(EventLoop &, const ConfigBlock &)
 static InputStreamPtr
 input_youtube_open(const char *uri, Mutex &mutex)
 {
-	if(!StringEqualsCaseASCII(uri, "https://www.youtube.com", 19))
-		return nullptr;
-
+	// TODO fix command injection
 	char *cmd = xstrcatdup("youtube-dl --extract-audio --get-url --youtube-skip-dash-manifest ", uri);
 	FILE *stream = popen(cmd, "r");
 	free(cmd);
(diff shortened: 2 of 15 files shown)


MaxKellermann:
This PR is confusing because there are so many incremental commits which fix mistakes in previous commits of this PR. Can you fold those fixup commits?

src/input/TaggedInputStream.hxx
```diff
@@ -0,0 +1,41 @@
...
+class TaggedInputStream final : public ProxyInputStream {
+	std::unique_ptr<Tag> tag;
+
+public:
+	TaggedInputStream(InputStreamPtr _input, std::unique_ptr<Tag> &&_tag) :
+		ProxyInputStream(std::move(_input)), tag(std::move(_tag))
+	{}
+
+	std::unique_ptr<Tag> ReadTag() noexcept override {
+		return std::exchange(tag, nullptr);
```
MaxKellermann, line 37: This hides the tags that may be provided by the inner InputStream.

src/input/TaggedInputStream.hxx
```diff
@@ -0,0 +1,41 @@
...
+
+#ifndef MPD_TAGGED_INPUTSTREAM
+#define MPD_TAGGED_INPUTSTREAM
+
+#include "../ProxyInputStream.hxx"
+#include "tag/Tag.hxx"
+
+#include <memory>
+
+class TaggedInputStream final : public ProxyInputStream {
```
MaxKellermann, line 28: API documentation, please

src/input/TaggedInputStream.hxx
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef MPD_TAGGED_INPUTSTREAM
+#define MPD_TAGGED_INPUTSTREAM
+
+#include "../ProxyInputStream.hxx"
+#include "tag/Tag.hxx"
```
MaxKellermann, line 24: Forward declaration should be possibe

src/input/plugins/YoutubeInputPlugin.cxx
```diff
@@ -0,0 +1,182 @@
...
+
+#include <yajl/yajl_tree.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static Domain youtube_domain("youtube");
+
+static const char *input_youtube_prefixes[] = {
```
MaxKellermann, line 41: Missing `const`

src/input/plugins/YoutubeInputPlugin.cxx
```diff
@@ -0,0 +1,182 @@
...
+	"https",
+	nullptr
+};
+
+static void
+input_youtube_init(EventLoop &, const ConfigBlock &)
+{
+	if(WEXITSTATUS(system("youtube-dl --version > /dev/null")) != 0)
```
MaxKellermann, line 49: `system()` should never be used; spawning child processes really is complicated, for example `system()` doesn't restore the signal mask and many other things.

src/util/ExecOpen.cxx
```diff
@@ -0,0 +1,62 @@
...
+
+	AtScopeExit(&fd) {
+		close(fd[1]);
+	};
+
+	posix_spawn_file_actions_t spawn_file;
+	posix_spawn_file_actions_init(&spawn_file);
+	posix_spawn_file_actions_adddup2(&spawn_file, fd[1], STDOUT_FILENO);
+
+	errno = posix_spawnp(pid, "youtube-dl", &spawn_file, nullptr, const_cast<char**>(args), nullptr);
```
MaxKellermann, line 47: This doesn't restore the signal mask either....
Also why does this hard-code the program "youtube-dl"?

src/util/ExecOpen.cxx
```diff
@@ -0,0 +1,62 @@
...
+	if(errno) {
+		return nullptr;
+	}
+
+	return fdopen(fd[0], "r");
```
MaxKellermann, line 52: Why use stdio here?

src/util/ExecOpen.hxx
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+#ifndef EXECOPEN_HXX
+#define EXECOPEN_HXX
+
+#include <stdio.h>
+
+/*
+ * like popen(3) except it doesn't use shell
+ * Returns NULL on error, with errno set
```
MaxKellermann, line 27: Throw exception on error.

src/util/ExecOpen.hxx
```diff
@@ -0,0 +1,38 @@
...
+ */
+FILE *exec_open(int *pid, const char *cmd, const char *const args[]);
```
MaxKellermann, line 29: What is this `pid` parameter and why is it not `pid_t`?

src/util/ExecOpen.hxx
```diff
@@ -0,0 +1,38 @@
...
+
+/*
+ * Wait for child pid to exit
+ * Returns child return code
+ * On error, returns -1 and errno is set
```
MaxKellermann, line 34: And what shall the caller do when an error occurs? What happens to the child process?

src/util/meson.build
```diff
@@ -29,6 +29,7 @@ util = static_library(
   'ByteReverse.cxx',
   'format.c',
   'BitReverse.cxx',
+  'ExecOpen.cxx',
```
MaxKellermann, line 32: This should not be in libutil, because libutil should be portable generic code without any dependencies. Better put it in libsystem.

src/util/ExecOpen.cxx
```diff
@@ -0,0 +1,62 @@
...
+	int fd[2];
+	if(pipe2(fd, O_CLOEXEC) < 0) {
+		return nullptr;
+	}
+
+	AtScopeExit(&fd) {
+		close(fd[1]);
+	};
+
+	posix_spawn_file_actions_t spawn_file;
```
MaxKellermann, line 43: This doesn't work on Windows, so your code will break the Windows build.

src/input/plugins/YoutubeInputPlugin.cxx
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	/* Read youtube-dl output */
+	std::string json;
+	char buf[512];
+	ssize_t size = 0;
+	while(true) {
+		size = fread(buf, sizeof(char), 512, stream);
+		if(size > 0)
+			json.append(buf, size);
```
MaxKellermann, line 79: What if the process just happens to print a few terabytes of data? This will crash MPD.

src/input/plugins/YoutubeInputPlugin.cxx
```diff
@@ -0,0 +1,182 @@
...
+	int pid;
+	FILE *stream = exec_open(&pid, "youtube-dl", args);
+	if(!stream) {
+		LogErrno(youtube_domain, "Can't spawn youtube-dl");
+		return nullptr;
+	}
+
+	/* Read youtube-dl output */
+	std::string json;
+	char buf[512];
```
MaxKellermann, line 74: Move this buffer inside the loop.

src/input/plugins/YoutubeInputPlugin.cxx
```diff
@@ -0,0 +1,182 @@
...
+	ssize_t size = 0;
```
MaxKellermann, line 75: What's the point of initializing this variable with this arbitrary value?

MaxKellermann (changes requested):

Rio6:
Fixed the smaller things. But I think I'll need to redo how it spawns and communicates with youtube-dl.

arcnmx:
Hm, it might be worth combining efforts? I think the main difference between this and the ytdl plugin is the use of yajl to parse incrementally, and perform nonblocking reads from the process via the event loop. This also ties into the RemoteTagScanner functionality as well that interacts with the playlist plugin flat vs full mode issue.

For the input stream I think I had explored using `ProxyInputStream` but don't recall how far I got or if it was missing something I wanted it to do. The main reason I haven't pushed ahead with my plugin yet (besides having not finished the docs) is that I still intend to fix the input stream to be able to recover from expired URLs - resuming playback after a long pause sometimes requires obtaining a new URL.

Rio6:
If your branch is open for prs then I'd be glad to try add things to it :) If you don't mind.

arcnmx:
Oh of course feel free to, the latest should be on the ytdl branch.

PR #964, opened 2020-09-27
clang-tidy: change several bind calls to lambdas

neheb:
Found with modernize-avoid-bind.

The defaulted constructor is needed as normally, an implicitly deleted
one is used.

### DIFF
--- a/src/lib/dbus/Message.hxx
+++ b/src/lib/dbus/Message.hxx
@@ -48,6 +48,8 @@ class Message {
 public:
 	Message() noexcept = default;
 
+	Message(const Message &src) = default;
+
 	Message(Message &&src) noexcept
 		:msg(std::exchange(src.msg, nullptr)) {}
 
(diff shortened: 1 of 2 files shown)


MaxKellermann:
> The defaulted constructor is needed as normally, an implicitly deleted one is used.

I don't understand that.

neheb:
https://en.cppreference.com/w/cpp/language/default_constructor

Note the implicit sections. This one is implicitly declared as deleted.

MaxKellermann:
I know that part, but why is it needed?

neheb:
I don't think a deleted constructor can be used?

MaxKellermann:
True, but why is it needed?

neheb:
I assume the lambdas use it?

MaxKellermann:
This is supposed to be a refactoring-only PR, isn't it? Why does it change the code behavior?
Also note that there's a good reason why this copy constructor doesn't exist: the way you declared it, it will crash! You can't simply declare a defaulted copy constructor - this class manages a raw pointer allocated by an external library, and you can't have two C++ objects owning the raw pointer at the same time.

neheb:
Let's ignore this.

PR #965, opened 2020-09-27
clang-tidy: add explicit deleted constructors

neheb:

### DIFF
--- a/src/archive/plugins/Bzip2ArchivePlugin.cxx
+++ b/src/archive/plugins/Bzip2ArchivePlugin.cxx
@@ -71,6 +71,9 @@ public:
 			 Mutex &mutex);
 	~Bzip2InputStream() noexcept override;
 
+	Bzip2InputStream(const Bzip2InputStream &) = delete;
+	Bzip2InputStream &operator=(const Bzip2InputStream &) = delete;
+
 	/* virtual methods from InputStream */
 	[[nodiscard]] bool IsEOF() const noexcept override;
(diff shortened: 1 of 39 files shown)


MaxKellermann:
Why?

neheb:
https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-special-member-functions.html

I would guess this is to avoid different compilers having different behaviors.

MaxKellermann:
What compiler would have a different behavior? The behavior is defined by the standard.
If a class has a move constructor/operator, the copy constructor/operator is implicitly deleted. That's fine. I don't agree that it should be explicitly deleted. The standard defines exactly the behavior I want there. Explicitly deleting these only creates unnecessary code noise (boilerplate code) and reduces code readability. The syntax for those operators/constructors is too complicated and error prone already. I want less of that, not more.
There are a few good pieces in your change: for example, `class InputStream` is copyable, but it should not be. Deleting its copy constructor/operator should be enough to take account for all sub classes. The proxy database changes are good.
What I don't understand is why you declare defaulted destructors. That part is not explained in the core guidelines. They say you should always declare all 5, but doesn't explain what use a defaulted destructor is.

neheb:
I initially declared the destructors as deleted. Then it did not compile.

neheb:
Note that the check has several options. I think AllowMissingMoveFunctionsWhenCopyIsDeleted and AllowSoleDefaultDtor would be what you want.

neheb:
Did just that.

PR #978, opened 2020-10-14
clang-tidy: convert to all/any_of

neheb:
Found with readability-use-anyofallof

### DIFF
--- a/src/command/FileCommands.cxx
+++ b/src/command/FileCommands.cxx
@@ -113,12 +113,9 @@ IsValidName(const StringView s) noexcept
 	if (s.empty() || !IsAlphaASCII(s.front()))
 		return false;
 
-	for (const char ch : s) {
-		if (!IsAlphaASCII(ch) && ch != '_' && ch != '-')
-			return false;
-	}
-
-	return true;
+	return std::none_of(s.begin(), s.end(), [=](const auto &ch) {
+		return !IsAlphaASCII(ch) && ch != '_' && ch != '-';
+	});
 }
 
 gcc_pure
--- a/src/db/plugins/ProxyDatabasePlugin.cxx
+++ b/src/db/plugins/ProxyDatabasePlugin.cxx
@@ -356,11 +356,9 @@ SendConstraints(mpd_connection *connection, const SongFilter &filter)
 						 filter.ToExpression().c_str());
 #endif
 
-	for (const auto &i : filter.GetItems())
-		if (!SendConstraints(connection, *i))
-			return false;
-
-	return true;
+	return std::all_of(
+		filter.GetItems().begin(), filter.GetItems().end(),
+		[=](const auto &item) { return SendConstraints(connection, *item); });
 }
 
 static bool
@@ -896,11 +894,8 @@ IsFilterFullySupported(const SongFilter &filter,
 	(void)connection;
 #endif
 
-	for (const auto &i : filter.GetItems())
-		if (!IsFilterSupported(*i))
-			return false;
-
-	return true;
+	return std::all_of(filter.GetItems().begin(), filter.GetItems().end(),
+			   [](const auto &item) { return IsFilterSupported(*item); });
 }
 
 gcc_pure
--- a/src/song/Filter.cxx
+++ b/src/song/Filter.cxx
@@ -429,29 +429,27 @@ SongFilter::Match(const LightSong &song) const noexcept
 bool
 SongFilter::HasFoldCase() const noexcept
 {
-	for (const auto &i : and_filter.GetItems()) {
-		if (auto t = dynamic_cast<const TagSongFilter *>(i.get())) {
-			if (t->GetFoldCase())
-				return true;
-		} else if (auto u = dynamic_cast<const UriSongFilter *>(i.get())) {
-			if (u->GetFoldCase())
-				return true;
-		}
-	}
+	return std::any_of(
+		and_filter.GetItems().begin(), and_filter.GetItems().end(),
+		[](const auto &item) {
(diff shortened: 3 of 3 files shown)


src/command/FileCommands.cxx
```diff
@@ -113,12 +113,9 @@ IsValidName(const StringView s) noexcept
...
 		return false;
 
-	for (const char ch : s) {
-		if (!IsAlphaASCII(ch) && ch != '_' && ch != '-')
-			return false;
-	}
-
-	return true;
+	return std::all_of(s.begin(), s.end(), [=](const auto &ch) {
+		return !IsAlphaASCII(ch) && ch != '_' && ch != '-';
```
MaxKellermann, line 117: This check is inverted

neheb: Fixed.

src/db/plugins/ProxyDatabasePlugin.cxx
```diff
@@ -356,11 +356,9 @@ SendConstraints(mpd_connection *connection, const SongFilter &filter)
...
 #endif
 
-	for (const auto &i : filter.GetItems())
-		if (!SendConstraints(connection, *i))
-			return false;
-
-	return true;
+	return std::all_of(
+		filter.GetItems().begin(), filter.GetItems().end(),
+		[=](const auto &item) { return !SendConstraints(connection, *item); });
```
MaxKellermann, line 361: .... aaaaand this check is also inverted!

src/db/plugins/ProxyDatabasePlugin.cxx
```diff
@@ -896,11 +894,8 @@ IsFilterFullySupported(const SongFilter &filter,
 	(void)connection;
 #endif
 
-	for (const auto &i : filter.GetItems())
-		if (!IsFilterSupported(*i))
-			return false;
-
-	return true;
+	return std::all_of(filter.GetItems().begin(), filter.GetItems().end(),
+			   [](const auto &item) { return !IsFilterSupported(*item); });
```
MaxKellermann, line 898: And so is this - inverted!

src/db/update/ExcludeList.cxx
```diff
@@ -74,17 +74,12 @@ ExcludeList::Check(Path name_fs) const noexcept
 		}
 	}
 
-	for (const auto &i : patterns) {
-		try {
-			if (i.Check(NarrowPath(name_fs).c_str()))
-				return true;
-		} catch (...) {
-		}
```
MaxKellermann, line 82: You removed the try/catch. Why?

src/song/Filter.cxx
```diff
@@ -429,29 +429,23 @@ SongFilter::Match(const LightSong &song) const noexcept
...
-		} else if (auto u = dynamic_cast<const UriSongFilter *>(i.get())) {
-			if (u->GetFoldCase())
-				return true;
-		}
-	}
-
-	return false;
+	return std::any_of(and_filter.GetItems().begin(), and_filter.GetItems().end(),
+			   [](const auto &item) {
+				   return dynamic_cast<const TagSongFilter *>(item.get())
```
MaxKellermann, line 434: This will crash if the dynamic_cast fails.

src/song/Filter.cxx
```diff
@@ -429,29 +429,23 @@ SongFilter::Match(const LightSong &song) const noexcept
...
+						  ->GetFoldCase() ||
+					  dynamic_cast<const UriSongFilter *>(item.get())
```
MaxKellermann, line 436: Another nullptr dereference.

src/db/plugins/ProxyDatabasePlugin.cxx
neheb: Seems the clang-tidy people have some fixing to do...

src/song/Filter.cxx
neheb: Good point.

src/db/update/ExcludeList.cxx
neheb: Wasn't sure about this. I assume it would have to be included in the lambda.

src/db/plugins/ProxyDatabasePlugin.cxx
MaxKellermann: This hunk is correct now, but there's double negation, making it more complicated than necessary: why not `all_of` without the `!`?

src/db/plugins/ProxyDatabasePlugin.cxx
```diff
@@ -896,11 +894,8 @@ IsFilterFullySupported(const SongFilter &filter,
 	(void)connection;
 #endif
 
-	for (const auto &i : filter.GetItems())
-		if (!IsFilterSupported(*i))
-			return false;
-
-	return true;
+	return std::none_of(filter.GetItems().begin(), filter.GetItems().end(),
+			    [](const auto &item) { return !IsFilterSupported(*item); });
```
MaxKellermann, line 898: Same here - double negation.

src/db/update/ExcludeList.cxx
MaxKellermann: I don't understand what you assumed. The try/catch is still missing, making this function crash on error.

src/db/plugins/ProxyDatabasePlugin.cxx
neheb: Fixed.

src/db/update/ExcludeList.cxx
neheb: Got rid of the conversion. It makes no sense honestly.

PR #983, opened 2020-10-20
system/FileDescriptor: fix Duplicate result

arcnmx:
`dup2` returns `new_fd`, so assuming the intention was for `true` to indicate success, this needs to check for `== new_fd.Get()` or `>= 0` or `!= -1` instead.


MaxKellermann:
Your PR contains explanatory text, but the explanation is missing in the commit message.

arcnmx:
Sorry, is that better?

MaxKellermann:
This is still less text compared to this PR text, but acceptable.
Rule of thumb: if you add text to a PR, this means that your commit messages are bad. Don't add text to PRs, this should never be necessary.

PR #1003, opened 2020-11-09
Explicitly remove socket FDs from EventLoop before closing

arcnmx:
`SocketEvent` knows the socket is open and registered so it seems unnecessary to rely on `AbandonFD` and assume the kernel will remove the fd from the epoll list when it can be done explicitly.

### Why this is relevant

- `AbandonFD` assumes that upon closing the socket, the FD will be automatically removed from the epoll list. That fd is associated with a reference to the `SocketEvent`, so this is an important and dangerous assumption to get wrong. In the case that the FD isn't immediately removed from the list by the kernel, the event loop can crash due to the `SocketEvent` being destroyed and it being a use-after-free bug at that point.
- If a socket FD happens to be duplicated, then closing the SocketEvent FD will not automatically remove it from epoll, and will trigger said bug/crash. It is only automatically removed when all FD references to the underlying socket/resource are closed?
- A `fork()` is one example where a socket FD can be duplicated and result in this situation.
    - `CLOEXEC` might be considered mitigation for this but also introduces a race condition where the crash can occur between a `fork()` and `exec()` without additional synchronization to freeze the event loop.

One could argue the mpd event loop isn't fork-safe, and thus should be allowed to use `AbandonFD` however it likes. A decision on whether this is intended should probably be declared; but either way this fix seems appropriate in cases where `Abandon` isn't actually necessary. It also might be possible to fix `AbandonFD` to mark the `SocketEvent` as removed without using `EPOLL_CTL_DEL`?

(I can add some of these details to the commit message if you like?)

### DIFF
--- a/src/event/SocketEvent.cxx
+++ b/src/event/SocketEvent.cxx
@@ -44,7 +44,7 @@ SocketEvent::Close() noexcept
 		return;
 
 	if (std::exchange(scheduled_flags, 0) != 0)
-		loop.AbandonFD(*this);
+		loop.RemoveFD(fd.Get(), *this);
 	fd.Close();
 }
 


MaxKellermann:
A good PR never needs text. Add descriptions to commit messages instead of ever writing text in a PR.
I like that MPD's event loop is designed in a way which allows omitting `epoll_ctl()` calls.
Currently, there is just one code location in MPD which forks, and that's the known-broken experimental "pipe" output plugin which nobody should ever use.
If anybody ever adds more `fork()` calls to MPD, he shall properly handle synchronization. File descriptors get opened and closed all the time, but forking is a very rare event, and that's not something we should optimize for.

arcnmx:
> A good PR never needs text. Add descriptions to commit messages instead of ever writing text in a PR.

The text is to initiate a discussion, so I can know how to proceed with the PR and how to handle fork() in general. It doesn't seem appropriate to ask questions in a commit message?

> I like that MPD's event loop is designed in a way which allows omitting epoll_ctl() calls.

Does that mean that it should always do so? In order words, does this PR's change seem reasonable or unnecessary?

> If anybody ever adds more fork() calls to MPD, he shall properly handle synchronization

#223 introduces a fork+exec in a plugin, and the introduction of `Close` + `AbandonFD` revealed the crash above. So would it be more appropriate to block on fork and wait for the exec before allowing the event loop to continue? That seems doable.

MaxKellermann:
> Does that mean that it should always do so? In order words, does this PR's change seem reasonable or unnecessary?

There's no absolute truth, it's an opinion. Your opinion to change this is reasonable, because it is a proper fix to a real problem. But it's not the path I'd like to choose.

> #223 introduces a fork+exec in a plugin, and the introduction of `Close` + `AbandonFD` revealed the crash above. So would it be more appropriate to block on fork and wait for the exec before allowing the event loop to continue? That seems doable.

Yes, because forks are a very rare event. Forking is dangerous, and several things need to be done to do it safely (e.g. restore the signal block mask). Synchronizing with the epoll_wait() calls is just one more thing.

arcnmx:
> There's no absolute truth, it's an opinion. Your opinion to change this is reasonable, because it is a proper fix to a real problem. But it's not the path I'd like to choose.

Well, up to you. I'd call it an enhancement rather than a fix given that `AbandonFD` is considered a feature - even with this patch applied a plugin's `fork()` can't rely on the fact that `AbandonFD` won't still be used somewhere else. Although it can't be relied on as a fix, it still seems nicer to be explicit about FD removal where possible.

> Yes, because forks are a very rare event. Forking is dangerous, and several things need to be done to do it safely (e.g. restore the signal block mask). Synchronizing with the epoll_wait() calls is just one more thing.

Hm so I guess this might need to be some sort of global mutex as part of the epoll backend? Since there seem to be multiple (at least 3?) different event loops running, and all of them need to be blocked at the same time.

MaxKellermann:
You are right, this is difficult to solve when there are multiple EventLoop instances in different threads.
I merged a modified version of your commit: 0a81e462dbe8dd27826280031c2e832bba8d96a1
And yes please, next time, add the whole explanation to the commit message. No explanations in GitHub comments - these will be lost after merging.

Issue #967, opened 2020-09-29
Some audio files does not play under Windows WASAPI exclusive mode


parkmino:
### Describe the bug
MPD does not play mp3 and flac 96KHz files under WASAPI exclusive mode while there is no problem with flac 44.1KHz.
Connection is lost or showing the following error message.

`ERROR: Failed to open "wasapi" (wasapi); Format is not supported: AUDCLNT_E_UNSUPPORTED_FORMAT`

## Version

MPD 0.22 for Windows

## Log

[mpd.log](https://github.com/MusicPlayerDaemon/MPD/files/5300110/mpd.log)

MaxKellermann:
@ibmibmibm please have a look

ibmibmibm:
Seems that 96kHz is not supported by your audio card.

MaxKellermann:
That is a condition which should be handled by the output plugin, by forcing a supported sample rate!

ibmibmibm:
Ok, I'll make this change and open a pull request

parkmino:
FYI, my audio card supports up to 192KHz.
It is strange not to play mp3.

ibmibmibm:
@parkmino Could you try to play that file using this command:
```
mpv [musicfile] --msg-level=ao/wasapi=v --audio-exclusive --no-audio-display
```
with [mpv](https://github.com/mpv-player/mpv) for windows, and collect the logs for me? Thanks.

parkmino:
@ibmibmibm Please find the attached file.

[mpv.log](https://github.com/MusicPlayerDaemon/MPD/files/5576976/mpv.log)

MaxKellermann:
da642b289049836f2d89b65726ef1aa5229b5f82

PR #1014, opened 2020-11-26
src/output: add algorithm for finding usable AudioFormat

ibmibmibm:
* Add ComWorker to run all COM relative function in same thread.
* Add Interrupt interface.
* Reopen device on error.
* Use WinEvent for as a condition_variable without lock.

See https://github.com/MusicPlayerDaemon/MPD/issues/967

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -261,6 +261,7 @@ sources = [
   'src/LogInit.cxx',
   'src/ls.cxx',
   'src/Instance.cxx',
+  'src/win32/ComWorker.cxx',
   'src/win32/Win32Main.cxx',
   'src/MusicBuffer.cxx',
   'src/MusicPipe.cxx',
--- a/src/mixer/plugins/WasapiMixerPlugin.cxx
+++ b/src/mixer/plugins/WasapiMixerPlugin.cxx
@@ -19,7 +19,7 @@
 
 #include "mixer/MixerInternal.hxx"
 #include "output/plugins/WasapiOutputPlugin.hxx"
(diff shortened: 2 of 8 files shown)


MaxKellermann:
Can you split this commit, so I can cherry-pick the fix for the stable branch?
I wouldn't like to merge such a huge commit with so many major possibly-breaking changes into the stable branch.

MaxKellermann:
Sorry for letting you wait for so long, and thanks for putting so much work into splitting the PR.
I had difficulties with several of your commits because my GCC-mingw build doesn't have threading enabled; this works in GCC only through pthread emulation. Therefore, std::future/std::promise is unavailable, as well as std::mutex (which is why I've been carrying the `using Mutex = CriticalSection` around for so long).
I wish GCC had proper C++ standard library support for native Win32 threading... but all these years, that didn't happen.

I don't want to use pthread emulation for MPD/Windows builds, so I'll try to figure out how to do it without those C++ threading features.

And I'll try to backport the format detection code to the stable branch without the threading changes. This is the most important change here.

Meanwhile, I've cherry-picked a few trivial commits. It would be great if you could keep your PR branch rebased on my master branch.

ibmibmibm:
I've implemented std::future and std::promise for mingw32 without pthread.
The implements are inspired by [microsoft stl](https://github.com/microsoft/STL/blob/main/stl/inc/future)

MaxKellermann:
Cherry-picked to v0.22 with some cleanups
da642b289049836f2d89b65726ef1aa5229b5f82

PR #1018, opened 2020-11-29
lib/ffmpeg/Filter: Add define required for avutil

DanySpin97:
Same problem reported in #861 , fixed in a different way.
Note that:

- `#define  __STDC_CONSTANT_MACROS` is already used among mpd codebase.
- this definition can be replaced by including `stdint.h` (or `cstdint`), because internally it includes this define (feel free to go ahead and read stdint.h source code under `/usr/include/c++/<gcc_ver>/cstdint`)

This is a mpd bug and I am looking forward to have it fixed :)

### DIFF
--- a/src/lib/ffmpeg/Filter.hxx
+++ b/src/lib/ffmpeg/Filter.hxx
@@ -20,6 +20,9 @@
 #ifndef MPD_FFMPEG_FILTER_HXX
 #define MPD_FFMPEG_FILTER_HXX
 
+/* necessary because libavutil/common.h uses UINT64_C */
+#define __STDC_CONSTANT_MACROS
+
 #include "Error.hxx"
 
 extern "C" {


MaxKellermann:
> This is a mpd bug and I am looking forward to have it fixed :)

This is not a MPD bug. This is a FFmpeg bug. FFmpeg's headers fail to compile because they did not include the necessary headers.
Your code is just a workaround for this FFmpeg bug.

DanySpin97:
> FFmpeg's headers fail to compile because they did not include the necessary headers.

They are telling you to either include stdint.h or add that define. By failing to comply, mpd fails to build.

Your argument doesn't make much sense since that define is literally used across all lib/ffmpeg sources _inside_ mpd. Please decide what you exactly want and follow that.

MaxKellermann:
> They are telling you to either include stdint.h or add that define.

Citation needed.

> Your argument doesn't make much sense since that define is literally used across all lib/ffmpeg sources _inside_ mpd. Please decide what you exactly want and follow that.

So you say that my assertion that this isn't a MPD bug but a FFmpeg bug is wrong because the same workaround exists already in other MPD sources? *This* doesn't make any sense.

MaxKellermann:
92bfdffa42304fb11663c30c2870cc0dfa205a3e

DanySpin97:
> So you say that my assertion that this isn't a MPD bug but a FFmpeg bug is wrong because the same workaround exists already in other MPD sources?

I am saying that, bug or not, it wouldn't make much sense to refuse the this one considering ` __STDC_CONSTANT_MACROS` is already defined inside other sources.

From https://stackoverflow.com/a/16194477:
> UINT64_C macro is defined in standard C header stdint.h, but only for non-cplusplus compiled code.

My guess is that they want C++ applications to manually define that.

[Here](https://stackoverflow.com/a/50764569) in a minimal FFmpeg example that define is used. And almost every project that uses FFmpeg has it too. Personally, I think it is not a bug in FFmpeg, but still it is a really bad choice for library users. Feel free to disagree with me, I am not an expert neither a FFmpeg developer.

I just think that correctly building the software is more important than deciding which fault is it, so thanks for accepting this MR :)

I am also asking on #ffmpeg IRC channel to get some details about this.

DanySpin97:
Upstream issue: http://ffmpeg.org/pipermail/ffmpeg-devel/2020-November/273181.html

Issue #430, opened 2018-11-29
'tidal' And 'Qobuz' are unavailable


LinArcX:
Hi. i want to use ncmpcpp alongside mdp. follow instructions in:
https://wiki.archlinux.org/index.php/Music_Player_Daemon
https://wiki.archlinux.org/index.php/Ncmpcpp
everything seems fine, but when i try to update list of musics in ncmpcpp, these logs appear:

```
» mpd --stdout --no-daemon --verbose
config_file: loading file /home/linarcx/.config/mpd/mpd.conf
libsamplerate: libsamplerate converter 'Fastest Sinc Interpolator'
vorbis: Xiph.Org libVorbis 1.3.6
opus: libopus 1.3
sndfile: libsndfile-1.0.28
hybrid_dsd: The Hybrid DSD decoder is disabled because it was not explicitly enabled                                                 
wildmidi: configuration file does not exist: /etc/timidity/timidity.cfg                                                              
simple_db: reading DB
output: No 'audio_output' defined in config file
output: Attempt to detect audio output device
output: Attempting to detect a alsa audio device
output: Successfully detected a alsa audio device
exception: Input plugin 'tidal' is unavailable: No Tidal application token configured                                                
exception: Input plugin 'qobuz' is unavailable: No Qobuz app_id configured                                                           
curl: version 7.62.0
curl: with OpenSSL/1.1.1
avahi: Initializing interface
exception: RTIOThread could not get realtime scheduling, continuing anyway: sched_setscheduler failed: Operation not permitted       
avahi: Client changed to state 101
avahi: Client is CONNECTING
inotify: initializing inotify
inotify: watching music directory
client: [0] opened from [::1]:57724
client: [0] process command "status"
exception: Failed to read mixer for 'My ALSA Device': no such mixer control: PCM
client: [0] command returned 0
client: [0] process command "plchanges "0""
client: [0] command returned 0
client: [0] process command "outputs"
client: [0] command returned 0
client: [0] process command "decoders"
client: [0] command returned 0
client: [0] process command "outputs"
client: [0] command returned 0
client: [0] process command "outputs"
client: [0] command returned 0
client: [0] process command "idle"
client: [0] command returned 1
client: [0] process command "update "/""
update: spawned thread for update job id 1
client: [0] command returned 0
update: starting
client: [0] process command "idle"
client: [0] command returned 1
client: [0] process command "status"
client: [0] command returned 0
client: [0] process command "idle"
[...]

PR #1026, opened 2020-12-13
Document that all plugins are enabled by default

gwerbin:
This was surprising behavior for me and at least a few other users in Issue #430. Documenting it will help reduce confusion by users.


MaxKellermann:
This text is kind of redundant, as the "enabled" flag is already documented here: https://github.com/MusicPlayerDaemon/MPD/blob/master/doc/user.rst#configuring-input-plugins

Issue #1039, opened 2021-01-03
Webdav not working with Nextcloud


PVince81:
### Describe the bug
1. Set music_directory to a Webdav URL connecting to a Nextcloud instance: "https://mpd:password@nextcloud.hostname.com/remote.php/dav/files/mpd/Music/"
2. Restart mpd service
3. Run `mpc rescan`
4. Observe Nextcloud's Apache access log

## Expected Behavior
Scanning process goes through all directories.

## Actual Behavior
Only one PROPFIND is made and then the scanning stops.

## Version
mpd 0.22.3 (from Arch ARM packages on Rpi1 model B)

### mpd
TODO

### Apache access log
```
xx.xx.xx.xx - - [03/Jan/2021:17:52:00 +0100] "PROPFIND /remote.php/dav/files/mpd/Music/ HTTP/1.1" 401 5509 "-" "Music Player Daemon 0.22.3"
xx.xx.xx.xx - mpd [03/Jan/2021:17:52:00 +0100] "PROPFIND /remote.php/dav/files/mpd/Music/ HTTP/1.1" 207 1883 "-" "Music Player Daemon 0.22.3"
```
First request is from auth, the second one is the authenticated one.

## Notes
I looked at the source code and I believe that the Webdav response processing is incomplete: https://github.com/MusicPlayerDaemon/MPD/blob/abbd980671944bc2d7a043a8481f4168ce878ae7/src/storage/plugins/CurlStorage.cxx#L436

A successful PROPFIND response returns the HTTP status code 207, not 200. So probably the related code paths should check for more 2xx status codes.

PVince81:
Next up:
- [ ] test with master and provide logs
- [ ] check Nextcloud logs / find out if the request is constructed properly
- [ ] try replacing 200 with 207 and see if it works

PVince81:
wow, compiling mpd on the Rpi 1 Model B took more than 3 hours... and it seems that rerunning `ninja -C output/release` starts from scratch, which means that for every code change attempt I'll need a few hours

PVince81:
okay, got the verbose mode output from mpd after compiling from master: when trying to scan Webdav, it says something like:
```
update: Not a directory: https://mpd:password@nextcloud.hostname.com/remote.php/dav/files/mpd/Music/
```

so need to check the exact request and how the response is processed

PVince81:
Note: could also be a namespace parsing issue when parsing the collection tag

PVince81:
[...]

PR #1040, opened 2021-01-04
Fixed Webdav storage scanning to be more RFC compliant

PVince81:
### Changes

- Removed additional "a:prop" in PROPFIND request to match [RFC 4918 section 9.1.3](https://tools.ietf.org/html/rfc4918#section-9.1.3). (bug)
- Added Content-Type header as the body is not a true multipart POST.
- Fixed parsing propstat so that the correct propstat block is used when extracting relevant property values
- Added support for scheme-less URI in response href

### Ticket

Fixes https://github.com/MusicPlayerDaemon/MPD/issues/1039

### Motivation

The above were needed to make scanning work with Nextcloud which implements Webdav in a stricter way.

### Tests

Tested against a Nextcloud 20 instance which also contained file names with Chinese characters.
Scanning worked in deeper directory trees and music could be played.

### Notes

Some of the code might be slightly clumsy as it's my first time with C++11
and due to my struggles with strings I didn't try and add further hardening in the code (like in case of accidental double slashes)

I've only tested this with Nextcloud, not sure what Webdav server was used when testing the original implementation. I didn't find anything related in .travis.yml, might be good to add some scanning tests with various storage backends at some point :sweat_smile:

### DIFF
--- a/src/storage/plugins/CurlStorage.cxx
+++ b/src/storage/plugins/CurlStorage.cxx
@@ -516,7 +516,6 @@ private:
 	 */
 	gcc_pure
 	StringView HrefToEscapedName(const char *href) const noexcept {
-		StringView relative_path;
 		StringView path = uri_get_path(href);
 		if (path == nullptr)
 			return nullptr;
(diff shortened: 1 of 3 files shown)


src/storage/plugins/CurlStorage.cxx
```diff
@@ -493,24 +516,31 @@ class HttpListDirectoryOperation final : public PropfindOperation {
...
 		if (path == nullptr)
 			return nullptr;
 
 		/* kludge: ignoring case in this comparison to avoid
 		   false negatives if the web server uses a different
 		   case */
-		path = StringAfterPrefixIgnoreCase(path, base_path.c_str());
-		if (path == nullptr || path.empty())
+		relative_path = StringAfterPrefixIgnoreCase(path, base_path.c_str());
+		if (relative_path == nullptr || relative_path.empty()) {
```
PVince81, line 528: at first I wanted to use `uri_is_relative` on `path` but I didn't find how to convert a StringView to a c_str as it doesn't have a c_str method...

MaxKellermann: Maybe we should migrate `uri_is_relative()` to take a `StringView`?
`StringView` / `std::string_view` are not null-terminated, so there cannot be a `.c_str()` method.
A kludge for now could be to cast to `std::string` and call its `.c_str()` - which would duplicate the string.

PVince81: I tried with `uri_is_relative` but it wouldn't work because those Webdav hrefs do start with "/".

So I migrated and used `uri_has_scheme` instead to decide what base path to check against.
See 74b2fc7fdca9be13cbbe4cb52b2fab573b3cf82c

MaxKellermann: Of course... if it starts with a slash, it's not relative, it's an absolute URI path.

MaxKellermann:
Good work, thanks!

PVince81:
A mistake slipped in, see https://github.com/MusicPlayerDaemon/MPD/issues/1043

PR #1047, opened 2021-01-13
WIP: Continue work on input_cache

TilCreator:
(This pr is mostly for seeing what this fork is doing since that is hard to see on github.)

My fixes are mostly wip, see the TODOs for further information.
In my tests I was able to use an unstable sshfs connection (make shure to use the timeout options so that the fs calls actually fail).
I also tested with curl / webdav (Remove [/src/input/cache/Manager.cxx:97-98](https://github.com/TilCreator/MPD/blob/695a0d851c226aeed836fcb113734219d89f5704/src/input/cache/Manager.cxx#L97-L98) for that), but that somehow still fails even if cached, didn't look further into it.

I will continue working on this, but probably not fast. So help with the TODOs (also ideas how to fix them) and more are welcome.

### DIFF
--- a/src/Partition.cxx
+++ b/src/Partition.cxx
@@ -80,7 +80,7 @@ PrefetchSong(InputCacheManager &cache, const char *uri) noexcept
 static void
 PrefetchSong(InputCacheManager &cache, const DetachedSong &song) noexcept
 {
-	PrefetchSong(cache, song.GetURI());
+	PrefetchSong(cache, song.GetRealURI());
 }
 
 inline void
--- a/src/Partition.cxx
+++ b/src/Partition.cxx
@@ -91,7 +91,7 @@ Partition::PrefetchQueue() noexcept
 
 	auto &cache = *instance.input_cache;
 
-	int next = playlist.GetNextPosition();
+	int next = playlist.GetNextPosition(1);
 	if (next >= 0)
 		PrefetchSong(cache, playlist.queue.Get(next));
 
--- a/src/Partition.cxx
+++ b/src/Partition.cxx
@@ -91,11 +91,14 @@ Partition::PrefetchQueue() noexcept
 
 	auto &cache = *instance.input_cache;
 
-	int next = playlist.GetNextPosition(1);
-	if (next >= 0)
-		PrefetchSong(cache, playlist.queue.Get(next));
-
-	// TODO: prefetch more songs
+	// TODO: Find better way to guess how many songs to try to prefetch, this implementation could flush a cached file that is needed earlier when the most recent cached file.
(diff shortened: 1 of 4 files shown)


src/queue/Playlist.hxx
```diff
@@ -114,7 +114,7 @@ struct playlist {
 	int GetCurrentPosition() const noexcept;
 
 	gcc_pure
-	int GetNextPosition() const noexcept;
+	int GetNextPosition(unsigned index) const noexcept;
```
MaxKellermann, line 117: Please document this - the method is currently completely undocumented, that's bad already (sorry...)...

src/queue/Playlist.cxx
```diff
@@ -343,15 +343,15 @@ playlist::GetCurrentPosition() const noexcept
...
 		return -1;
 
 	if (queue.single != SingleMode::OFF && queue.repeat)
 		return queue.OrderToPosition(current);
-	else if (queue.IsValidOrder(current + 1))
-		return queue.OrderToPosition(current + 1);
+	else if (queue.IsValidOrder(current + index))
+		return queue.OrderToPosition(current + index);
 	else if (queue.repeat)
 		return queue.OrderToPosition(0);
```
MaxKellermann, line 356: Incomplete: in repeat mode, this should properly calculate the overflow

src/queue/Playlist.hxx
MaxKellermann, line 117: Can we have `index=1` as default value, so existing callers don't need to be migrated?
Maybe rename the parameter to `offset`?

src/Partition.cxx
```diff
@@ -80,7 +80,7 @@ PrefetchSong(InputCacheManager &cache, const char *uri) noexcept
 static void
 PrefetchSong(InputCacheManager &cache, const DetachedSong &song) noexcept
 {
-	PrefetchSong(cache, song.GetURI());
+	PrefetchSong(cache, song.GetRealURI());
```
MaxKellermann, line 83: The commit message isn't good - it doesn't explain what needs to be fixed, and why this change fixes the issue.

src/Partition.cxx
```diff
@@ -225,6 +228,7 @@ Partition::OnIdleMonitor(unsigned mask) noexcept
 void
 Partition::OnGlobalEvent(unsigned mask) noexcept
 {
+	// TODO This doesn't seam to trigger when the random property of the queue is changed, this is needed for the pre caching
```
MaxKellermann, line 231: A "sync with player" is only needed if the one next song changes; the player doesn't care for anything but the current and the next song. And the next song only if it has already finished decoding the current one; then it will start decoding the next one. That's why.

PR #1051, opened 2021-01-17
clang-tidy: convert C casts to C++

neheb:
Found with google-readability-casting.


MaxKellermann:
I don't see a point in this change. This is called "-readability-" but by being more verbose, it reduces readability.

neheb:
Right. The pointer stuff has to use the full name. The single type ones can do x()

PR #1057, opened 2021-01-20
queue/PlaylistEdit: MoveRange fix for large end

iyefrat:
"playlist::MoveRange" now sets "end" to be the playlist length if it's
larger than that (previously threw an error).
Fixes commands like "move 5: 1" failing due to partial ranges getting
parsed as having "end" be the numerical limit.

### DIFF
--- a/src/queue/PlaylistEdit.cxx
+++ b/src/queue/PlaylistEdit.cxx
@@ -323,7 +323,8 @@ void
 playlist::MoveRange(PlayerControl &pc,
 		    unsigned start, unsigned end, int to)
 {
-	if (!queue.IsValidPosition(start) || !queue.IsValidPosition(end - 1))
+	end = GetLength() < end ? GetLength() : end;
+	if (!queue.IsValidPosition(start))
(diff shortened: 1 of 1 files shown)


MaxKellermann:
This would not only allow open-ended ranges, but arbitrary but invalid end positions. So that solution is too broad.

iyefrat:
Do you mean about end positions that are larger than the playlist length? Because [the delete command](https://github.com/MusicPlayerDaemon/MPD/blob/594dfe572b72f780428a5305fc702c025f29fc93/src/queue/PlaylistEdit.cxx#L280) also seems to normalize end positions that are too large to the end of the playlist.

MaxKellermann:
For moving songs, this should be exact, because if another client has changed the queue meanwhile, you might be moving different songs than you intended. So rather get an error because the song range has just become invalid than doing unintended things.
For delete, one could argue that if some of the songs had already been deleted, MPD should just silently delete the others (and not complain). Today, I wouldn't argue that, but I guess that code was written by me long ago.

PR #1068, opened 2021-02-01
src/tag/meson.build: fix GenParseName linking

ffontaine:
Set link_language to cpp for GenParseName to avoid the following build failure due to gcc being used to link a C and C++ source file:

```
FAILED: src/tag/GenParseName
/usr/bin/gcc  -o src/tag/GenParseName src/tag/GenParseName.p/GenParseName.cxx.o src/tag/GenParseName.p/Names.c.o -Wl,--as-needed -Wl,--no-undefined -Wl,-O1
/usr/bin/ld: src/tag/GenParseName.p/GenParseName.cxx.o: in function `std::_Rb_tree<std::basic_string_view<char, std::char_traits<char> >, std::pair<std::basic_string_view<char, std::char_traits<char> > const, TagType>, std::_Select1st<std::pair<std::basic_string_view<char, std::char_traits<char> > const, TagType> >, std::less<std::basic_string_view<char, std::char_traits<char> > >, std::allocator<std::pair<std::basic_string_view<char, std::char_traits<char> > const, TagType> > >::_M_erase(std::_Rb_tree_node<std::pair<std::basic_string_view<char, std::char_traits<char> > const, TagType> >*)':
GenParseName.cxx:
(.text._ZNSt8_Rb_treeISt17basic_string_viewIcSt11char_traitsIcEESt4pairIKS3_7TagTypeESt10_Select1stIS7_ESt4lessIS3_ESaIS7_EE8_M_eraseEPSt13_Rb_tree_nodeIS7_E[_ZNSt8_Rb_treeISt17basic_string_viewIcSt11char_traitsIcEESt4pairIKS3_7TagTypeESt10_Select1stIS7_ESt4lessIS3_ESaIS7_EE8_M_eraseEPSt13_Rb_tree_nodeIS7_E]+0x23): undefined reference to `operator delete(void*)'
```

Fixes:
 - http://autobuild.buildroot.org/results/871e1362c44e5b68a149e6a5dd3caf99ea0d904a

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/tag/meson.build
+++ b/src/tag/meson.build
@@ -2,6 +2,7 @@ generate_parse_name = executable(
   'GenParseName',
   'GenParseName.cxx',
   'Names.c',
+  link_language: 'cpp',
   native: true,
 )
 


MaxKellermann:
Why is this necessary? I expect Meson to figure out the language, because it should already know it's C++.
Is this a workaround for a Meson bug? If yes, which one?
We can't just add workarounds for bugs in third-party software without documenting it properly.

ffontaine:
I don't know if this a "known" meson bug, the only thing that I know is that it is raised when cross-compiling mpd on buildroot with latest meson (i.e. "0.56.2").

MaxKellermann:
I think you should report it to the Meson project first. If they say it's really a bug in MPD's `meson.build`, they can explain. But right now, I don't understand why this should be necessary.

MaxKellermann:
Closing until somebody explains why this is a MPD problem, and not a Meson bug.

joerg-krause:
I took a closer look at this issue. It's neither a MPD nor a Meson bug. In fact it is a Buildroot issue passing the wrong compiler to Meson:

```
Using 'CXX_FOR_BUILD' from environment with value: '/usr/bin/gcc'
C++ compiler for the build machine: /usr/bin/gcc (gcc 8.3.0 "gcc (Debian 8.3.0-6) 8.3.0")
C++ linker for the build machine: /usr/bin/gcc ld.bfd 2.31.1
```

This should be `/usr/bin/g++` of course.

ffontaine:
Thanks @joerg-krause, will you send a patch to buildroot?

joerg-krause:
> Thanks @joerg-krause, will you send a patch to buildroot?

@ffontaine Yes, I'm on it :smile:

MaxKellermann:
Thanks @joerg-krause!

ffontaine:
Yes, thanks a lot @joerg-krause, we'll be able to remove one of our patch on mpd. The only one left is the patch that adds atomic dependency to fix build failures on sparc or arm toolchains (https://git.buildroot.net/buildroot/tree/package/mpd/0001-src-event-meson.build-add-atomic-dependency-for-spar.patch).

PR #1072, opened 2021-02-03
Feature - add cacert for curl plugin

AndriiZ:
Currently it is not possible to set cacert for curl plugin, thus this setting may be hard to find. 

In my case, neither ~/.curlrc nor /etc/curlrc does not affect MPD curl plugin and https streams fail.
With this option, I have set cacert option in mpd.conf and https stream plays properly

### DIFF
--- a/src/input/plugins/CurlInputPlugin.cxx
+++ b/src/input/plugins/CurlInputPlugin.cxx
@@ -148,6 +148,8 @@ static struct curl_slist *http_200_aliases;
 /** HTTP proxy settings */
 static const char *proxy, *proxy_user, *proxy_password;
 static unsigned proxy_port;
+/** CA CERT settings*/
+static const char *cacert;
 
 static bool verify_peer, verify_host;
 
@@ -309,7 +311,7 @@ CurlInputStream::OnData(ConstBuffer<void> data)
 
 	if (data.size > GetBufferSpace()) {
 		AsyncInputStream::Pause();
-		throw CurlResponseHandler::Pause{};
+		throw CurlRequest::Pause();
 	}
 
 	AppendToBuffer(data.data, data.size);
@@ -378,6 +380,7 @@ input_curl_init(EventLoop &event_loop, const ConfigBlock &block)
 
 	verify_peer = block.GetBlockValue("verify_peer", default_verify);
 	verify_host = block.GetBlockValue("verify_host", default_verify);
+        cacert = block.GetBlockValue("cacert");
 }
 
 static void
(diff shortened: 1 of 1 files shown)


src/input/plugins/CurlInputPlugin.cxx
```diff
@@ -309,7 +311,7 @@ CurlInputStream::OnData(ConstBuffer<void> data)
 
 	if (data.size > GetBufferSpace()) {
 		AsyncInputStream::Pause();
-		throw CurlResponseHandler::Pause{};
+		throw CurlRequest::Pause();
```
MaxKellermann: Why this change?

src/input/plugins/CurlInputPlugin.cxx
```diff
@@ -378,6 +380,7 @@ input_curl_init(EventLoop &event_loop, const ConfigBlock &block)
 
 	verify_peer = block.GetBlockValue("verify_peer", default_verify);
 	verify_host = block.GetBlockValue("verify_host", default_verify);
+        cacert = block.GetBlockValue("cacert");
```
MaxKellermann: Bad indent

src/input/plugins/CurlInputPlugin.cxx
```diff
@@ -431,9 +434,11 @@ CurlInputStream::InitEasy()
 		request->SetOption(CURLOPT_PROXYUSERPWD,
 				   StringFormat<1024>("%s:%s", proxy_user,
 						      proxy_password).c_str());
+        if (cacert != nullptr)
+            request->SetOption(CURLOPT_CAINFO, cacert);
```
MaxKellermann, line 438: So CURL calls this CAINFO and you call it cacert... and it's undocumented!

src/input/plugins/CurlInputPlugin.cxx
```diff
@@ -431,9 +434,11 @@ CurlInputStream::InitEasy()
...
 
-	request->SetVerifyPeer(verify_peer);
-	request->SetVerifyHost(verify_host);
+	request->SetOption(CURLOPT_SSL_VERIFYPEER, verify_peer ? 1L : 0L);
+	request->SetOption(CURLOPT_SSL_VERIFYHOST, verify_host ? 2L : 0L);
```
MaxKellermann: Why this change?

MaxKellermann (changes requested):

src/input/plugins/CurlInputPlugin.cxx
AndriiZ: Sorry, started with 0.22.4, swithed to main, so there is this change in the last commit

src/input/plugins/CurlInputPlugin.cxx
AndriiZ: https://curl.se/libcurl/c/CURLOPT_CAINFO.html
CURLOPT_CAINFO - path to Certificate Authority (CA) bundle 
and from example 
curl_easy_setopt(curl, CURLOPT_CAINFO, "/etc/certs/cabundle.pem");

Do you think I should use another name? I have selected cacert because .curlrc has option cacert for CA bundle

src/input/plugins/CurlInputPlugin.cxx
AndriiZ: Sorry, started with 0.22.4, switched to main, so there is this change in the last commit

src/input/plugins/CurlInputPlugin.cxx
AndriiZ: Fixed

MaxKellermann:
If you fix mistakes in a commit that is not yet merged, amend the commit instead of adding fixup commits. I can't review a PR with bad commits plus fixup commits.

AndriiZ:
I'll create new one

PR #1073, opened 2021-02-03
Add add cacert for curl plugin

AndriiZ:
Currently it is not possible to set cacert for curl plugin, thus this setting may be hard to find.

In my case, neither ~/.curlrc nor /etc/curlrc does not affect MPD curl plugin and https streams fail.
With this option, I have set cacert option in mpd.conf and https stream plays properly

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,8 @@ ver 0.23 (not yet released)
   - new command "getvol"
 
 ver 0.22.5 (not yet released)
+* output
+  - httpd: error handling on Windows improved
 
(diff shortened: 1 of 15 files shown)


MaxKellermann:
Why did you create a new PR? The problem is not the PR, the problem is your local git branch (and the commits therein).
Also note that your PR contains text which is not present in any commit message. If you feel the need to add text to a PR, it only means that your commit messages are bad.

AndriiZ:
squashed into one

PR #1074, opened 2021-02-03
Add cacert for curl plugin 

AndriiZ:
* Add cacert option for Curl plugin. Allows to set cacert for curl lib in mpd.conf


MaxKellermann:
Documentation is still missing.

PR #1078, opened 2021-02-07
Add cacert option for curl plugin

AndriiZ:
* add cacert option for Curl plugin. Allows to set cacert for curl lib
* Added documentation line into mpdconf.example with explanation for cacert option

Co-authored-by: AndriiZ <github@zhuk.xyz>


MaxKellermann:
Documentation needs to be added to `doc/plugins.rst` (the plugin reference). That reference document must have a list of all options.
The file `doc/mpdconf.example` doesn't need an example of each and every option. This is an exotic options that only few users will ever need, so let's keep it out of there.

PR #1079, opened 2021-02-07
Fix: Separate Conductor from Performer

vkostas:
Conductor was incorrectly saved to Performer tag in MPD database


MaxKellermann:
Cherry-picked to the stable branch: a59f1b21a6b302e5d42b00e8b6f3276979b683b4

PR #1080, opened 2021-02-07
AdditionalTags

vkostas:
Additional standard ID3 tags (exist in both ID3v2.3/ID3v2.4):
BPM, ComposerSort, EncoderSettings, FileOwner, FileType, InitialKey, Lyricist, MediaType, OriginalAlbum, OriginalArtist, Publisher, SubTitle.
Some tags introduced in ID3v2.4 (e.g. SetSubTitle, Mood, MusicianCredits) or changed their internal code frame (e.g. InvolvedPeople) are not included. Can be added later.

### DIFF
--- a/src/tag/Id3Scan.cxx
+++ b/src/tag/Id3Scan.cxx
@@ -38,6 +38,10 @@
 #define ID3_FRAME_COMPOSER "TCOM"
 #endif
 
+#ifndef ID3_FRAME_COMPOSER_SORT
+#define ID3_FRAME_COMPOSER_SORT "TSOC"
+#endif
+
 #ifndef ID3_FRAME_DISC
 #define ID3_FRAME_DISC "TPOS"
 #endif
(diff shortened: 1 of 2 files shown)


MaxKellermann:
Your code doesn't even compile and the commit message is bad.
The tags you're using don't exist in MPD core set of tags. I havn't heard of any argument for adding those to the core set (because you didn't name any), so I'm not convinced of this idea.
MPD clients can query arbitrary tags with the `readcomments` command, so not each and every tag type needs to be part of the core set.

vkostas:
OK, Max. I am new to this and I am missing a lot of pieces (that's why I asked for some MPD design details in the forum).
I don't know anything about MPD design, nor the core set.
Even pull request would compile and work it could again be (fairly) rejected due to design decision, I am not aware of.
If someone could provide anything it will be great.
Until then, my contribution is useless and a waste of time (for me, you and others maybe).
I feel sorry, I cannot help further. :-)

MaxKellermann:
This is a list of tags supported by MPD: https://www.musicpd.org/doc/html/protocol.html#tags
Additional ID3 tags can only be supported if they are already in this list. You can propose to extend this list (but you didn't, so far).
As I wrote, I'm not convinced that this list needs to be extended, and nobody tried to convince me.

PR #1110, opened 2021-02-25
time/ISO8601: don't use glibc extension in strptime.

ericonr:
Per the manual for strptime, %F is equivalent %Y-%m-%d, so use that
directly.


MaxKellermann:
Cherry-picked to v0.22.x: 8d80280ab9ac9ec17abd981990771ea9927767a3

PR #1111, opened 2021-02-25
storage/plugins/CurlStorage: don't use glibc extension in

ericonr:
ParseTimePoint.

%Z is a glibc extension to strptime, and is a no-op there, due to the
mapping between timezone names and their definition (especially when the
name comes from a different machine) being ambiguous / impossible.  Time
in HTTP headers is guaranteed to be UTC.

Passing an unknown format to strptime() implementations that don't
support it will generally cause them to return NULL, which will lead to
ParseTimePoint throwing an exception and ParseTimeStamp using an
unnecessary fallback.

Since the timezone name goes at the end of the string, we don't need to
use %Z to skip it (could be an issue in a different time stamp format),
so simply removing %Z works best.

### DIFF
--- a/src/storage/plugins/CurlStorage.cxx
+++ b/src/storage/plugins/CurlStorage.cxx
@@ -193,7 +193,7 @@ ParseTimeStamp(const char *s)
 {
 	try {
 		// TODO: make this more robust
-		return ParseTimePoint(s, "%a, %d %b %Y %T %Z");
+		return ParseTimePoint(s, "%a, %d %b %Y %T");
 	} catch (...) {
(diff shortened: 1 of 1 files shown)


ericonr:
If the expectation was that `strptime` would convert the time based on the timezone name, that can be implemented (https://stackoverflow.com/questions/761791/converting-between-local-times-and-gmt-utc-in-c-c#761811), so let me know. Otherwise this patch doesn't change anything on glibc and makes this function not error out on musl and others.

MaxKellermann:
Cherry-picked to v0.22.x: c95e3dc065be71bc66dd3b7b3f7b208f541e9341

Issue #1119, opened 2021-03-09
cross-origin header support for http output


bitkeeper:
## Feature request
The current http output doesn't provide a header for cross-origin support. This prevents to use the mpd http stream directly from an other webapplication due the origin from the webpage differs from then the audio stream.

Example snippet of a html page that use an audiostream:
```
<html>
   <body>
        <audio src="http://yourip:8000" crossOrigin="anonymous"></audio>
    </body>
</html>
```

The current provided headers are:
```
Cache-Control: no-cache, no-store
Connection: close
Content-Type: audio/flac
Pragma: no-cache
```

**Required**
Adding support would require to add the following header to http response:
```
Access-Control-Allow-Origin: *
```

**Affected files:**

- "src/output/plugins/httpd/HttpdClient.cxx" function  `HttpdClient::SendResponse`
- "src/output/plugins/httpd/IcyMetaDataServer.cxx" function `icy_server_metadata_header`

**Current workarround:**
Setup proxy between web client and mpd by using other webserver like nginx.

MaxKellermann:
Should be a configurable setting; there's a reason browsers don't allow it by default. Implementing this is easy, why don't you submit a PR?

bitkeeper:
It isn't the security setting from the browser, but from the webserver. MPD in this case.  

In prevents cross posting of requests (serving request coming from pages that aren't served by the same server).
In the context of MPD it has no meaningful meaning to make it an option, it can allowed to be always on.

PR submit under way.

bitkeeper:
This issue is cloded because it is fixed by the commit in PR #1120 which is cherry picked into the 0.22.x branch.

PR #1120, opened 2021-03-09
Added cross-origin header to http headers of the http output.

bitkeeper:
The current http output doesn't provide a header for cross-origin support. This prevents to use the mpd http stream directly from an other webapplication due the origin from the webpage differs from then the audio stream.

The fix is to add the following header to the http response:
Access-Control-Allow-Origin: *

Related to #1119.


MaxKellermann:
Cherry-picked to v0.22.x: c08a8581eed767564d1afe359fdcf6ccbdb4d744

PR #1122, opened 2021-03-13
output/wasapi: use calculated new buffer instead old one

ibmibmibm:


MaxKellermann:
4bcdcca7f503064812a0da804d938af7c97d5b6c
Cherry-picked to v0.22.x

PR #1138, opened 2021-04-05
fix cross compilation with Debian 9

neheb:
In OpenWrt, some use Debian 9 to compile. Unfortunately, MPD requires
the host compiler to be GCC7 or above. This allows MPD to be cross
compiled with an old host.


MaxKellermann:
Not supported, and I don't want any such kludges.

Issue #293, opened 2018-06-10
replace libmodplug with libxmp


tidux:
https://github.com/cmatsuoka/libxmp is a library for playing tracker modules, similar to libmodplug, but more actively maintained and with much wider format support.  It already compiles everywhere mpd runs, including Android and Haiku.

MaxKellermann:
Why should libmodplug support be removed?

tidux:
Its functionality is a subset of libxmp's.  If mpd is going to link to one, it should be xmp.  I suppose it's possible to maintain both as options, but I can't think of any good reason to keep modplug over xmp.  Something like half my collection won't play with mpd+libmodplug, but will with libxmp based players.

asiekierka:
How does libopenmpt, which is a maintained descendant of libmodplug, fare?

MaxKellermann:
Any (free/open source) decoder library is acceptable. Writing a plugin for MPD is pretty easy ([look at this plugin, for example](https://github.com/MusicPlayerDaemon/MPD/blob/master/src/decoder/plugins/WildmidiDecoderPlugin.cxx)), and is the best way to get into MPD development. I hope there's a volunteer for this. I'd gladly help the volunteer, but I concentrate on other MPD things currently, so I won't be addressing this one for a while.

PR #1153, opened 2021-05-10
Add openmpt decoder plugin

GrimReaperFloof:
This PR adds the OpenMPT decoder plugin, based on libopenmpt.

Before you merge, I have some questions:
What does `python/build/libs.py` do? Is this a Windows thing where it downloads the libraries, or does it also apply to Unix-based systems?

There's a small amount of code re-use:
```cpp
	const auto buffer = mod_loadfile(nullptr, is);
	if (buffer.IsNull()) {
		LogWarning(openmpt_domain, "could not load stream");
		return false;
	}

	openmpt::module mod(buffer.data, buffer.size);
	delete[] buffer.data;
```
Is there a way to make a function out of it like how it's done in `ModplugDecoderPlugin.cxx`? When I tried I got into problems with returning `openmpt::module` iirc.

Also, are there any memory leaks or memory-related bugs in my code? I'm not that experienced in C/C++. While developing the OpenMPT decoder plugin, I did come across a segmentation fault. I debugged that with gdb and it was at line 155 and 160. After reading the libopenmpt documentation a bit more I figured libopenmpt apparently needs the count to be divided by two. Hence the `OPENMPT_FRAME_SIZE / 2 / sizeof(float)`. Is this fine, or is there any problem with that? Do I need to change anything to the `OPENMPT_FRAME_SIZE` variable? Perhaps integrating the math into it?

In the modplug decoder there's a check for a file being too large. Is that a limitation in libmodplug or is it a limitation in MPD? I left out the file size check in my OpenMPT decoder plugin. Do I need to add it?

Do you have any more suggestions for improving this code?

Also #293 might be somewhat relevant to this PR.

Also, it would be nice if there was a guide on implementing MPD plugins, so the API and how to go about doing things is a bit more clear.

### DIFF
--- a/doc/plugins.rst
+++ b/doc/plugins.rst
@@ -505,6 +505,8 @@ Module player based on `libopenmpt <https://lib.openmpt.org>`_.
 
    * - Setting
      - Description
+   * - **repeat_count**
+     - Set how many times the module repeats. -1: repeat forever. 0: play once, repeat zero times (the default). n>0: play once and repeat n times after that.
    * - **stereo_separation**
      - Sets the stereo separation. The supported value range is [0,200]. Defaults to 100.
    * - **interpolation_filter 0|1|2|4|8**
--- a/doc/plugins.rst
+++ b/doc/plugins.rst
@@ -520,7 +520,7 @@ Module player based on `libopenmpt <https://lib.openmpt.org>`_.
    * - **emulate_amiga yes|no**
      - Enables the Amiga resampler for Amiga modules. This emulates the sound characteristics of the Paula chip and overrides the selected interpolation filter. Non-Amiga module formats are not affected by this setting. Defaults to yes.
    * - **emulate_amiga_type**
-     - Configures the filter type to use for the Amiga resampler. Supported values are: "auto": Filter type is chosen by the library and might change. This is the default. "a500": Amiga A500 filter. "a1200": Amiga A1200 filter. "unfiltered": BLEP synthesis without model-specific filters. The LED filter is ignored by this setting. This filter mode is considered to be experimental and might change in the future. Defaults to "auto".
+     - Configures the filter type to use for the Amiga resampler. Supported values are: "auto": Filter type is chosen by the library and might change. This is the default. "a500": Amiga A500 filter. "a1200": Amiga A1200 filter. "unfiltered": BLEP synthesis without model-specific filters. The LED filter is ignored by this setting. This filter mode is considered to be experimental and might change in the future. Defaults to "auto". Requires libopenmpt 0.5 or higher.
 
 mpcdec
 ------
--- a/doc/plugins.rst
+++ b/doc/plugins.rst
@@ -494,6 +494,32 @@ Module player based on MODPlug.
    * - **loop_count**
      - Number of times to loop the module if it uses backward loops. Default is 0 which prevents looping. -1 loops forever.
 
+openmpt
+-------
+
+Module player based on `libopenmpt <https://lib.openmpt.org>`_.
+
+.. list-table::
+   :widths: 20 80
+   :header-rows: 1
+
+   * - Setting
+     - Description
+   * - **stereo_separation**
+     - Sets the stereo separation. The supported value range is [0,200]. Defaults to 100.
+   * - **interpolation_filter 0|1|2|4|8**
+     - Sets the interpolation filter. 0: internal default. 1: no interpolation (zero order hold). 2: linear interpolation. 4: cubic interpolation. 8: windowed sinc with 8 taps. Defaults to 0.
+   * - **override_mptm_interp_filter yes|no**
+     - If `interpolation_filter` has been changed, setting this to yes will force all MPTM modules to use that interpolation filter. If set to no, MPTM modules will play with their own interpolation filter regardless of the value of `interpolation_filter`. Defaults to no.
+   * - **volume_ramping**
+     - Sets the amount of volume ramping done by the libopenmpt mixer. The default value is -1, which indicates a recommended default value. The meaningful value range is [-1..10]. A value of 0 completely disables volume ramping. This might cause clicks in sound output. Higher values imply slower/softer volume ramps.
+   * - **sync_samples yes|no**
+     - Syncs sample playback when seeking. Defaults to yes.
+   * - **emulate_amiga yes|no**
+     - Enables the Amiga resampler for Amiga modules. This emulates the sound characteristics of the Paula chip and overrides the selected interpolation filter. Non-Amiga module formats are not affected by this setting. Defaults to yes.
+   * - **emulate_amiga_type**
+     - Configures the filter type to use for the Amiga resampler. Supported values are: "auto": Filter type is chosen by the library and might change. This is the default. "a500": Amiga A500 filter. "a1200": Amiga A1200 filter. "unfiltered": BLEP synthesis without model-specific filters. The LED filter is ignored by this setting. This filter mode is considered to be experimental and might change in the future. Defaults to "auto".
+
 mpcdec
 ------
 
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -138,6 +138,7 @@ option('gme', type: 'feature', description: 'Game Music Emulator decoder plugin'
 option('mad', type: 'feature', description: 'MP3 decoder using libmad')
 option('mikmod', type: 'feature', description: 'MikMod decoder plugin')
 option('modplug', type: 'feature', description: 'Modplug decoder plugin')
+option('openmpt', type: 'feature', description: 'OpenMPT decoder plugin')
 option('mpcdec', type: 'feature', description: 'Musepack decoder plugin')
 option('mpg123', type: 'feature', description: 'MP3 decoder using libmpg123')
 option('opus', type: 'feature', description: 'Opus decoder plugin')
--- a/python/build/libs.py
+++ b/python/build/libs.py
@@ -123,6 +123,15 @@ libmodplug = AutotoolsProject(
     ],
 )
 
+libopenmpt = AutotoolsProject(
+    'https://lib.openmpt.org/files/libopenmpt/src/libopenmpt-0.5.8+release.autotools.tar.gz',
+    '61de7cc0c011b10472ca16adcc123689',
+    'lib/libopenmpt.a',
+    [
+        '--disable-shared', '--enable-static'
+    ],
+)
+
 wildmidi = CmakeProject(
     'https://codeload.github.com/Mindwerks/wildmidi/tar.gz/wildmidi-0.4.3',
     '498e5a96455bb4b91b37188ad6dcb070824e92c44f5ed452b90adbaec8eef3c5',
--- a/src/decoder/DecoderList.cxx
+++ b/src/decoder/DecoderList.cxx
@@ -44,6 +44,7 @@
 #include "plugins/WildmidiDecoderPlugin.hxx"
 #include "plugins/MikmodDecoderPlugin.hxx"
 #include "plugins/ModplugDecoderPlugin.hxx"
+#include "plugins/OpenmptDecoderPlugin.hxx"
 #include "plugins/MpcdecDecoderPlugin.hxx"
 #include "plugins/FluidsynthDecoderPlugin.hxx"
 #include "plugins/SidplayDecoderPlugin.hxx"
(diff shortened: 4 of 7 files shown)


MaxKellermann:
> This PR adds the OpenMPT decoder plugin, based on libopenmpt.
> 
> Before you merge, I have some questions:
> What does `python/build/libs.py` do? Is this a Windows thing where it downloads the libraries, or does it also apply to Unix-based systems?

It's a kludge that compiles third-party libraries for Windows and Android. There are only declarations; to actually enable a library, add it to `win32/build.py` and `android/build.py`, but only if you believe the plugin is really useful for these platforms.

> Is there a way to make a function out of it like how it's done in `ModplugDecoderPlugin.cxx`? When I tried I got into problems with returning `openmpt::module` iirc.

What kind of problems?

I'll read your code later - but from a first glance, I saw that you copied `mod_loadfile()`, but I don't like code duplication. Maybe a first step would be to move this function to a separate source file, and export it.

> In the modplug decoder there's a check for a file being too large. Is that a limitation in libmodplug or is it a limitation in MPD? I left out the file size check in my OpenMPT decoder plugin. Do I need to add it?

It's a limitation of everybody's computer! Imagine you have a file weighing 100 Petabytes, and then imagine MPD asking the operating system to allocate this amount of RAM. There must be a limit. The current one is arbitrary.

> Also, it would be nice if there was a guide on implementing MPD plugins, so the API and how to go about doing things is a bit more clear.

... it would be nice if somebody wrote a guide after writing such a plugin!

GrimReaperFloof:
> It's a limitation of everybody's computer! Imagine you have a file weighing 100 Petabytes, and then imagine MPD asking the operating system to allocate this amount of RAM. There must be a limit. The current one is arbitrary.

Why add an arbitrary limit though? What if someone does in fact have petabytes of RAM?

MaxKellermann:
That's why the limit is low enough to avoid overloading anybody's computer, but still high enough that nobody has ever complained, ever.
An advanced approach to this problem could be memory-mapping the file, and let the kernel manage the allocations.

Answering your question "why" - because that's how this code was written, it proved to just work, nobody had a problem, and so there was no pain for anybody to engineer a better solution.

GrimReaperFloof:
Alright, I moved the `mod_loadfile()` function to a separate source file.

```cpp
	const auto buffer = mod_loadfile(nullptr, is);
	if (buffer.IsNull()) {
		LogWarning(openmpt_domain, "could not load stream");
		return false;
	}

	openmpt::module mod(buffer.data, buffer.size);
	delete[] buffer.data;
```
I was still unable to do something about the above code I was talking about earlier. If I want to put that into a function I might have to return a pointer, although when I tried, it gave me a warning and the openmpt decoder plugin broke after compiling. Probably because the pointer would become invalid as it leaves the scope. Maybe I need to use malloc? I'm not sure about how to get the size of openmpt modules in order to do that. I have never used malloc before.
But to be honest, this isn't so bad. It's only 6 lines of code and the return values slightly differ between the functions. In `ModplugDecoderPlugin.cxx` this code is offloaded to another function, but when calling that function from the other functions, it still needs an if statement to check if it returned a nullptr. Kinda similar to how it checks for `buffer.isNull` in my code. The only real thing a separate function for this would abstract away would be `delete[] buffer.data;`. So I don't think this is really that big of a deal. What do you think?

MaxKellermann:
I have no idea what you mean, and what broke.
No, you don't need `malloc()`. Never use it.
(Actually you should also never use `new` and `delete` - the existing Modplug plugin isn't so good. This code isn't exception-safe; it's always better to use smart pointers in C++. For example, MPD has the `AllocatedArray` class which automatically frees the allocated memory, and remembers the size. The LoadFile function could return `AllocatedArray<std::byte>`.)

GrimReaperFloof:
Is this alright? Did I use AllocatedArray correctly? Any more suggestions?

GrimReaperFloof:
Also is it possible to use DecoderPlugin.WithContainer with an InputStream instead of a Path?
Or should I just use Path if I want to use containers?

MaxKellermann:
> Also is it possible to use DecoderPlugin.WithContainer with an InputStream instead of a Path?
> Or should I just use Path if I want to use containers?

The API currently supports only local files. This should be improved (and you're welcome to help), but that's how it is currently.
(I'll review your code later - sorry for the delay, dayjob is nagging)

GrimReaperFloof:
Should I in the meantime prefer implementing InputStream and not support subsongs? Or should I implement Path and support subsongs?

GrimReaperFloof:
I have just pushed a version of the openmpt decoder plugin that uses Path and supports subsongs on the `add-openmpt-decoder_Path-version` branch on my MPD fork repository.

GrimReaperFloof:
Actually InputStream should be preferred yeah. Path can't handle files inside archives. After all, subsongs isn't used a lot in modules, and the majority of cases a subsong is found, it's a duplicate of the song or a small part of it. But I do know of some modules that make legitimate use of subsongs. But I don't think the positives of subsongs outweigh the negatives of using Path. I guess I'll implement subsongs when WithContainer supports InputStream.

MaxKellermann:
Doesn't compile:
```
../../src/decoder/plugins/OpenmptDecoderPlugin.cxx: In function ‘void mod_decode(DecoderClient&, InputStream&)’:
../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:86:6: error: ‘class openmpt::module’ has no member named ‘ctl_set_boolean’
   86 |  mod.ctl_set_boolean("seek.sync_samples", openmpt_sync_samples);
      |      ^~~~~~~~~~~~~~~
../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:87:6: error: ‘class openmpt::module’ has no member named ‘ctl_set_boolean’
   87 |  mod.ctl_set_boolean("render.resampler.emulate_amiga", openmpt_emulate_amiga);
      |      ^~~~~~~~~~~~~~~
../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:88:6: error: ‘class openmpt::module’ has no member named ‘ctl_set_text’
   88 |  mod.ctl_set_text("render.resampler.emulate_amiga_type", openmpt_emulate_amiga_type);
      |      ^~~~~~~~~~~~
```

MaxKellermann:
I've cherry-picked the last two patches (refactoring); please rebase your PR branch and resolve the conflicts.

GrimReaperFloof:
> Doesn't compile:
> 
> ```
> ../../src/decoder/plugins/OpenmptDecoderPlugin.cxx: In function ‘void mod_decode(DecoderClient&, InputStream&)’:
> ../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:86:6: error: ‘class openmpt::module’ has no member named ‘ctl_set_boolean’
>    86 |  mod.ctl_set_boolean("seek.sync_samples", openmpt_sync_samples);
>       |      ^~~~~~~~~~~~~~~
> ../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:87:6: error: ‘class openmpt::module’ has no member named ‘ctl_set_boolean’
>    87 |  mod.ctl_set_boolean("render.resampler.emulate_amiga", openmpt_emulate_amiga);
>       |      ^~~~~~~~~~~~~~~
> ../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:88:6: error: ‘class openmpt::module’ has no member named ‘ctl_set_text’
>    88 |  mod.ctl_set_text("render.resampler.emulate_amiga_type", openmpt_emulate_amiga_type);
>       |      ^~~~~~~~~~~~
> ```

Strange. What libopenmpt version did you use? I was using 0.5.8 while developing the plugin.

GrimReaperFloof:
I see those functions are only available since libopenmpt 0.5.0. Should I add version checks and call the appropriate functions?

MaxKellermann:
Yes, if you know a certain version is required, it should be part of the dependency check in the Meson file.

GrimReaperFloof:
I see some other decoder plugins like GME and Wildmidi have some version checking using the preprocessor. Do you have any idea how I could do this in libopenmpt? To get the library version you have to call `openmpt::get_library_version()`, but this is not a constant expression so I don't think I can use it in the preprocessor. I just tried this and it does compile but it never runs the code I want it to even if the version is over 0x50000. And if I just do an if-statement in normal C++, it won't compile with libopenmpt 0.4 like in your case. Do you have any idea what I can do? There seems to be a preprocessor variable called OPENMPT_STRING_LIBRARY_VERSION, but in the libopenmpt documentation it says this is deprecated. Is there something I can do with Meson perhaps?

GrimReaperFloof:
Okay so I just added some thing to the meson.build to set `HAVE_LIBOPENMPT_VERSION_0_5` when the version of libopenmpt is greater than 0.5, but there's a problem. It never executes the code for libopenmpt versions 0.5 and above.
I have checked the Meson manual and I believe I'm doing everything right here, yet it still doesn't work.
`libopenmpt_dep.version().version_compare('>= 0.5')` does return `True`. I have checked using a `message()` but for some reason it never runs the code I want it to.
Do you have any idea what's going on?

dgcampea:
> Okay so I just added some thing to the meson.build to set `HAVE_LIBOPENMPT_VERSION_0_5` when the version of libopenmpt is greater than 0.5, but there's a problem. It never executes the code for libopenmpt versions 0.5 and above.
> I have checked the Meson manual and I believe I'm doing everything right here, yet it still doesn't work.
> `libopenmpt_dep.version().version_compare('>= 0.5')` does return `True`. I have checked using a `message()` but for some reason it never runs the code I want it to when I build using libopenmpt 0.5.
> Do you have any idea what's going on?

Not a meson expert but you need to add `#include "decoder/Features.h"` to `src/decoder/plugins/OpenmptDecoderPlugin.cxx` according to <https://mesonbuild.com/Configuration.html#a-full-example>.

GrimReaperFloof:
@dgcampea Wow, that worked! Thanks! I have added the `decoder/Features.h` import and `git commit --amend`ed it as broken commits don't look very good in the history imo.

GrimReaperFloof:
That was a rebase

MaxKellermann:
Doesn't compile:
```../../src/decoder/plugins/OpenmptDecoderPlugin.cxx: In function ‘void mod_decode(DecoderClient&, InputStream&)’:
../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:99:70: error: passing ‘bool’ chooses ‘int’ over ‘unsigned int’ [-Werror=sign-promo]
   99 |  mod.ctl_set("seek.sync_samples", std::to_string(openmpt_sync_samples));
      |                                                                      ^
../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:99:70: error:   in call to ‘std::string std::__cxx11::to_string(int)’ [-Werror=sign-promo]
../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:100:84: error: passing ‘bool’ chooses ‘int’ over ‘unsigned int’ [-Werror=sign-promo]
  100 |  mod.ctl_set("render.resampler.emulate_amiga", std::to_string(openmpt_emulate_amiga));
      |                                                                                    ^
../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:100:84: error:   in call to ‘std::string std::__cxx11::to_string(int)’ [-Werror=sign-promo]
```
```

GrimReaperFloof:
> Doesn't compile:
> 
> ```
> ../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:99:70: error: passing ‘bool’ chooses ‘int’ over ‘unsigned int’ [-Werror=sign-promo]
>    99 |  mod.ctl_set("seek.sync_samples", std::to_string(openmpt_sync_samples));
>       |                                                                      ^
> ../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:99:70: error:   in call to ‘std::string std::__cxx11::to_string(int)’ [-Werror=sign-promo]
> ../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:100:84: error: passing ‘bool’ chooses ‘int’ over ‘unsigned int’ [-Werror=sign-promo]
>   100 |  mod.ctl_set("render.resampler.emulate_amiga", std::to_string(openmpt_emulate_amiga));
>       |                                                                                    ^
> ../../src/decoder/plugins/OpenmptDecoderPlugin.cxx:100:84: error:   in call to ‘std::string std::__cxx11::to_string(int)’ [-Werror=sign-promo]
> ```
Huh. I can reproduce it on libopenmpt 0.3 but for me they're only warnings, and compilation continues without further issues. How should I solve this anyway? Casting to unsinged int?

GrimReaperFloof:
Hopefully this fixes it. I don't see the warning anymore now.

PR #1157, opened 2021-05-15
flac encoder: enable Ogg chaining

jprjr:


jprjr:
Apologies for noise, realized I should have a case for just plain Ogg FLAC (without chaining), since that's often used in Icecast/Shoutcast applications.

PR #1169, opened 2021-05-23
Fix building under GCC 11

forkbomb9:
The build for v022.8 was failing due to -Werror=format-security tripping in GCC 11. This lint was already disabled with Clang, so I just removed the `#ifdef`s.

I don't know in which version GCC enabled this by default, but maybe we should add a check for  `GCC>=version_with_format-security` and `Clang `

### DIFF
--- a/src/util/RuntimeError.hxx
+++ b/src/util/RuntimeError.hxx
@@ -35,7 +35,7 @@
 
 #include <stdio.h>
 
-#ifdef __clang__
+#if defined(__clang__) || defined(__GNUC__)
 #pragma GCC diagnostic push
 // TODO: fix this warning properly
 #pragma GCC diagnostic ignored "-Wformat-security"
@@ -59,7 +59,7 @@ FormatInvalidArgument(const char *fmt, Args&&... args) noexcept
 	return std::invalid_argument(buffer);
 }
 
-#ifdef __clang__
+#if defined(__clang__) || defined(__GNUC__)
 #pragma GCC diagnostic pop
 #endif
 


MaxKellermann:
This should be used only if GCC or clang is used, not unconditionally.

forkbomb9:
Fixed. Now it's checking for `__clang__` or __GNUC__`.

Tested with Clang 11.1.0 and GCC 11.1.0.

MaxKellermann:
You don't need to check both macros, because clang defines `__GNUC__`. But doing both checks is ok for me.
What I don't understand is why you encounter this (bogus) warning at all. MPD's `meson.build` does not enable `-Wformat-security` (and it's not enabled implicitly by other warning options), nor does Meson itself. Did you enable it manually? If yes, then why?

MaxKellermann:
> I don't know in which version GCC enabled this by default

According to GCC documentation, this is not enabled by default in any GCC version: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
Are you using a patched unofficial GCC version?

forkbomb9:
I was building the mpd-git AUR package. And turns out the problem is that `makepkg.conf` enables `-Werror=format-security`: https://aur.archlinux.org/cgit/aur.git/tree/makepkg.conf?h=pacman-git#n42

So I'm not sure if this has to be fixed here or in the package's side.

As this was already disabled for Clang, I thought it could be also disabled for GCC, but maybe it isn't a wise decision?

MaxKellermann:
> As this was already disabled for Clang, I thought it could be also disabled for GCC, but maybe it isn't a wise decision?

In clang (unlike GCC), this option is implicitly enabled with `-Wformat`: https://clang.llvm.org/docs/DiagnosticsReference.html
Therefore this needed to be worked around.

But in GCC, it's not enabled implicitly by any other warning option. Here, no workaround is necessary. If Arch decides to enable additional warning options globally, and there are false positives, this should be fixed by Arch, not by MPD. This warning is bogus.

MaxKellermann:
btw. `--werror`/`-Werror` is an option for developers, not for users. Warnings should never be a build breakage for users. Warnings are just that - warnings. Developers should fix them, but users should not need to care.

MaxKellermann:
I've cherry-picked your commit to the stable branch v0.22.x: dbe12a6b9005df084d4fde776e00805cb0b14492
.. even though I think this is the wrong place to fix it. But since this is a library which is obsolete and will be replaced soon, I don't care much.

forkbomb9:
All right! Thanks for your time!

PR #1170, opened 2021-05-23
meson: set only sanitizers for fuzzer when unspecified

catenacyber:
That is when meson option b_sanitize is not used

cc @MaxKellermann
cf https://github.com/google/oss-fuzz/pull/5812

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -118,7 +118,10 @@ if get_option('buildtype') != 'debug'
 endif
 
 if get_option('fuzzer')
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Cherry-picked to the stable branch v0.22.x: d3a680cc87ebeea8c54f99a74c521048fdf8e8a9

catenacyber:
Thanks, do you know when it will be merged to master ?

PR #1174, opened 2021-05-27
windows fixes

neheb:
The last one is only a partial fix.

### DIFF
--- a/src/input/plugins/ArchiveInputPlugin.cxx
+++ b/src/input/plugins/ArchiveInputPlugin.cxx
@@ -23,6 +23,7 @@
 #include "archive/ArchiveFile.hxx"
 #include "../InputStream.hxx"
 #include "fs/LookupFile.hxx"
+#include "fs/NarrowPath.hxx"
 #include "fs/Path.hxx"
 #include "Log.hxx"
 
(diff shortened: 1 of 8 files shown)


src/input/plugins/ArchiveInputPlugin.cxx
```diff
@@ -54,5 +55,5 @@ OpenArchiveInputStream(Path path, Mutex &mutex)
 	}
 
 	return archive_file_open(arplug, l.archive)
-		->OpenStream(l.inside.c_str(), mutex);
+		->OpenStream(NarrowPath(l.inside), mutex);
```
MaxKellermann, line 58: What makes you believe the Windows ANSI code page is the correct one here?

MaxKellermann:
Did you compile and test all those changes on Windows?

src/input/plugins/CdioParanoiaInputPlugin.cxx
```diff
@@ -192,7 +193,7 @@ input_cdio_open(const char *uri,
 		throw std::runtime_error("Unable find or access a CD-ROM drive with an audio CD in it.");
 
 	/* Found such a CD-ROM with a CD-DA loaded. Use the first drive in the list. */
-	const auto cdio = cdio_open(device.c_str(), DRIVER_UNKNOWN);
+	const auto cdio = cdio_open(NarrowPath(device), DRIVER_UNKNOWN);
```
MaxKellermann, line 196: This change does not make sense. Look at where the `device` value comes from.

MaxKellermann:
No reply

PR #1186, opened 2021-06-10
Rough cut for DoP on OSS

constantinfuerst:
## Overview
What: DSD over PCM for the OSS output plugin.
Why: When using non-linux *NIX ALSA will not be available and therefore DSD gets resampled which is undesirable if DAC supports it via DoP.
## Introduction
This is a very rough cut for enabling dop under oss. I have little experience with this project and the code may contain some leftovers from experimentation. Please also excuse it being one huge commit. I changed some things locally to make the default branch of mpd compile under FreeBSD which I did not want to include here.
I am very much open to discussion and comments, please rip the code apart :^)
## Testing
This has been tested with a Sabaj Da2 under FreeBSD. Input files were of many formats: wavpack-dsd64, dsf-dsd64, dsf-dsd128, dsf-dsd256 (noisy). Up to and including DSD128 work flawless, DSD playback is taking place according to the DACs status LED. CPU usage is at 0%.
## Notes
The code written is quite hacky. This is the first "proof of work" implementation, any improvements or so can still take place. As I said, rip it apart.

### DIFF
--- a/src/output/plugins/OssOutputPlugin.cxx
+++ b/src/output/plugins/OssOutputPlugin.cxx
@@ -48,7 +48,9 @@
 #if defined(__OpenBSD__) || defined(__NetBSD__)
 # include <soundcard.h>
 #else /* !(defined(__OpenBSD__) || defined(__NetBSD__) */
+
 # include <sys/soundcard.h>
+
 #endif /* !(defined(__OpenBSD__) || defined(__NetBSD__) */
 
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Please fix your whitespace changes. You switched every line from tabs to spaces, which makes reviewing your code submission impossible, because every line is modified.

PR #1194, opened 2021-06-22
Enable DOP on OSS

constantinfuerst:
### Explanation
This adds support for DOP using the PcmExport function if the macro ENABLE_DSD is defined. If enabled within the config-file using "dop", the boolean dop_setting will be true. If DSD input is encountered and the setting is on, it is checked whether the oss-device supports the required samplerate. If that is the case, dop_active is set to true and conversion of the input is prevented. If the sample rate is not supported, conversion to S32 is requested. When playing back, the PcmExport is used to pack the incoming stream into PCM.
### Reasoning
This is required for OSs without the required driver support for native DSD playback that also have no ALSA. Mainly *BSD users are the target audience for this functionality, as ALSA here is only a proxy without full functionality.
### Requirements
- DAC that supports the DOP standard
- Building with OSS, DSD and S32-Format
### Supported Formats / Required PCM Formats
DSF, DFF and WavPack-DSD will work.

DSD64,  1 Channel -> S24:176.4kHz (untested, lack of time / missing samples)
DSD64,  2 Channel -> S24:352.8kHz
DSD64,  4 Channel -> S24:705.6kHz (untested, lmissing equipment)
DSD128, 1 Channel -> S24:352.8kHz (untested, lack of time / missing samples)
DSD128, 2 Channel -> S24:705.6kHz
DSD256, 1 Channel -> S24:705.6kHz (untested, lack of time / missing samples)
### Changes
- inclusion of required files
- adding new domain for logging
- adding dop_satisfied private function
- adding required member variables for storing dop state and for dop-packing
- adding dop boolean parameter to many functions that are required to act a little differently when dop is active
### Testing
This has been tested to work with a Sabaj Da2 on FreeBSD, where the red status indicator LED clearly shows that DSD playback is taking place, instead of purple for "hi-res" which is seen when converting.
### Issues
I have not tested this with S24 and right now AFMT_S32_NE is required. If not defined, ENABLE_DSD will be undef'ed. This will be addressed in a bit, however no DAC which supports DOP but not 32Bit is known to me. Also, AFMT_S32_NE is not defined when building on FreeBSD which is why this is just blatantly defined in the file at the moment.
Additionally, the new dop-option is not added into any documentation whatsoever.

### DIFF
--- a/src/output/plugins/OssOutputPlugin.cxx
+++ b/src/output/plugins/OssOutputPlugin.cxx
@@ -463,7 +463,11 @@ sample_format_to_oss(SampleFormat format
  */
 gcc_const
 static SampleFormat
-sample_format_from_oss(int format) noexcept
+sample_format_from_oss(int format
+#ifdef ENABLE_DSD
+        , bool dop
+#endif
+) noexcept
 {
 	switch (format) {
 	case AFMT_S8:
@@ -484,6 +488,9 @@ sample_format_from_oss(int format) noexcept
 
 #ifdef AFMT_S32_NE
 	case AFMT_S32_NE:
+#ifdef ENABLE_DSD
+        if (dop) return SampleFormat::DSD;
+#endif
 		return SampleFormat::S32;
 #endif
 
--- a/src/output/plugins/OssOutputPlugin.cxx
+++ b/src/output/plugins/OssOutputPlugin.cxx
@@ -510,12 +510,19 @@ static bool
 oss_probe_sample_format(FileDescriptor fd, SampleFormat sample_format,
 			SampleFormat *sample_format_r,
 			int *oss_format_r
+#ifdef ENABLE_DSD
+        , bool dop
+#endif
 #ifdef AFMT_S24_PACKED
 			, PcmExport &pcm_export
 #endif
 			)
 {
-	int oss_format = sample_format_to_oss(sample_format);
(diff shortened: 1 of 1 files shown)


src/output/plugins/OssOutputPlugin.cxx
```diff
@@ -68,6 +68,21 @@ class OssOutput final : AudioOutput {
...
+     *
+     * @see http://dsd-guide.com/dop-open-standard
+     *
+     * this is default in oss as no other dsd-method is known to man
+     */
+    bool dop_setting = false;
+    bool dop_active = false;
+
+    PcmExport::Params dop_params;
+    Manual<PcmExport> dop_export;
```
MaxKellermann, line 91: Wrong coding style - indent with tabs, not 4 spaces!

MaxKellermann:
I folded all commits into one and added this PR text as commit message. Your way of splitting patches made no sense at all, and the commit messages were bad.
955502f881e1569ad0b5646af304411c7b91767a
And I did a dozen bug fixes and simplified the code - after that, those 200 new code lines were reduces to 100.
I have no DoP capable hardware here; please test my modifications.

MaxKellermann:
> The messages just said what the change was and/or achieved but alas.

And still you felt the need to add text to this PR. All of that information was missing in the commit messages.

Issue #1177, opened 2021-05-29
Windows build fails on Linux (Debian testing i386)


skidoo23:
### Describe the bug
$ ../../win32/build.py --64 and $ ../../win32/build.py --32 failing.

## Expected Behavior
Build without error.

## Actual Behavior
Please note logfile below.

## Version

MPD from git/master

## Log

user@machine:~/src/MPD/output/win32$ ../../win32/build.py --32
Traceback (most recent call last):
  File "/home/user/src/MPD/output/win32/../../win32/build.py", line 94, in <module>
    from build.libs import *
  File "/home/user/src/MPD/python/build/libs.py", line 126, in <module>
    libopenmpt = AutotoolsProject(
  File "/home/user/src/MPD/python/build/autotools.py", line 14, in __init__
    MakeProject.__init__(self, url, md5, installed, **kwargs)
  File "/home/user/src/MPD/python/build/makeproject.py", line 9, in __init__
    Project.__init__(self, url, md5, installed, **kwargs)
  File "/home/user/src/MPD/python/build/project.py", line 24, in __init__
    if name is None: name = m.group(1)
AttributeError: 'NoneType' object has no attribute 'group'

PR #1196, opened 2021-06-23
Fix android build error when confronted with package versions ending with '+revision_information'

Yetangitu:
The script seems to assume package version numbers always end in numeric versions with an optional alpha-suffix. Alas, were it only so simple... Sometimes the package is called fizzbang-1.2.3+release_info in which case the build fails. No more!

### DIFF
--- a/python/build/project.py
+++ b/python/build/project.py
@@ -20,7 +20,7 @@ class Project:
             self.base = base
 
         if name is None or version is None:
-            m = re.match(r'^([-\w]+)-(\d[\d.]*[a-z]?[\d.]*)$', self.base)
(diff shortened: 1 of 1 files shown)


MaxKellermann:
What's the point of this change? Is this preparation to add a new library or update an existing one with such a version number?

MaxKellermann:
So this is about https://github.com/MusicPlayerDaemon/MPD/issues/1177
I cherry-picked it to 2052b461afd3ff06dc30cef06e92a580e85c1a99 and added a "Closes" line to the commit message.

Yetangitu:
> What's the point of this change? Is this preparation to add a new library or update an existing one with such a version number?

The point is to make the thing build in the face of libraries with such version numbers, these are used in e.g. Debian and derivatives.

Issue #1124, opened 2021-03-14
"password" configuration option sets wrong permissions


6arms1leg:
<!-- See https://www.musicpd.org/help/ -->
## Bug report
### Describe the bug
For the `password` option in the configuration file the [documentation says](https://www.musicpd.org/doc/html/user.html?highlight=password#permissions-and-passwords):

> "control - Allows all other player and playlist manipulations.
> admin - Allows database updates and allows shutting down MPD."

So I placed the following in my `mpd.conf`:

```
password                "admin@read,add,control,admin"
default_permissions     "read,add,control"
```

With the following intention:
Clients providing the `admin` password can do anything, clients not providing any password can do anything except database updates and shutting down MPD.

However, clients are still able to update the database without providing the password.  This was confirmed with several clients on different devices.

## Expected Behavior

With the above mentioned setup, according to the above mentioned documentation section, I would expect that only clients that provide the password are able to perform a database update.

To my understanding the line `default_permissions     "read,add,control"` means: "Allow every connected client to do everything except database updates and shutting down the MPD server."

## Actual Behavior

Connected clients are able to update the database although they don´t have the `admin` permission.

## Version
<!-- Paste the output of "mpd --version" here -->
```
~ $ mpd --version                                                                                                                                               
Music Player Daemon 0.22.6 (v20210212-1125-g123d157397)                                                                                                         
Copyright 2003-2007 Warren Dukes <warren.dukes@gmail.com>                                                                                                       
Copyright 2008-2018 Max Kellermann <max.kellermann@gmail.com>                                                                                                   
This is free software; see the source for copying conditions.  There is NO                                                                                      
warranty; not even MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.                                                                                         
                                                                                                                                                                
Database plugins:                                                                                                                                               
 simple proxy                                                                                                                                                   
                                                                                                                                                                
Storage plugins:                                                                                                                                                
 local smbclient nfs curl                                                                                                                                       
                                                                                                                                                                
Neighbor plugins:                                                                                                                                               
 smbclient                                                                                                                                                      
                                                                                                                                                                
Decoders plugins:                                                                                                                                               
 [mad] mp3 mp2                                                                                                                                                  
 [vorbis] ogg oga                                                                                                                                               
 [oggflac] ogg oga                                                                                                                                              
 [flac] flac                                                                                                                                                    
 [opus] opus ogg oga                                                                                                                                            
[...]

PR #1197, opened 2021-06-24
doc/user.rst: rectify admin permission

naglis:
Updating the database no longer requires the `admin` permission, only `control` is needed (changed in 2abad0f479).

See also: #1124


MaxKellermann:
Cherry-picked to v0.22.x: eb630ca655c7380abebc38a9675e9485afb917b5

PR #1214, opened 2021-07-28
decoder/ffmpeg: check if long_name is not null

sambazley:


MaxKellermann:
Is null legal in the FFmpeg API?

sambazley:
It can be null if FFmpeg is compiled with --enable-small, as it is for Android. (https://gitlab.freedesktop.org/gstreamer/gst-libav/-/issues/16)

Since 6f539cfcd64b2c6e5e4f5dd0a3d2f577e9c219c3 the Android service crashes when an mp3 is played.

PR #1240, opened 2021-08-21
clang-tidy: make deleted members public

neheb:
Found with modernize-use-equals-delete


MaxKellermann:
What's the point?

neheb:
The check activates because this is a C++98 practice of declaring deleted constructors private to prevent anyone from using them. With = delete, there's no need for them to be private.

MaxKellermann:
> With = delete, there's no need for them to be private.

And there's no need for them to be public either. So, again: what's the point?

neheb:
Right. It's just best practice.

PR #1243, opened 2021-08-21
Fix compile error on macOS 11.15.2 (introduced by commit 30e3ef4)

dhocker:
Fixes issue #1232


neheb:
:)

Maybe the codebase should be ran through include-what-you-use.

PR #1249, opened 2021-08-26
add missing virtual

neheb:
There are virtual members in these classes.


neheb:
cppcheck warning:

```
warning: Class 'ServerSocket' which has virtual members does not have a virtual destructor. [virtualDestructor]
```

It seems they don't inherit one either.

MaxKellermann:
Yes, but why do *you* believe they are missing? Just because some random tool believes that a class with virtual members must have a virtual destructor? Why is that so? Can you explain this requirement?

PR #1250, opened 2021-08-26
unique_ptr conversions

neheb:
Avoids manual new/delete.

### DIFF
--- a/src/RemoteTagCache.cxx
+++ b/src/RemoteTagCache.cxx
@@ -44,7 +44,7 @@ RemoteTagCache::Lookup(const std::string &uri) noexcept
 	KeyMap::insert_commit_data hint;
 	auto [tag, value] = map.insert_check(uri, Item::Hash(), Item::Equal(), hint);
 	if (value) {
-		auto item = new Item(*this, uri);
+		auto item = std::make_unique<Item>(*this, uri);
 		map.insert_commit(*item, hint);
 		waiting_list.push_back(*item);
 		lock.unlock();
--- a/src/lib/upnp/Discovery.cxx
+++ b/src/lib/upnp/Discovery.cxx
@@ -177,12 +177,12 @@ UPnPDeviceDirectory::OnAlive(const UpnpDiscovery *disco) noexcept
 	if (isMSDevice(UpnpDiscovery_get_DeviceType_cstr(disco)) ||
 	    isCDService(UpnpDiscovery_get_ServiceType_cstr(disco))) {
 		try {
-			auto *downloader = new Downloader(*this, *disco);
+			auto downloader = std::make_unique<Downloader>(*this, *disco);
 
 			try {
 				downloader->Start();
 			} catch (...) {
-				BlockingCall(GetEventLoop(), [downloader](){
+				BlockingCall(GetEventLoop(), [&downloader](){
 						downloader->Destroy();
 					});
 
--- a/src/mixer/plugins/AlsaMixerPlugin.cxx
+++ b/src/mixer/plugins/AlsaMixerPlugin.cxx
@@ -81,7 +81,7 @@ class AlsaMixer final : public Mixer {
 	snd_mixer_t *handle;
 	snd_mixer_elem_t *elem;
 
-	AlsaMixerMonitor *monitor;
+	std::unique_ptr<AlsaMixerMonitor> monitor;
 
 	/**
 	 * These fields are our workaround for rounding errors when
@@ -282,7 +282,7 @@ AlsaMixer::Setup()
 	snd_mixer_elem_set_callback_private(elem, this);
 	snd_mixer_elem_set_callback(elem, ElemCallback);
 
-	monitor = new AlsaMixerMonitor(event_loop, handle);
+	monitor = std::make_unique<AlsaMixerMonitor>(event_loop, handle);
 }
 
 void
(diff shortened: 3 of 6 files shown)


src/input/plugins/QobuzInputPlugin.cxx
```diff
@@ -33,7 +33,7 @@
 
 #include <memory>
 
-static QobuzClient *qobuz_client;
+static std::unique_ptr<QobuzClient> qobuz_client;
```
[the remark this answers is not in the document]
neheb: Removed.

src/lib/upnp/Discovery.cxx
```diff
@@ -177,12 +177,12 @@ UPnPDeviceDirectory::OnAlive(const UpnpDiscovery *disco) noexcept
 	if (isMSDevice(UpnpDiscovery_get_DeviceType_cstr(disco)) ||
 	    isCDService(UpnpDiscovery_get_ServiceType_cstr(disco))) {
 		try {
-			auto *downloader = new Downloader(*this, *disco);
+			auto downloader = std::make_unique<Downloader>(*this, *disco);
```
MaxKellermann, line 180: This change will lead to a crash. Once started, this object must be destroyed only inside the `GetEventLoop()`'s thread (that's why `BlockingCall()` to move the `Destroy()` call into that thread upon error), and only when it's finished. But there are two mistakes here:
- your code unconditionally deletes the object at the end of this scope, leading to use-after-free crash bug
- your code doesn't delete the object in the correct crash, leading to various kinds of race conditions

src/mixer/plugins/AlsaMixerPlugin.cxx
```diff
@@ -81,7 +81,7 @@ class AlsaMixer final : public Mixer {
 	snd_mixer_t *handle;
 	snd_mixer_elem_t *elem;
 
-	AlsaMixerMonitor *monitor;
+	std::unique_ptr<AlsaMixerMonitor> monitor;
```
MaxKellermann, line 84: This one is pointless overhead, because it's going to be deleted in `Close()`, which is guaranteed to be called if `Open()` has been successful. `~AlsaMixer()` calling `~unique_ptr<AlsaMixerMonitor>` will never see any non-nullptr value, thus the overhead for checking again is pointless.
Clearly, this isn't proper RAII C++ code, but your change doesn't improve this in any way. It just puts some lipstick on...

src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -66,7 +66,7 @@ class RecorderOutput final : AudioOutput {
 	/**
 	 * The destination file.
 	 */
-	FileOutputStream *file;
+	std::unique_ptr<FileOutputStream> file;
```
MaxKellermann, line 69: Same as in `AlsaMixer`. Pointless overhead.

src/output/plugins/RecorderOutputPlugin.cxx
```diff
@@ -251,14 +249,13 @@ RecorderOutput::ReopenFormat(AllocatedPath &&new_path)
...
 
-	auto *new_file = new FileOutputStream(new_path);
+	auto new_file = std::make_unique<FileOutputStream>(new_path);
 
 	AudioFormat new_audio_format = effective_audio_format;
 
 	try {
 		encoder = prepared_encoder->Open(new_audio_format);
 	} catch (...) {
-		delete new_file;
```
MaxKellermann, line 261: Being able to avoid these would be nice, but not at the cost of adding pointless overhead to `~RecorderOutput()`. Maybe call `file = new_file.release()` later in this method?

src/output/plugins/httpd/HttpdOutputPlugin.cxx
```diff
@@ -89,7 +89,7 @@ HttpdOutput::Unbind() noexcept
 inline void
 HttpdOutput::AddClient(UniqueSocketDescriptor fd) noexcept
 {
-	auto *client = new HttpdClient(*this, std::move(fd), GetEventLoop(),
+	auto client = std::make_unique<HttpdClient>(*this, std::move(fd), GetEventLoop(),
 				       !encoder->ImplementsTag());
 	clients.push_front(*client);
```
MaxKellermann, line 94: This will crash. You create a new `HttpClient`, add it to the linked list, and upon leaving `AddClient()`, the `HttpdClient()` will be freed. What happens to the element still in the linked list?

src/output/plugins/snapcast/SnapcastOutputPlugin.cxx
```diff
@@ -103,7 +103,7 @@ SnapcastOutput::Unbind() noexcept
 inline void
 SnapcastOutput::AddClient(UniqueSocketDescriptor fd) noexcept
 {
-	auto *client = new SnapcastClient(*this, std::move(fd));
+	auto client = std::make_unique<SnapcastClient>(*this, std::move(fd));
```
MaxKellermann, line 106: Same use-after-free crash bug as the one you added to `HttpdOutput::AddClient()`.

src/RemoteTagCache.cxx
```diff
@@ -44,7 +44,7 @@ RemoteTagCache::Lookup(const std::string &uri) noexcept
 	KeyMap::insert_commit_data hint;
 	auto [tag, value] = map.insert_check(uri, Item::Hash(), Item::Equal(), hint);
 	if (value) {
-		auto item = new Item(*this, uri);
+		auto item = std::make_unique<Item>(*this, uri);
```
MaxKellermann, line 47: Same crash bug as in `HttpdClient::AddClient()`, see below.

MaxKellermann (commented):
Further code review shows that all changes in this PR are either pointless overhead adding no safety at all, or introduce crash bugs.

PR #1261, opened 2021-09-11
replace several emplace_back with push_back

neheb:
The former is pointless when the types match and when there is no need
to call the constructor.

Verified with clang-tidy's modernize-use-emplace.

### DIFF
--- a/src/PlaylistFile.cxx
+++ b/src/PlaylistFile.cxx
@@ -239,7 +239,7 @@ try {
 				continue;
 		}
 
-		contents.emplace_back(std::move(uri_utf8));
+		contents.push_back(std::move(uri_utf8));
 		if (contents.size() >= playlist_max_length)
 			break;
 	}
--- a/src/command/DatabaseCommands.cxx
+++ b/src/command/DatabaseCommands.cxx
@@ -295,13 +295,13 @@ handle_list(Client &client, Request args, Response &r)
 			return CommandResult::ERROR;
 		}
 
-		tag_types.emplace_back(group);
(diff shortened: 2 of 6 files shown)


MaxKellermann:
> The former is pointless when the types match

But ... what exactly is the point of changing this?

> there is no need to call the constructor.

This sounds like you avoid a constructor call, but in fact both methods call the constructor.

neheb:
push_back generates smaller code than emplace_back.

As far as the constructor goes,

```
push_back(1, "test")
```

Does not compile whereas the emplace_back version does.

MaxKellermann:
> push_back generates smaller code than emplace_back.

How come?

> Does not compile whereas the emplace_back version does.

I don't understand. And your reply doesn't appear to have anything to do with my assertion that both methods call a constructor.

neheb:
> > push_back generates smaller code than emplace_back.
> 
> How come?
No idea. What I've noticed is in the case of multi parameter emplace_back it's the same. But single parameter causes a size increase. I've noticed this mostly with const char * and std::string.
> 
> > Does not compile whereas the emplace_back version does.
> 
> I don't understand. And your reply doesn't appear to have anything to do with my assertion that both methods call a constructor.

MaxKellermann:
These changes make no sense. Both versions should lead to the same machine code; if they don't, it's an artefact of current compiler versions.
Anyway, the existing code expresses more clearly what should be happening, so I disagree with these changes, even if they lead to slightly-smaller (but not proven better) machine code with current compilers.

PR #1318, opened 2021-10-29
pulse: link statically

neheb:
The stock pkgconfig files for pulseaudio wrongly do not include lib/pulseaudio as a library path.

This causes undefined function errors while linking. Adding static here allows it to link.


MaxKellermann:
I don't understand. This is a workaround for a PulseAudio bug? And the PulseAudio bug is a missing path in `Libs.private`, a pkg-config variable which is used for static linking? And this static-linking bug prevents dynamic-linking? And the fix for this static-linking bug is to switch from dynamic-linking to static-linking? This doesn't appear to make any sense.
Besides that, I'm not interested in workarounds for bugs in third-party packages. People who suffer from this bug shall use the latest PulseAudio version, or edit the pkg-config file. But your suggestion isn't good; it just silently links statically (which can cause problems of its own), but there's not even a comment explaining the workaround, and the workaround is even used for unaffected PulseAudio versions.

MaxKellermann:
No reply.

PR #1326, opened 2021-11-04
python/build/libs.py: update libopenmpt configure flags

ibmibmibm:


MaxKellermann:
Can we also avoid building the EXE with some cmake option?

ibmibmibm:
https://github.com/MusicPlayerDaemon/MPD/pull/1327

PR #1357, opened 2021-11-23
use cinttypes header

neheb:
inttypes.h is deprecated.


MaxKellermann:
btw. you don't need to split stuff into several PRs (as long as you properly separate commits); just put all the refactoring in one PR, and if something needs further discussion, I know how to cherry-pick the others. (But you can keep several branches/PRs if you prefer that; just know that I don't care.)

neheb:
OK

PR #1359, opened 2021-11-24
various cleanups

neheb:


maxmitti (commented):
The temporaries are there for a reason.
With your changes the contexts are accessed after they are destroyed.

maxmitti (commented):
Regarding https://github.com/MusicPlayerDaemon/MPD/pull/1359/commits/d019d7c664784faca0fd2394f121b494aea4ea20:

Not only is the code uglier than before (my opinion), but now operations on invalid sockets are performed instead of being skipped in error cases.

neheb:
ah yeah that's true.

src/LocateUri.cxx
```diff
@@ -121,18 +119,19 @@ LocateUri(UriPluginKind kind,
...
 #endif
 				     );
-	else if (uri_has_scheme(uri))
+	if (uri_has_scheme(uri))
 		return LocateAbsoluteUri(kind, uri
 #ifdef ENABLE_DATABASE
 					 , storage
 #endif
 					 );
-	else
```
MaxKellermann, line 136: Removing all the `else`s has nothing to do with braced init list conversion. Please don't hide changes like this one in unrelated commits. I don't agree with this change; while the `else` is redundant here, it still demonstrates to the human reader that this function consists of many alternatives; the control flow is made visible. And it is more robust this way; one missing `return` in one branch will be complained about by the compiler.

MaxKellermann:
>  work around clang-tidy warning
> It thinks it can simplify the expression because of the define.

What?? Apart from the serious regression mentioned by @maxmitti, how is this an improvement? Fail and return early in case of an error.

src/lib/nfs/Connection.cxx
```diff
@@ -595,11 +595,10 @@ NfsConnection::MountInternal()
 
 	if (nfs_mount_async(context, server.c_str(), export_name.c_str(),
 			    MountCallback, this) != 0) {
-		auto e = FormatRuntimeError("nfs_mount_async() failed: %s",
-					    nfs_get_error(context));
 		nfs_destroy_context(context);
 		context = nullptr;
-		throw e;
+		throw FormatRuntimeError("nfs_mount_async() failed: %s",
+					 nfs_get_error(context));
```
MaxKellermann, line 601: This will crash because `context==nullptr` now (which should be *very* obvious because the very line preceding this one does it). There's a reason I constructed the exception first, and did the `throw` only after cleaning up!

src/output/plugins/PulseOutputPlugin.cxx
```diff
@@ -502,10 +502,8 @@ PulseOutput::WaitConnection()
 		case PA_CONTEXT_FAILED:
 			/* failure */
 			{
-				auto e = MakePulseError(context,
-							"failed to connect");
 				DeleteContext();
-				throw e;
+				throw MakePulseError(context, "failed to connect");
```
MaxKellermann, line 506: Same here: this will crash.

src/fs/StandardDirectory.cxx
```diff
@@ -127,7 +127,7 @@ static AllocatedPath GetStandardDir(int folder_id)
 
 #ifdef USE_XDG
 
-static const char home_prefix[] = "$HOME/";
+static constexpr auto home_prefix = "$HOME/";
```
MaxKellermann, line 130: This is not a `std::array` conversion, is it?

src/playlist/plugins/SoundCloudPlaylistPlugin.cxx
```diff
@@ -189,11 +188,8 @@ SoundCloudJsonData::String(StringView value) noexcept
 inline bool
 SoundCloudJsonData::MapKey(StringView value) noexcept
 {
-	const auto *i = key_str;
-	while (*i != nullptr && !StringStartsWith(*i, value))
-		++i;
-
-	key = SoundCloudJsonData::Key(i - key_str);
+	auto c = std::count_if(key_str.begin(), key_str.end(), [=](auto &&str){ return !StringStartsWith(str, value); });
```
MaxKellermann, line 191: This code is not equal to the old code.

src/fs/StandardDirectory.cxx
neheb: More like a C array conversion to const char*. IIRC clang-tidy warns on it.

neheb:
Updated.

src/fs/StandardDirectory.cxx
MaxKellermann: "some random tool warns" alone does not convince me. If you can't explain a change, please do not submit. And please do not submit changes which you did not test.

neheb: It's gone now.

src/lib/curl/Escape.cxx
```diff
@@ -60,7 +60,7 @@ CurlUnescape(CURL *curl, StringView src) noexcept
 	int outlength;
 	CurlString tmp(curl_easy_unescape(curl, src.data, src.size,
 					  &outlength));
-	return std::string(tmp.c_str(), outlength);
+	return {tmp.c_str(), size_t(outlength)};
```
ibmibmibm, line 63: This is a c-style cast to `size_t`
should use `static_cst<size_t>(outlength)`

MaxKellermann:
4e276256c088694c5a009633ef4469233884312f

PR #1392, opened 2021-12-26
Fix non-x86 Android build error: needs_exe_wrapper

sambazley:
Fixes build error when compiling for arm64 Android:
```
lib/src/libmpdclient-2.19/meson.build:1:0: ERROR: Unknown options: "needs_exe_wrapper"
```

### DIFF
--- a/python/build/meson.py
+++ b/python/build/meson.py
@@ -53,19 +53,21 @@ pkgconfig = '{toolchain.pkg_config}'
         f.write(f"""
 [properties]
 root = '{toolchain.install_prefix}'
-
-[built-in options]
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Why does this error occur with ARM builds? Which Meson version are you using?
There's one major problem with your solution: it assumes that your build machine is x86.

sambazley:
Sorry, I misunderstood what that option does. The commit should be correct now. The error was caused because needs_exe_wrapper was set under [built-in options], but should be set under [properties].

MaxKellermann:
Cherry-picked to v0.23.xfbecb05bf47dfdda61dac953b78f998273772b0d

PR #1398, opened 2022-01-04
Android: Detect output change with ACTION_AUDIO_BECOMING_NOISY

sambazley:
Improves the changes made in 57687779befd796b79b1a569daf9e48eac3aea26 by
using AudioManager.ACTION_AUDIO_BECOMING_NOISY rather than listening for
wired headset unplug events or Bluetooth headset disconnect events. This
method is more flexible, allowing the feature to work on other types of
audio output device, as well as Bluetooth devices that don't set their
device class correctly. This change also has the benefit of being more
responsive, pausing the audio before it is rerouted to the built-in
speaker.

https://developer.android.com/guide/topics/media-apps/volume-and-earphones


MaxKellermann:
Cherry-picked to v0.23.x: 49837033755dbfc9da951892a3a73c4ba4541602

PR #1430, opened 2022-02-02
Fix DSCP LE value

dgcampea:
Correct value is 0x04 since the lower 2 ECN bits need to be accounted for.

### DIFF
--- a/src/net/DscpParser.cxx
+++ b/src/net/DscpParser.cxx
@@ -69,7 +69,7 @@
 #endif
 
 #ifndef IPTOS_DSCP_LE
-#define IPTOS_DSCP_LE 0x01
+#define IPTOS_DSCP_LE 0x04
 #endif
 
 static constexpr struct {


MaxKellermann:
Your PR contains text which is missing in the commit message.
I doubt this change is correct - what's the source for your information? Why do you believe the value is incorrect?

dgcampea:
[RFC 8622, section 6](https://datatracker.ietf.org/doc/html/rfc8622#section-6) indicates that the _codepoint_ is '000001' (or '0x01').

Quoting RFC 2474:
```

   Six bits of the DS field are used as a codepoint (DSCP) to select the
   PHB a packet experiences at each node.  A two-bit currently unused
   (CU) field is reserved and its definition and interpretation are
   outside the scope of this document.  The value of the CU bits are
   ignored by differentiated services-compliant nodes when determining
   the per-hop behavior to apply to a received packet.

   The DS field structure is presented below:


        0   1   2   3   4   5   6   7
      +---+---+---+---+---+---+---+---+
      |         DSCP          |  CU   |
      +---+---+---+---+---+---+---+---+

        DSCP: differentiated services codepoint
        CU:   currently unused

   In a DSCP value notation 'xxxxxx' (where 'x' may equal '0' or '1')
   used in this document, the left-most bit signifies bit 0 of the DS
   field (as shown above), and the right-most bit signifies bit 5.
```

In fact, peeking at <netinet/ip.h> shows this:
```
#define IPTOS_DSCP_MASK         0xfc
#define IPTOS_DSCP(x)           ((x) & IPTOS_DSCP_MASK)
```
Where `0xfc = b11111100`, so `0x01` as a DSCP value is actually changing the ECN bits.
(It can be confusing at first but [RFC 1700](https://datatracker.ietf.org/doc/html/rfc1700) lists the conventions for byte and bit endianness).

PR #1439, opened 2022-02-13
doc/user.rst: Clarify how MPD reads metadata

wylfen:
As per our discussion on IRC. I didn't want to go into too much detail about each specific decoder, especially since I don't know which ones support metadata and which ones don't. This should put it as succinctly as possible.


MaxKellermann:
Cherry-picked to v0.23.x 59da778009eaabacc2c648bfd00c395b0112c123

PR #1452, opened 2022-02-20
Add (resurrect?) doxygen support

sp1ff:
Added a `doxygen` option to the `doc` build. It makes use of the already-present but unused file `doxygen.conf.in`. The option is set to false by default.

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -37,9 +37,9 @@ if get_option('manpages')
   )
 endif
 
-doxygen = find_program('doxygen', required: false)
-if get_option('doxygen') and doxygen.found()
+if get_option('doxygen')
 
+  doxygen = find_program('doxygen', required: true)
   conf_data = configuration_data()
   conf_data.set('VERSION', meson.project_version())
   conf_data.set('abs_top_srcdir', meson.source_root())
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -36,3 +36,22 @@ if get_option('manpages')
     install_dir: [join_paths(get_option('mandir'), 'man1'), join_paths(get_option('mandir'), 'man5')],
   )
 endif
+
+doxygen = find_program('doxygen', required: false)
+if get_option('doxygen') and doxygen.found()
+
+  conf_data = configuration_data()
+  conf_data.set('VERSION', meson.project_version())
+  conf_data.set('abs_top_srcdir', meson.source_root())
+  doxyfile = configure_file(
+    input: 'doxygen.conf.in',
+    output: 'Doxyfile',
+    configuration: conf_data
+  )
+  
+  custom_target('Doxygen',
+    command: [doxygen, doxyfile],
+    output: ['api'],
+    build_by_default: true
+  )
+endif
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,6 +1,7 @@
 option('documentation', type: 'feature', description: 'Build documentation')
 option('html_manual', type: 'boolean', value: true, description: 'Build the HTML manual')
 option('manpages', type: 'boolean', value: true, description: 'Build manual pages')
+option('doxygen', type: 'boolean', value: false, description: 'Build doxygen source documentation')
 
 option('syslog', type: 'feature', description: 'syslog support')
 option('inotify', type: 'boolean', value: true, description: 'inotify support (for automatic database update)')


doc/meson.build
```diff
@@ -36,3 +36,22 @@ if get_option('manpages')
     install_dir: [join_paths(get_option('mandir'), 'man1'), join_paths(get_option('mandir'), 'man5')],
   )
 endif
+
+doxygen = find_program('doxygen', required: false)
```
MaxKellermann, line 40: Why look up doxygen if the option is disabled?

doc/meson.build
```diff
@@ -36,3 +36,22 @@ if get_option('manpages')
...
+if get_option('doxygen') and doxygen.found()
```
MaxKellermann, line 41: With `-Ddoxygen=true`, failure to find doxygen should be a fatal error.

MaxKellermann (changes requested):
What's your motivation to add doxygen rules? MPD had it for many years, and nobody ever looked at the HTML API documentation.

doc/meson.build
sp1ff: Good point -- fixed.

doc/meson.build
sp1ff: Yes. Corrected.

sp1ff:
> What's your motivation to add doxygen rules? MPD had it for many years, and nobody ever looked at the HTML API documentation.

My motivation is to be able to look at the HTML API documentation :) MPD is nearly 200K lines of C++ and I thought the doxgyen-generated docs would be useful to me as a first-time contributor trying to learn the code base. That said, if I am really the only one, feel free to close this PR & I'll just leave the change on my fork for my personal use.

MaxKellermann:
If you think it's easier than reading the source code, that's fine. Whenever I see doxygen-generated HTML pages, I feel lost, and I prefer the raw source code. But that's personal taste, and having these Meson rules is fine as long as they don't bother people who don't enable them.

doc/meson.build
```diff
@@ -36,3 +36,22 @@ if get_option('manpages')
     install_dir: [join_paths(get_option('mandir'), 'man1'), join_paths(get_option('mandir'), 'man5')],
   )
 endif
+
+if get_option('doxygen')
+
+  doxygen = find_program('doxygen', required: true)
```
eli-schwartz, line 42: `required: true` is the default, so it is okay to skip passing it as a kwarg.

meson_options.txt
```diff
@@ -1,6 +1,7 @@
 option('documentation', type: 'feature', description: 'Build documentation')
 option('html_manual', type: 'boolean', value: true, description: 'Build the HTML manual')
 option('manpages', type: 'boolean', value: true, description: 'Build manual pages')
+option('doxygen', type: 'boolean', value: false, description: 'Build doxygen source documentation')
```
eli-schwartz, line 4: Maybe this should be a feature option instead, then doxygen can be looked up unconditionally and in the disabled case it will log "skipped: feature doxygen is disabled". Then you can handle the doc setup by simply checking if doxygen was found.

See how -Ddocmentation is handled for sphinx-build.

doc/meson.build
```diff
@@ -36,3 +36,22 @@ if get_option('manpages')
     install_dir: [join_paths(get_option('mandir'), 'man1'), join_paths(get_option('mandir'), 'man5')],
   )
 endif
+
+if get_option('doxygen')
```
eli-schwartz, line 40: If sphinx-build is not found, this file exits early and never reaches the doxygen checks.

MaxKellermann: True, this is because of the `subdir_done()` kludge earlier in this file; it's there to simplify the rules, but with Doxygen, the file needs to grow more complex.

meson_options.txt
MaxKellermann: Agree, that would be better, but since nobody really needs to build the docs (only those with an explicit preference), the default value should be "disabled" to reduce build overhead. The same could probably be done for the `documentation` flag. Nobody builds documentation (but me).

doc/meson.build
eli-schwartz: Actually it can be made even less complex. :)

Since 0.49.0 you can do:
```meson
find_program('fooprog', required: false, disabler: true)
```

And if it is not found, instead of doing .found() checks on it, you can simply add custom targets making use of it directly, and those targets -- because they use a disabler object in their definition -- get converted to disablers and have no effect.

So any Meson function call that attempts to use a disabler gets "infected" by the disabler, becoming a no-op. If doxygen isn't found, the attempt to build doxygen docs is no-oped and disappears. If sphinx-build isn't found, the attempt to run sphinx-build to create the manual / manpages is no-oped and disappears.

MaxKellermann: Good hint. MPD is at `>=0.56.0`, so we can easily use this Meson feature.

PR #1461, opened 2022-02-22
WasapiOutputPlugin pause bug fix

RichardSchorrig:
Wasapi output plugin won't start playing after being paused

### DIFF
--- a/src/output/plugins/wasapi/WasapiOutputPlugin.cxx
+++ b/src/output/plugins/wasapi/WasapiOutputPlugin.cxx
@@ -468,9 +468,19 @@ try {
 
 		case Status::PLAY:
 			break;
-		}
+		}		
 
 		UINT32 write_in_frames = buffer_size_in_frames;
+		DWORD mode = 0;
+		AtScopeExit(&) {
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Your commit message is very short and lacks an explanation. What exact issue is being fixed? What was causing it? How does your change solve it?

RichardSchorrig:
for the bug description, please see https://github.com/MusicPlayerDaemon/MPD/issues/1451#issue-1145061606

The cause is that the scope guard in the WASAPI work thread (WasapiOutputPlugin.cxx, function WasapiOutputThread::Work(), in the while (true) loop) is set up too 'late' in the execution. There is one condition ("if (data_in_frames >= buffer_size_in_frames)") when it is hit, the loop will continue without executing the scope guard. This scope guard is responsible for emptying the buffer again, and if the buffer is not emptied, the above mentioned condition will stay true.

MaxKellermann:
Cherry-picked to v0.23.x with proper commit message aeaef855077384f4d669dc8013948a17b5ce38cc

PR #1480, opened 2022-03-09
fix pre-open/enable edge cases PipeWireOutput::SetVolume

ArsenArsen:
this patchset prevents a crash due to there being no thread loop yet, as well as (re)storing the volume under more conditions.

### DIFF
--- a/src/output/plugins/PipeWireOutputPlugin.cxx
+++ b/src/output/plugins/PipeWireOutputPlugin.cxx
@@ -311,6 +311,9 @@ PipeWireOutput::PipeWireOutput(const ConfigBlock &block)
 void
 PipeWireOutput::SetVolume(float _volume)
 {
+	if (thread_loop == nullptr)
+		return;
+
 	const PipeWire::ThreadLoopLock lock(thread_loop);
 
 	float newvol = _volume*_volume*_volume;
--- a/src/output/plugins/PipeWireOutputPlugin.cxx
+++ b/src/output/plugins/PipeWireOutputPlugin.cxx
@@ -88,7 +88,7 @@ class PipeWireOutput final : AudioOutput {
 	float volume = 1.0;
 
 	PipeWireMixer *mixer = nullptr;
-	unsigned channels;
+	unsigned channels = 0;
 
 	/**
 	 * The active sample format, needed for PcmSilence().
@@ -128,7 +128,7 @@ class PipeWireOutput final : AudioOutput {
 	 * done each time after the pw_stream got created, thus this
 	 * flag gets set by Open().
 	 */
-	bool restore_volume;
+	bool restore_volume = false;
 
 	bool interrupted;
 	bool paused;
@@ -311,14 +311,18 @@ PipeWireOutput::PipeWireOutput(const ConfigBlock &block)
 void
 PipeWireOutput::SetVolume(float _volume)
 {
-	if (thread_loop == nullptr)
+	if (thread_loop == nullptr) {
+		// set volume before we are even enabled, restore it later
+		restore_volume = true;
+		volume = _volume;
 		return;
+	}
 
 	const PipeWire::ThreadLoopLock lock(thread_loop);
 
 	float newvol = _volume*_volume*_volume;
 
-	if (stream != nullptr && !restore_volume) {
+	if (stream != nullptr && !restore_volume && channels != 0) {
 		float vol[MAX_CHANNELS];
 		std::fill_n(vol, channels, newvol);
 
@@ -326,6 +330,10 @@ PipeWireOutput::SetVolume(float _volume)
 				  SPA_PROP_channelVolumes, channels, vol,
 				  0) != 0)
 			throw std::runtime_error("pw_stream_set_control() failed");
+	} else {
+		// we seem to be called before a stream is opened, restore the
+		// volume later instead
+		restore_volume = true;
 	}
 
 	volume = _volume;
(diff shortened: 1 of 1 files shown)


src/output/plugins/PipeWireOutputPlugin.cxx
```diff
@@ -311,21 +311,29 @@ PipeWireOutput::PipeWireOutput(const ConfigBlock &block)
...
+		volume = _volume;
 		return;
+	}
 
 	const PipeWire::ThreadLoopLock lock(thread_loop);
 
 	float newvol = _volume*_volume*_volume;
 
-	if (stream != nullptr && !restore_volume) {
+	if (stream != nullptr && !restore_volume && channels != 0) {
```
MaxKellermann, line 325: What's the point of this new check? If `stream != nullptr`, then `channels` is also set.

src/output/plugins/PipeWireOutputPlugin.cxx
```diff
@@ -128,7 +128,7 @@ class PipeWireOutput final : AudioOutput {
 	 * done each time after the pw_stream got created, thus this
 	 * flag gets set by Open().
 	 */
-	bool restore_volume;
+	bool restore_volume = false;
```
MaxKellermann, line 131: What's the point of this initializer?

src/output/plugins/PipeWireOutputPlugin.cxx
```diff
@@ -88,7 +88,7 @@ class PipeWireOutput final : AudioOutput {
 	float volume = 1.0;
 
 	PipeWireMixer *mixer = nullptr;
-	unsigned channels;
+	unsigned channels = 0;
```
MaxKellermann, line 91: Another unnecessary initializer

src/output/plugins/PipeWireOutputPlugin.cxx
```diff
@@ -311,6 +311,9 @@ PipeWireOutput::PipeWireOutput(const ConfigBlock &block)
 void
 PipeWireOutput::SetVolume(float _volume)
 {
+	if (thread_loop == nullptr)
```
MaxKellermann: Did you ever observe that crash, or is this just a theory?

ArsenArsen: I have. Setting the volume before playing anything can reproduce it consistently.

```
[i] ~$ mpc clear
volume: 81%   repeat: off   random: off   single: off   consume: off
(reverse-i-search)`': ^C
[i] ~ 130 $ systemctl --user restart mpd.service
[i] ~$ mpc volume 100
MPD error: Connection closed by the server
[i] ~ 1 $
```

src/output/plugins/PipeWireOutputPlugin.cxx
ArsenArsen: I observed a case where this clause would run with an uninitialized channel value leading to a stack corruption. I will attempt reproducing it again and get back to you

src/output/plugins/PipeWireOutputPlugin.cxx
MaxKellermann: `output_mixer_set_volume()` checks whether the output is enabled. This must be a bug elsewhere, and your change is just a bad workaround.

src/output/plugins/PipeWireOutputPlugin.cxx
MaxKellermann: Maybe the real problem is that `Enable()` doesn't initialize `stream`?

MaxKellermann: I fixed that now in v0.23.x: 9975905fafcf49c829e7b79cd66e9da3aab509a8

src/output/plugins/PipeWireOutputPlugin.cxx
```diff
@@ -323,6 +330,10 @@ PipeWireOutput::SetVolume(float _volume)
 				  SPA_PROP_channelVolumes, channels, vol,
 				  0) != 0)
 			throw std::runtime_error("pw_stream_set_control() failed");
+	} else {
+		// we seem to be called before a stream is opened, restore the
+		// volume later instead
+		restore_volume = true;
```
MaxKellermann, line 336: This is redundant; it is already set by `Open()`.

src/output/plugins/PipeWireOutputPlugin.cxx
```diff
@@ -311,6 +311,13 @@ PipeWireOutput::PipeWireOutput(const ConfigBlock &block)
 void
 PipeWireOutput::SetVolume(float _volume)
 {
+	if (thread_loop == nullptr) {
```
MaxKellermann, line 314: As I already wrote, this is a bad workaround. Note that this code change doesn't match your commit message: `thread_loop` is **not** set when the output is opened, but when it is enabled.

src/output/plugins/PipeWireOutputPlugin.cxx
ArsenArsen: ah! good catch

MaxKellermann:
Your commit messages are lacking. They don't explain anything.

ArsenArsen:
I can reword when I get home tonight

MaxKellermann:
You force-pushed, but you did not improve the commit messages

MaxKellermann:
Cherry-picked to v0.23.x 35dbc1a90c5110fbf6ab18e98c43463d1b4ea4f3

ArsenArsen:
Thank you!

Issue #428, opened 2018-11-28
mpd_status_get_kbit_rate returns 0 for ogg files


quimup:
In clients GMPC and Quimup (and othres?) OGG files always show 0 kbps. I believe the same is true for APE. It's been like this for quite some time (i.e. several MPD versions). 

regards,

JS

MaxKellermann:
0 is the magic value for "don't know". Ask the client developers to show this accordingly. No client should ever show "0".

quimup:
Max,
I know 0 means "don't know". But that's just another way to describe the bug, isn't it? MPD should know, after all. The exact same code worked fine some time ago, both in Quimup and GMPC.

JS

MaxKellermann:
Then you should give more information. For example the last MPD version which worked. Hey, I don't even know what version you're using currently!

quimup:
Sorry: Currently using 0.20.23 from Debian (testing). I can't remember what was the last version that worked. Let me know if you need more info.

JS

MaxKellermann:
Will reopen as soon as you've posted all the information.

quimup:
You can't be bothered to say what "all the information" is?

MaxKellermann:
You can't be bothered to read the link I posted?

MaxKellermann:
Oh, sorry, I only posted it in the other bug, this is it: https://www.musicpd.org/help/
But I explicitly asked you to tell the last working MPD version. If you can't remember, then find out.

quimup:
Max,

I tried to install 0.21.3 but it is not in any of the Debian repos yet. Other sources have unmet dependencies, so I am stuck at 0.20.23

Finding out what was the last version that worked is not doable. Could be more that a year ago: that's digital archeology. Besides, the problem may be caused by one of the libraries that MPD uses.


I hope this is all you need:



///////////////////////// ~$ mpd --version /////////////////////////

Music Player Daemon 0.20.23

Copyright (C) 2003-2007 Warren Dukes <warren.dukes@gmail.com>
Copyright 2008-2017 Max Kellermann <max.kellermann@gmail.com>
This is free software; see the source for copying conditions.  There is NO
warranty; not even MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Database plugins:
 simple proxy upnp

Storage plugins:
 local smbclient nfs curl

Neighbor plugins:
 smbclient upnp

Decoders plugins:
 [mad] mp3 mp2
[...]

PR #1483, opened 2022-03-12
decoder/opus: Implement bitrate calculation

tsweet64:
This implements obtaining bitrates when playing Ogg/Opus files with the `opus` decoder plugin. The bitrate is calculated per ogg packet by obtaining its size in bits, multiplying by the sample rate, and dividing by the number of samples of audio decoded from said packet. It seems to work fine from my testing and the values returned appear accurate. It should not divide by zero since 0 `nframes` values result in a `return` prior to this line.

For reference, [here](https://github.com/xiph/opusfile/blob/cf218fb54929a1f54e30e2cb208a22d08b08c889/src/opusfile.c#L1814) is how `libopusfile` does it; similar calculations though it seems it might do it over a larger span than just one packet.

This would partially address #428, but I have not yet determined how to do this for the `vorbis` plugin.

### DIFF
--- a/src/decoder/plugins/OpusDecoderPlugin.cxx
+++ b/src/decoder/plugins/OpusDecoderPlugin.cxx
@@ -327,9 +327,15 @@ MPDOpusDecoder::HandleAudio(const ogg_packet &packet)
 			return;
 	}
 
-	// the size of the opus packet in bits / number of samples * samp/sec / 1000
+	/* Formula for calculation of bitrate of the current opus packet:
+	* bits_sent_into_decoder = packet.bytes * 8
+	* bytes_of_audio_decoded = nframes * frame_size
+	* samples_of_audio_decoded_per_channel = bytes_of_audio_decoded / channels / 2
+	* 1/seconds_decoded = opus_sample_rate / samples_of_audio_decoded_per_channel
+	* kbits = bits_sent_into_decoder * 1/seconds_decoded / 1000
+	*/
 	int channels = opus_packet_get_nb_channels((const unsigned char*)packet.packet);
-	uint16_t kbits = (unsigned int)packet.bytes*channels*8 * opus_sample_rate / (nframes * frame_size / 2) / 1000;
+	uint16_t kbits = (unsigned int)packet.bytes*8 * opus_sample_rate / (nframes * frame_size / channels / 2) / 1000;
 
 	/* apply the "skip" value */
 	if (skip >= (unsigned)nframes) {
--- a/src/decoder/plugins/OpusDecoderPlugin.cxx
+++ b/src/decoder/plugins/OpusDecoderPlugin.cxx
@@ -329,13 +329,10 @@ MPDOpusDecoder::HandleAudio(const ogg_packet &packet)
 
 	/* Formula for calculation of bitrate of the current opus packet:
 	* bits_sent_into_decoder = packet.bytes * 8
-	* bytes_of_audio_decoded = nframes * frame_size
-	* samples_of_audio_decoded_per_channel = bytes_of_audio_decoded / channels / 2
-	* 1/seconds_decoded = opus_sample_rate / samples_of_audio_decoded_per_channel
+	* 1/seconds_decoded = opus_sample_rate / nframes
 	* kbits = bits_sent_into_decoder * 1/seconds_decoded / 1000
 	*/
-	int channels = opus_packet_get_nb_channels((const unsigned char*)packet.packet);
-	uint16_t kbits = (unsigned int)packet.bytes*8 * opus_sample_rate / (nframes * frame_size / channels / 2) / 1000;
+	uint16_t kbits = (unsigned int)packet.bytes*8 * opus_sample_rate / nframes / 1000;
 
 	/* apply the "skip" value */
 	if (skip >= (unsigned)nframes) {
--- a/src/decoder/plugins/OpusDecoderPlugin.cxx
+++ b/src/decoder/plugins/OpusDecoderPlugin.cxx
@@ -327,6 +327,10 @@ MPDOpusDecoder::HandleAudio(const ogg_packet &packet)
 			return;
 	}
 
+	// the size of the opus packet in bits / number of samples * samp/sec / 1000
+	int channels = opus_packet_get_nb_channels((const unsigned char*)packet.packet);
+	uint16_t kbits = (unsigned int)packet.bytes*channels*8 * opus_sample_rate / (nframes * frame_size / 2) / 1000;
+
 	/* apply the "skip" value */
 	if (skip >= (unsigned)nframes) {
 		skip -= nframes;
@@ -361,7 +365,7 @@ MPDOpusDecoder::HandleAudio(const ogg_packet &packet)
 	const size_t nbytes = nframes * frame_size;
 	auto cmd = client.SubmitData(input_stream,
 				     data, nbytes,
-				     0);
+				     kbits);
 	if (cmd != DecoderCommand::NONE)
 		throw cmd;
 


src/decoder/plugins/OpusDecoderPlugin.cxx
```diff
@@ -327,6 +327,10 @@ MPDOpusDecoder::HandleAudio(const ogg_packet &packet)
 			return;
 	}
 
+	// the size of the opus packet in bits / number of samples * samp/sec / 1000
+	int channels = opus_packet_get_nb_channels((const unsigned char*)packet.packet);
+	uint16_t kbits = (unsigned int)packet.bytes*channels*8 * opus_sample_rate / (nframes * frame_size / 2) / 1000;
```
MaxKellermann, line 332: I ... don't understand this formula.
`packet.bytes*channels*8` is supposed the "size of the opus packet in bits", right? But ... why multiply with the number of channels?
I think you wrote `/1000` because MPD expects kilobits/s and not bits/s, right?
What is `nframes * frame_size / 2`? `nframes * frame_size` would be the uncompressed size of the `opus_decode()` result, but why do you need that? And what does the `/ 2` mean?
I'm very confused.

tsweet64: My bad, I could have done a better job documenting how this works. The basic idea of what this is trying to do is: (bits of input) \* (samples / second) / samples, yielding bits per second, which we divide by 1000 for kbits/second.

> But ... why multiply with the number of channels?

In theory this is supposed to be the number of bits of input that resulted in the corresponding decoded output. I'm honestly not sure why multiplying by the number of channels was necessary, but it was. I originally tried without this, but found that stereo files would report half their correct bitrate. With adding this, testing with both mono and stereo files yielded the correct results, so it seems to be necessary. There may be something weird about how ogg packages multiple channels or how the calling code reports the size.

> I think you wrote /1000 because MPD expects kilobits/s and not bits/s, right?

Yes.

> What is `nframes * frame_size / 2`?

This is the number of samples in the decoded audio. Given the size in bytes of the decoded audio, dividing by 2 gives the number of samples in 16-bit PCM since 2 bytes per sample.

MaxKellermann: Sorry, I still don't understand your explanation. None of this appears to make any sense. And since you're "honestly not sure" how this works, can't explain it, and did random things until it appeared to work, I don't have much trust in the correctness of your code.
To calculate the bit rate, you don't need the uncompressed size **at all**. If you need it, then your formula is wrong. It's as simple as that. Your formula calculates the uncompressed size, so it's wrong.
The bit rate formula is "number_of_bits_in_input_file / decoded_time". The output frame/sample size is nowhere to be found in this formula, don't you see? Why does your formula have it?

tsweet64: > And since you're "honestly not sure" how this works, can't explain it, and did random things until it appeared to work, I don't have much trust in the correctness of your code.

I have now figured out exactly why. Disregard what I said before about ogg packets. It's the output number of samples needs to be divided by the number of channels in order to obtain the correct time (since the two channels are played simultaneously), not the input that needs to be multiplied. It just turned out that multiplying the input size by this was mathematically equivalent to dividing the output size. I will push a new commit moving this value to the correct spot and adding better documentation. Sorry about the confusion.

Also, I have been testing this on many opus files for several days now and it's accurate as far as I can determine by comparison with what `mpv` reports, including reporting the exact intended bitrate of CBR files.

I'll explain the formula step-by-step, and with the units of the calculation resulting from each step

```
bits: number_of_bits_before_decoding = packet.bytes*8
bytes: number_of_bytes_of_audio_decoded = nframes * frame_size
samples: number_of_samples_of_audio_decoded_per_channel = number_of_bytes_of_audio_decoded / channels / 2     // because 16-bit PCM audio by definition contains exactly two bytes per sample.
1/seconds: reciprocal_of_decoded_time = opus_sample_rate / number_of_samples_of_audio_decoded
bits/seconds: bps = number_of_bits_before_decoding * reciprocal_of_seconds_of_audio_decoded
kilobits/seconds: kbits = bps / 1000
```

> The bit rate formula is "number_of_bits_in_input_file / decoded_time". The output frame/sample size is nowhere to be found in this formula, don't you see? Why does your formula have it?

Yes, of course, and that is exactly what I'm doing here. But we must calculate the decoded_time. We can't just pull this number out of thin air, we get it by analyzing the output. Also, the formula can also be number_of_bits \* 1 / decoded_time, as these are mathematically equivalent. That is the calculation being performed here.

>  To calculate the bit rate, you don't need the uncompressed size at all. If you need it, then your formula is wrong. It's as simple as that. Your formula calculates the uncompressed size, so it's wrong.

Again, the uncompressed size is used to calculate the decoded_time, which in turn is used to calculate the bit rate; it is not the bit rate itself. We can calculate the length in seconds of a segment of PCM audio based on the number of samples it contains and the sample rate. PCM audio provided by the opus decoder contains 48000 samples per second. This is constant. Each sample contains 2 bytes.

MaxKellermann: > Again, the uncompressed size is used to calculate the decoded_time

But you don't need it. You used it, but you don't need to use it, to make your formula much more complicated than necessary.

tsweet64: Can you suggest where else I could obtain the decoded_time? Perhaps I'm missing something.

src/decoder/plugins/OpusDecoderPlugin.cxx
```diff
@@ -327,9 +327,15 @@ MPDOpusDecoder::HandleAudio(const ogg_packet &packet)
...
+	/* Formula for calculation of bitrate of the current opus packet:
+	* bits_sent_into_decoder = packet.bytes * 8
+	* bytes_of_audio_decoded = nframes * frame_size
+	* samples_of_audio_decoded_per_channel = bytes_of_audio_decoded / channels / 2
+	* 1/seconds_decoded = opus_sample_rate / samples_of_audio_decoded_per_channel
+	* kbits = bits_sent_into_decoder * 1/seconds_decoded / 1000
+	*/
 	int channels = opus_packet_get_nb_channels((const unsigned char*)packet.packet);
-	uint16_t kbits = (unsigned int)packet.bytes*channels*8 * opus_sample_rate / (nframes * frame_size / 2) / 1000;
+	uint16_t kbits = (unsigned int)packet.bytes*8 * opus_sample_rate / (nframes * frame_size / channels / 2) / 1000;
```
MaxKellermann, line 338: Do you realize that `frame_size / channels / 2 == 1`?
`frame_size = channels * frame_size`, see:
https://github.com/MusicPlayerDaemon/MPD/blob/407fa2720a1beeda5f1ea6dd7a995f140cffe0de/src/decoder/plugins/OpusDecoderPlugin.cxx#L241
https://github.com/MusicPlayerDaemon/MPD/blob/407fa2720a1beeda5f1ea6dd7a995f140cffe0de/src/pcm/AudioFormat.hxx#L216-L219

... and you hard-code the sample size to `2`, so what you're effectively doing is `frame_size / frame_size`.

You want the number of frames, but you already have it (`nframes`), so you use it to calculate the uncompressed size (using the frame size), and then divide it by the frame size to get back to the number of frames. Duh!

src/decoder/plugins/OpusDecoderPlugin.cxx
MaxKellermann: `opus_sample_rate / nframes` - that's all.

src/decoder/plugins/OpusDecoderPlugin.cxx
tsweet64: Ohhhh, duh! I had not realized that, and that was the source of my confusion. I get it now, thank you and I apologize. Fixed commit coming right up... that really does make it a lot simpler.

MaxKellermann:
Squashed: 3a3f605a564bf5b05569ce3e52474775e89eb7bb

PR #1485, opened 2022-03-13
Remove bmp, tiff and add webp for coverimage filenames

jcorporation:
- supporting bmp and tiff seems outdated
- webp is more widely used for coverimages


MaxKellermann:
Cherry-picked to the stable branch: 5ad1a01d7ade6e6220f392c709808e3334cd01a7

PR #1513, opened 2022-04-18
src/lib/upnp: introduce Upnp namespace to fix link error with libnpupnp

aeolio:
libnpupnp declares a function trimstring() with scope external, which results in a link error due to multiple definitions of the same symbol. 

  [libnpupnp-4-2-1::src/inc/smallut.h::157]
  extern void trimstring(std::string& s, const char *ws = " \t");

  [mpd-0.23.6::src/lib/upnp/Utils.hpp::27]
  void
  trimstring(std::string &s, const char *ws) noexcept

Introduce a namespace for the mpd implementation of trimstring() to fix the following build error:

mips-buildroot-linux-musl/bin/ld: mips-buildroot-linux-musl/sysroot/usr/lib/libnpupnp.a(libnpupnp_la-smallut.o): in function `trimstring(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, char const*)':
smallut.cpp:(.text+0x114c): multiple definition of `trimstring(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, char const*)'; src/lib/upnp/libupnp.a.p/Util.cxx.o:Util.cxx:(.text._Z10trimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc+0x0): first defined here

### DIFF
--- a/src/lib/upnp/Device.cxx
+++ b/src/lib/upnp/Device.cxx
@@ -23,6 +23,8 @@
 
 #include <string.h>
 
+using namespace Upnp;
+
 /* this destructor exists here just so it won't get inlined */
 UPnPDevice::~UPnPDevice() noexcept = default;
 
--- a/src/lib/upnp/Util.cxx
+++ b/src/lib/upnp/Util.cxx
@@ -19,6 +19,8 @@
 
 #include "Util.hxx"
 
+namespace Upnp {
+
 /** Get rid of white space at both ends */
 void
 trimstring(std::string &s, const char *ws) noexcept
@@ -66,3 +68,5 @@ path_getfather(const std::string &s) noexcept
 	path_catslash(father);
 	return father;
 }
+
+} // namespace Upnp
--- a/src/lib/upnp/Util.hxx
+++ b/src/lib/upnp/Util.hxx
@@ -22,10 +22,14 @@
 
 #include <string>
 
+namespace Upnp {
+
 void
 trimstring(std::string &s, const char *ws = " \t\n") noexcept;
 
 std::string
 path_getfather(const std::string &s) noexcept;
+
+} // namespace Upnp
 
 #endif /* _UPNPP_H_X_INCLUDED_ */


neheb:
sounds like an npupnp bug.

aeolio:
It could be fixed in a similar way within npupnp also. Since npupnp has more (internal) dependencies, the scope of the change would be bigger.

neheb:
ping @medoc92

medoc92:
I suppose that you can't use a shared lib ? These symbols are not visible in the .so

neheb:
huh?

```
mangix@ryzen ~/d/npupnp (master)> strings ./.libs/libnpupnp.so | grep trimstring
_Z10trimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
_Z11rtrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
rtrimstring
_Z11ltrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
ltrimstring
_Z10trimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
_Z11ltrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
_Z11rtrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
```

neheb:
Interesting. So When compiling this normally I can't hit the error. But when passing -Ddefault_library=static I do.

medoc92:
About "strings" : that's not how you check for symbol visibility. Use nm or try a link.

I looked around, and there is a possible solution using partial linking, as described here:
https://stackoverflow.com/questions/3276474/symbol-hiding-in-static-libraries-built-with-xcode

You extract all the ,o from the .a, then perform a partial link, and make all the visibility=hidden so that the symbols which would be hidden in a .so are now local (invisible):

````
y$ ar x libnpupnp.a 
y$ ld -Ur -o libnpupnp.o libnpupnp_la-*.o
y$ nm libnpupnp.o | grep -i trimstring
00000000000149a0 T _Z10trimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
0000000000014900 T _Z11ltrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
0000000000001b26 t _Z11ltrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc.cold.91
0000000000014830 T _Z11rtrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
0000000000001b0c t _Z11rtrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc.cold.90
(reverse-i-search)`ob': git checkout inc/UpnpGl^Cal.h 
y$ objcopy --localize-hidden libnpupnp.o
y$ nm libnpupnp.o | grep -i trimstring
00000000000149a0 t _Z10trimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
0000000000014900 t _Z11ltrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
0000000000001b26 t _Z11ltrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc.cold.91
0000000000014830 t _Z11rtrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc
0000000000001b0c t _Z11rtrimstringRNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPKc.cold.90
````

Notice the switch from T to t.

You could recreate a .a as a final step if you really need it (e.g. rm libnpupnp.a; ar cr libnpupnp.a libnpupnp.o)

This seems to work fine for me, but more testing is probably needed to check that the result actually work (initialisations performed etc.)

neheb:
Wouldn't it be easier to enclose those functions in some Internal namespace in npupnp?

medoc92:
For modifying the code, I need a bit of time to think of the right way, and to release a new version. This will take time (I'll be unable to work on it in the following days). The partial link approach gives you a solution right now.

aeolio:
So essentially I have two options: (1) exclude npupnp from static builds or (2) implement a post-build /pre-install step that performs the object stripping and hope it works on all supported architectures? I still prefer my original solution, it is least-effort. 

I also looked at npupnp and decided not to do a fix within the library code, because it would have been more invasive (unless you limit it to trimstring(), which I consider cheating). 

This defect materialized in an automatic Buildroot test, by the way; the logs are here: http://autobuild.buildroot.net/results/a87/a8754431aad8808ae2d67e32d4a7973b00eda1c1/

medoc92:
The two options above are for fixing the problem "right now", not waiting for an npupnp code fix. It turns out that the latter is going to come faster than I expected yesterday.

There is a way to fix the problem with minuscule changes to npupnp, by just enclosing the smallut code in a namespace and add a "using" declaration to the include file itself. This way, (allmost) no changes are needed to the files actually calling the utilities, which is kind of important to me because there are a couple hundred (other projects).

It's sort of ugly and does not fix source level namespace pollution, but this is not the issue here (as smallut.h is not exported), and it does fix the library name space.

I just committed the change, and also issued libnpupnp 4.2.2 with it.

aeolio:
Tested and works, thank you all. 
I will close this pull request as not needed anymore.

neheb:
Sweet. I’ll update the meson wrap soon.

PR #1515, opened 2022-04-20
src/lib/icu: fix iconv() detection when libiconv is installed

aeolio:
iconv() integration may fail when libiconv is installed. In that case, the build may silently disable iconv or fail later during build, because -liconv is not specified.

Installation of libiconv hides the functionality provided by the C library (it overwrites usr/include/iconv.h and redefines symbols by redefining them, e.g. #define iconv_open libiconv_open)

Meson iconv() detection may accept a libiconf installation, but fails to provide an automatic dependency. The link stage subsequently fails with:

/nvmedata/autobuild/instance-29/output-1/host/lib/gcc/i686-buildroot-linux-uclibc/9.4.0/../../../../i686-buildroot-linux-uclibc/bin/ld: src/lib/icu/libicu.a.p/Converter.cxx.o: undefined reference to symbol 'libiconv_open'
/nvmedata/autobuild/instance-29/output-1/host/lib/gcc/i686-buildroot-linux-uclibc/9.4.0/../../../../i686-buildroot-linux-uclibc/bin/ld: /nvmedata/autobuild/instance-29/output-1/host/i686-buildroot-linux-uclibc/sysroot/usr/lib32/libiconv.so.2: error adding symbols: DSO missing from command line

This is probably also related to https://github.com/MusicPlayerDaemon/MPD/pull/928

Modify iconv() detection to first check for an installed libiconv and passing it as dependency to compiler.has_function(). The dependency object may be empty, which falls back to the current functionality of only detecting a builtin ficonv(). the libiconv dependency is propagated to the icu library, and thus integrated in the final build.

### DIFF
--- a/src/lib/icu/meson.build
+++ b/src/lib/icu/meson.build
@@ -12,17 +12,23 @@ if is_windows
   icu_sources += 'Win32.cxx'
 endif
 
+iconv_dep = []
 if icu_dep.found()
   icu_sources += [
     'Util.cxx',
     'Init.cxx',
   ]
 elif not get_option('iconv').disabled()
-  have_iconv = compiler.has_function('iconv', prefix : '#include <iconv.h>')
-  conf.set('HAVE_ICONV', have_iconv)
+  # an installed iconv library will make the builtin iconf() unavailable,
+  # so search for the library first and pass it as (possible) dependency
+  iconv_dep = compiler.find_library('libiconv', required: false)
+  have_iconv = compiler.has_function('iconv', 
+    dependencies: iconv_dep, 
+    prefix : '#include <iconv.h>')
   if not have_iconv and get_option('iconv').enabled()
     error('iconv() not available')
   endif
+  conf.set('HAVE_ICONV', have_iconv)
 endif
 
 icu = static_library(
@@ -31,6 +37,7 @@ icu = static_library(
   include_directories: inc,
   dependencies: [
     icu_dep,
+    iconv_dep,
     fmt_dep,
   ],
 )
--- a/src/lib/icu/meson.build
+++ b/src/lib/icu/meson.build
@@ -19,7 +19,7 @@ if icu_dep.found()
     'Init.cxx',
   ]
 elif not get_option('iconv').disabled()
-  # an installed iconv library will make the builtin iconf() unavailable,
+  # an installed iconv library will make the builtin iconv() unavailable,
   # so search for the library first and pass it as (possible) dependency
   iconv_dep = compiler.find_library('libiconv', required: false)
   have_iconv = compiler.has_function('iconv', 


src/lib/icu/meson.build
```diff
@@ -12,17 +12,23 @@ if is_windows
...
+iconv_dep = []
 if icu_dep.found()
   icu_sources += [
     'Util.cxx',
     'Init.cxx',
   ]
 elif not get_option('iconv').disabled()
-  have_iconv = compiler.has_function('iconv', prefix : '#include <iconv.h>')
-  conf.set('HAVE_ICONV', have_iconv)
+  # an installed iconv library will make the builtin iconf() unavailable,
```
neheb, line 22: iconf?

neheb:
sounds like a better solution than mine. cool.

src/lib/icu/meson.build
aeolio: Typo. I updated the pull request.

src/lib/icu/meson.build
```diff
@@ -12,17 +12,23 @@ if is_windows
...
     'Util.cxx',
     'Init.cxx',
   ]
 elif not get_option('iconv').disabled()
-  have_iconv = compiler.has_function('iconv', prefix : '#include <iconv.h>')
-  conf.set('HAVE_ICONV', have_iconv)
+  # an installed iconv library will make the builtin iconv() unavailable,
+  # so search for the library first and pass it as (possible) dependency
+  iconv_dep = compiler.find_library('libiconv', required: false)
+  have_iconv = compiler.has_function('iconv',
```
eli-schwartz, line 25: This may actually do odd things (probably just over-linking) on systems where the installed iconv library is split into dev / lib packages, and only the lib package is installed. The builtin header declarations would be used, but you check them via linking to the external library, and then add the external library as an mpd dependency.

There are also *really* fun edge cases alluded to in https://github.com/mesonbuild/meson/commit/1ac6fc0595f992621cbb79b7971ba07db1fa8e62 whereby trying to detect has_function('iconv') will misfire on MinGW because idk, __has_builtin returns true over there. This is why I implemented a custom snippet for that check.

aeolio: The commit you linked to is libintl; you probably mean mesonbuild/meson@1e19757

Using has_function() is indeed buggy, it leads to false positives on some systems, because, with with no `__builtin` but an installed library, the test code does nothing and just returns 0:

```
#ifdef __has_builtin`
  #if !__has_builtin({__builtin_}{func})
    #error "{__builtin_}{func} not found"
  #endif
#elif ! defined({func})
  {__builtin_}{func};
#endif
return 0;
```

This is why I check for the library first; it also ensures that the library is linked. Linking against libiconv is currently missing. 

The custom iconv dependency() clause seems to be a 0.60 feature, I do not know if MPD is willing to upgrade to a newer version. With meson < 0.60.0 the dependency() clause will have the same issues as the code currently used by MPD ...

eli-schwartz: Yeah, sorry for the mis-link. :)

PR #1518, opened 2022-04-23
subproject additions

neheb:
Tested on ARM32

Only issue is with npupnp: fixed with https://github.com/MusicPlayerDaemon/MPD/pull/1513

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -9,18 +9,7 @@ project(
     'cpp_std=c++17',
     'build.cpp_std=c++17',
     'warning_level=3',
-
-    # If we build those libraries as Meson subproject, they shall be
-    # linked statically into the MPD executable.
-    'expat:default_library=static',
-    'flac:default_library=static',
-    'fmt:default_library=static',
-    'gtest:default_library=static',
-    'libnpupnp:default_library=static',
-    'ogg:default_library=static',
-    'sqlite3:default_library=static',
-    'vorbis:default_library=static',
-    'zlib:default_library=static',
+    'default_library=static',
 
     # Not interested in compiler warnings from subprojects.
     'expat:werror=false',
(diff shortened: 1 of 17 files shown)


meson.build
```diff
@@ -9,18 +9,7 @@ project(
...
-    'expat:default_library=static',
-    'flac:default_library=static',
-    'fmt:default_library=static',
-    'gtest:default_library=static',
-    'libnpupnp:default_library=static',
-    'ogg:default_library=static',
-    'sqlite3:default_library=static',
-    'vorbis:default_library=static',
-    'zlib:default_library=static',
+    'default_library=static',
```
MaxKellermann, line 12: You removed the comment, leaving no explanation as to why this line is here. That's not acceptable.

src/lib/upnp/meson.build
```diff
@@ -14,13 +14,13 @@ if upnp_option == 'auto'
   upnp_dep = dependency('libupnp', version: '>= 1.8', required: false)
   conf.set('USING_PUPNP', upnp_dep.found())
   if not upnp_dep.found()
-    upnp_dep = dependency('libnpupnp', version: '>= 1.8', required: false)
```
MaxKellermann, line 17: Why did you drop the version requirement?

MaxKellermann:
Without the FLAC wrap, I can't cherry-pick the libogg wrap ...
```
flac-1.3.4/src/libFLAC/include/private/ogg_decoder_aspect.h:36:10: fatal error: 'ogg/ogg.h' file not found
```

MaxKellermann:
It's also obvious that you did not test those changes, because you removed variables which were still referenced by `{android/win32}/build.py`.
Enough.

PR #1519, opened 2022-04-24
fix -Wdeprecated warnings with C++20

neheb:
this must be explicit. The rest are C++11 deprecated warnings.


MaxKellermann:
As always, please tell exactly what warnings are fixed and explain these warnings and the "fixes" to me. I don't understand the diffs.

MaxKellermann:
After seeing and getting angry bout https://github.com/MusicPlayerDaemon/MPD/pull/1517 I decided spending more time on getting information from you is not worth my time. If you want to get your code merged, the PR should be properly documented from the start.

PR #1539, opened 2022-05-21
doc/user.rst: fix playlist plugin name option

naglis:
`playlist_plugin` blocks use `name` to identify the plugins.


MaxKellermann:
Cherry-picked to v0.23.x: bf9ffba4f768972b50f6b3f1c2d2847ac536c1ac

PR #1540, opened 2022-05-23
Fix deprecation warnings caused by name changes in OSX audio inerfaces

dhocker:
This should fix all of the deprecation warnings caused by audio interface name changes in macOS 12. I also had a fix for the OSX plugin compile error, but you already fixed that one. And, I also have a candidate fix for the snapcast compile error. If you need it I can create a PR for it.

As an aside, I have a mac MIni M1 and x86_64 mac so I can build MPD for both Apple hardware platforms.


MaxKellermann:
Cherry-picked to v0.23.x: c975d8b94316c86bf5950ed3abeba394e1263677

PR #1553, opened 2022-06-13
yajl/YajlResponseParser: add getJson()

HaikuZen:
added the filename folder.jpg to the list.

yajl/YajlResponseParser: now hold the jsonText, so clients can log if it's needed.


MaxKellermann:
Please remove the merge commit, it does't belong in a PR branch

MaxKellermann:
> yajl/YajlResponseParser: now hold the jsonText, so clients can log if it's needed.

This change adds a lot of overhead, but nobody uses the new field.

PR #1574, opened 2022-07-19
meson: use dependency('iconv')

neheb:
Properly deals with iconv, unlike the current solution. have_iconv fails
when libiconv CFLAGS are passed to the compiler. Tested under OpenWrt
with its CONFIG_BUILD_NLS, which adds libiconv include flags.

Requires meson version to be increased to 0.60.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2,7 +2,7 @@ project(
   'mpd',
   ['c', 'cpp'],
   version: '0.24',
-  meson_version: '>= 0.56.0',
+  meson_version: '>= 0.60.0',
   default_options: [
     'c_std=c11',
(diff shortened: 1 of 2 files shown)


MaxKellermann:
> Requires meson version to be increased to 0.60.

Not acceptable, because that prevents building with Debian Stable.

neheb:
> > Requires meson version to be increased to 0.60.
> 
> Not acceptable, because that prevents building with Debian Stable.

Debian 11 has meson 0.61.1

MaxKellermann:
That's bullseye-backports, not bullseye.
See https://packages.debian.org/bullseye/meson

PR #1575, opened 2022-07-20
meson: port ncpmc iconv solution

neheb:
Properly deals with iconv, unlike the current solution. have_iconv fails
when libiconv CFLAGS are passed to the compiler. Tested under OpenWrt
with its CONFIG_BUILD_NLS, which adds libiconv include flags.

https://github.com/MusicPlayerDaemon/ncmpc/commit/70a6519bd4b527b0cfde7343e2847be89e315506

### DIFF
--- a/src/lib/icu/meson.build
+++ b/src/lib/icu/meson.build
@@ -18,17 +18,25 @@ if icu_dep.found()
     'Util.cxx',
     'Init.cxx',
   ]
-elif not get_option('iconv').disabled()
-  # an installed iconv library will make the builtin iconv() unavailable,
-  # so search for the library first and pass it as (possible) dependency
-  iconv_dep = compiler.find_library('libiconv', required: false)
-  have_iconv = compiler.has_function('iconv', 
-    dependencies: iconv_dep, 
-    prefix : '#include <iconv.h>')
-  if not have_iconv and get_option('iconv').enabled()
-    error('iconv() not available')
+else
+  if meson.version().version_compare('>= 0.60')
+    iconv_dep = dependency('iconv', required: get_option('iconv'))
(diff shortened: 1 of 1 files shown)


MaxKellermann:
This fails with Meson 0.60+ if iconv is not available.

neheb:
I think this is what you meant.

eli-schwartz:
More or less. The disabled case still needs to set have_iconv to false so that conf.set sees that variable after dropping out of the if/elif, that's the cause of the CI failure.

src/lib/icu/meson.build
```diff
@@ -18,17 +18,26 @@ if icu_dep.found()
...
-  have_iconv = compiler.has_function('iconv', 
-    dependencies: iconv_dep, 
-    prefix : '#include <iconv.h>')
-  if not have_iconv and get_option('iconv').enabled()
-    error('iconv() not available')
+else
+  if meson.version().version_compare('>= 0.60')
+    iconv_dep = dependency('iconv', required: get_option('iconv'))
+    have_iconv = iconv_dep.found()
+    conf.set('HAVE_ICONV', have_iconv)
```
eli-schwartz, line 25: have_iconv is only used for conf.set, and for older meson versions to track whether another try should be done. For the 0.60.0 case, you don't need to set this bool variable at all, so you can just do:

```suggestion
    conf.set('HAVE_ICONV', iconv_dep.found())
```

neheb: Yeah. It''s from when I tried to keep changes to a minimum.

MaxKellermann:
Backported to v0.23.x: cc557c4d60f0b6545c6720ac0b11ddf03b1e5670

PR #1578, opened 2022-07-21
Add titleSort tag

Sonico98:
Hello!

With this PR I would like to add one more sorting tag to the list. Just like the already existing sorting tags, it can be used to force the way a client would sort a track within a list containing many others, like sorting "The Prayer" as simply "Prayer". In the same way, it can be used to easily sort and search for tracks that contain foreign characters: sorting "你好" as "Ni Hao" or "Hello", depending on how the user tagged the files. Finally, it allows properly sorting tracks containing roman numbers in their Title (displaying "Song IX" after "Song V", for example).

With the title of a song being the most useful way of differentiating one song from another, I think the inclusion of the tag is somewhat logical.


MaxKellermann:
Thanks, I merged it, even though your commit message was bad. Next time, please do not write any text into the PR, but instead write a proper commit message.

Sonico98:
Thanks for the feedback, I'll take it into account.

PR #1586, opened 2022-08-01
Provide forward and backward compatibility for macos audio API

dhocker:
This PR provides forward and backward compatibility at macos SDK 12.0. At SDK 12.0, API function names were changed essentially replacing occurrences of the word Master/master with Main/main. This change was test built on two different systems.

1. macos 10.15.7 with Xcode 12.4 and clang 12.0.0 on x86_64
2. macos 12.5 with Xcode 13.4.1 and clang 13.1.6 on arm64 (Apple silicon M1)

It should be noted that on macos 10.15.7 with Xcode 11.2 and clang 11.0, MPD will not build. The MPD documentation states that clang 11.0 is the minimum requirement, but clang 11.0 produces compile errors. Apparently the macos version of clang 11.0 is not fully compliant.

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -47,6 +47,15 @@
 #include <memory>
 #include <span>
 
+// Backward compatibility from OSX 12.0 API change
+#if (__MAC_OS_X_VERSION_MAX_ALLOWED >= 120000)
+	#define KAUDIO_OBJECT_PROPERTY_ELEMENT_MM kAudioObjectPropertyElementMain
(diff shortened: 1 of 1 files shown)


MaxKellermann:
This PR contains text which is missing in the commit message. PR text is lost after merging; please move the text to the commit message. A good PR doesn't need any text at all.

dhocker:
I updated the commit text from what I originally wrote in the PR. How I ended up with 3 commits is beyond me. git kept insisting that I merge the github branch back into my local branch which created an empty merge commit. I'm guessing you'll cherry pick the commit where the real changes are located.

MaxKellermann:
Cherry-picked to v0.23.x: bbc088ae4ea19767c102ca740765a30b98ffa96b

PR #1605, opened 2022-08-28
Add ConsumeMode oneshot, closes #925

### LINKED ISSUE #925: Support one-shot "consume" mode.
## Feature request
It would be useful to be able to toggle "consume" for a single track. This way, one can finish listening to the current song and be done with it, but allow the rest of the playlist to remain.

jcorporation:


src/queue/PlaylistState.cxx
```diff
@@ -94,7 +94,7 @@ playlist_state_save(BufferedOutputStream &os, const struct playlist &playlist,
 	os.Fmt(FMT_STRING(PLAYLIST_STATE_FILE_SINGLE "{}\n"),
 			  (int)playlist.queue.single);
 	os.Fmt(FMT_STRING(PLAYLIST_STATE_FILE_CONSUME "{}\n"),
-	       (unsigned)playlist.queue.consume);
+	       	  (int)playlist.queue.consume);
```
MaxKellermann, line 97: What's the point of changing this to `signed`, when the base type of your enum is `uint8_t`?

MaxKellermann:
Please add documentation

src/queue/PlaylistState.cxx
jcorporation: I copied it from playlist.queue.single. It is in both cases unnecessary?

MaxKellermann: Yes. There's not only no negative enum value, but those enums are explicitly declared to be unsigned. Casting them to signed makes no sense.

PR #1606, opened 2022-08-28
Add playlist_directory to config command response

jcorporation:


MaxKellermann:
Documentation, please

jcorporation:
Added.

MaxKellermann:
What happens if the user did not configure a playlist_directory?

jcorporation:
The value for playlist_directory is empty. Is this a problem for the mpd protocol? Is it better to do not print this configuration option?

MaxKellermann:
If no playlist is configured, then the path is "nulled", and is documented that it must not be used.
https://github.com/MusicPlayerDaemon/MPD/blob/5a7d2be77dffce47d2501cc03111e076be75097d/src/fs/AllocatedPath.hxx#L60-L61
The current `AllocatedPath` implementation does not crash, and there is no other debug-mode protection against this, but the code in this PR is wrong.
And yes, I think the setting should be omitted when there is no value for it. The same is being done for `music_directory`.

jcorporation:
Added `if (!spl_path.IsNull())`

PR #1618, opened 2022-09-18
Added command store {NAME} {REPLACE} to save the player queue to an existing playlist

geneticdrift:
by either replacing or appending to the existing list.

### DIFF
--- a/doc/protocol.rst
+++ b/doc/protocol.rst
@@ -1008,9 +1008,20 @@ remote playlists (absolute URI with a supported scheme).
 
 .. _command_save:
 
-:command:`save {NAME}`
+:command:`save {NAME} [MODE]`
     Saves the queue to
     `NAME.m3u` in the playlist directory.
+
+    ``MODE`` [#since_0_24]_
+        Optional argument. One of `create`, `append`, or `replace`.
+
+        `create`
+            The default. Create a new playlist.
+            Fail if a playlist with name ``NAME`` already exists.
+
+        `append`, `replace`
+            Append or replace an existing playlist.
+            Fail if a playlist with name ``NAME`` doesn\'t already exist.
 
 The music database
 ==================
--- a/src/PlaylistSave.cxx
+++ b/src/PlaylistSave.cxx
@@ -84,16 +84,25 @@ playlist_print_uri(BufferedOutputStream &os, const char *uri)
 }
 
 void
-spl_save_queue(const char *name_utf8, const Queue &queue)
+spl_save_queue(const char *name_utf8, PlaylistSaveMode save_mode, const Queue &queue)
 {
 	const auto path_fs = spl_map_to_fs(name_utf8);
 	assert(!path_fs.IsNull());
 
-	if (FileExists(path_fs))
-		throw PlaylistError(PlaylistResult::LIST_EXISTS,
-				    "Playlist already exists");
+	if (save_mode == PlaylistSaveMode::CREATE) {
+		if (FileExists(path_fs)) {
+			throw PlaylistError(PlaylistResult::LIST_EXISTS, "Playlist already exists");
+		}
+	}
+	else if (!FileExists(path_fs)) {
+		throw PlaylistError(PlaylistResult::NO_SUCH_LIST, "No such playlist");
+	}
 
-	FileOutputStream fos(path_fs);
+	FileOutputStream fos(path_fs,
+			     save_mode == PlaylistSaveMode::APPEND
+			     ? FileOutputStream::Mode::APPEND_EXISTING
+			     : FileOutputStream::Mode::CREATE);
+
 	BufferedOutputStream bos(fos);
 
 	for (unsigned i = 0; i < queue.GetLength(); i++)
@@ -106,7 +115,7 @@ spl_save_queue(const char *name_utf8, const Queue &queue)
 }
 
 void
-spl_save_playlist(const char *name_utf8, const playlist &playlist)
+spl_save_playlist(const char *name_utf8, PlaylistSaveMode save_mode, const playlist &playlist)
 {
-	spl_save_queue(name_utf8, playlist.queue);
+	spl_save_queue(name_utf8, save_mode, playlist.queue);
 }
--- a/src/PlaylistSave.hxx
+++ b/src/PlaylistSave.hxx
@@ -31,16 +31,19 @@ playlist_print_song(BufferedOutputStream &os, const DetachedSong &song);
 void
 playlist_print_uri(BufferedOutputStream &os, const char *uri);
 
-/**
- * Saves a queue object into a stored playlist file.
- */
+enum class PlaylistSaveMode {
+	CREATE,
+	APPEND,
+	REPLACE
+};
+
 void
-spl_save_queue(const char *name_utf8, const Queue &queue);
+spl_save_queue(const char *name_utf8, PlaylistSaveMode save_mode, const Queue &queue);
 
 /**
  * Saves a playlist object into a stored playlist file.
  */
 void
-spl_save_playlist(const char *name_utf8, const playlist &playlist);
+spl_save_playlist(const char *name_utf8, PlaylistSaveMode save_mode, const playlist &playlist);
 
 #endif
--- a/src/command/AllCommands.cxx
+++ b/src/command/AllCommands.cxx
@@ -184,7 +184,7 @@ static constexpr struct command commands[] = {
 	  handle_replay_gain_status },
 	{ "rescan", PERMISSION_CONTROL, 0, 1, handle_rescan },
 	{ "rm", PERMISSION_CONTROL, 1, 1, handle_rm },
-	{ "save", PERMISSION_CONTROL, 1, 1, handle_save },
+	{ "save", PERMISSION_CONTROL, 1, 2, handle_save },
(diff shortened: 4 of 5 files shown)


MaxKellermann:
I agree that both features added are useful, but I find the protocol design rather clumsy. We have "save" which does one thing, and now we have "store" (a word which has roughly the same subjective meaning as "save) which does two slightly different things, chosen by one obscure parameter.
I know the MPD protocol has accumulated quite a bit of obscurity and inconsistency across the last 19 years, but I think new commands should try to be better.
Maybe the mode should just be the second parametr to the existing "save" command? Just a rough draft:
- `save FOO` is the same as `save FOO create` (for backwards compatibility)
- `save FOO replace`
- `save FOO append` (does `append` allow creating a new file if it does not yet exist? Do we need both modes? Create-or-append and append-only-and-fail-if-not-exists?)

geneticdrift:
For `append` and `replace` I think the expectation is that the playlist `FOO` already exists. So following your draft I've changed the implementation to use the `save` command with an optional `MODE` argument, where `create` is the default, and the other optional values are `replace` and `append`.

jcorporation:
Great new feature. I had in the past already feature requests to implement this on client side.
As soon this feature is merged I will implement support for it in libmpdclient and in myMPD!

doc/protocol.rst
```diff
@@ -1006,21 +1006,20 @@ remote playlists (absolute URI with a supported scheme).
...
-:command:`save {NAME}`
+:command:`save {NAME} [MODE]`
     Saves the queue to
     `NAME.m3u` in the playlist directory.
 
     If a playlsit with name ``NAME`` already exists, the command fails with "Playlist already exists".
 
-.. _command_store:
+    ``MODE``
+        Optional argument [#since_0_xx_yy]_. One of `create`, `append`, or `replace`.
```
MaxKellermann, line 1016: since_0_xx_yy?

doc/protocol.rst
```diff
@@ -1006,21 +1006,20 @@ remote playlists (absolute URI with a supported scheme).
...
 
-:command:`store {NAME} {REPLACE}`
-    Saves the queue to an existing
-    `NAME.m3u` in the playlist directory.
-
-    If a playlsit with name ``NAME`` doesn\'t already exist, the command fails with "No such playlist".
+        `create`
+            The default. If a playlist with name ``NAME`` already exists, the command fails with "Playlist already exists".
```
MaxKellermann, line 1019: Actual error messages should not be documented, or else we can't ever rephrase them, because then clients will rely on the exact wording.

src/PlaylistSave.hxx
```diff
@@ -31,32 +31,38 @@ playlist_print_song(BufferedOutputStream &os, const DetachedSong &song);
...
-/**
- * Saves a queue object into a stored playlist file.
- */
-void
-spl_save_queue(const char *name_utf8, const Queue &queue);
+struct PlaylistSaveMode {
+private:
+	bool create = false;
+	bool replace = false;
+	bool append = false;
```
MaxKellermann, line 38: This is a rather complicated and bloated way to store this - why not an enum?

src/PlaylistSave.hxx
```diff
@@ -31,32 +31,38 @@ playlist_print_song(BufferedOutputStream &os, const DetachedSong &song);
...
- * Saves a playlist object into a stored playlist file.
- */
-void
-spl_save_playlist(const char *name_utf8, const playlist &playlist);
+	const char *argValidationError = "Expected one of 'create', 'append', 'replace'";
+
+public:
+	[[nodiscard]] static PlaylistSaveMode parseArg(const char* arg);
+
+	[[nodiscard]] inline bool is_valid() const noexcept {
```
MaxKellermann, line 45: Redundant `inline`, and please use CamelCase for method/class names. (Very old code often still uses lower-case with underscores.)

src/PlaylistSave.hxx
```diff
@@ -31,32 +31,38 @@ playlist_print_song(BufferedOutputStream &os, const DetachedSong &song);
...
+		return create || (append != replace);
+	}
+
+	[[nodiscard]] inline bool is_create_only() const noexcept {
+		return create && !(append || replace);
```
MaxKellermann, line 50: This... looks complicated.

src/PlaylistSave.hxx
```diff
@@ -31,32 +31,38 @@ playlist_print_song(BufferedOutputStream &os, const DetachedSong &song);
...
+};
 
-/**
- * Store a queue object into a stored playlist file.
- *
- * @arg replace true to relace the stored playlist contents, false to append to the list.
- */
 void
-spl_store_queue(const char *name_utf8, bool replace, const Queue &queue);
+spl_save_queue(const char *name_utf8, const PlaylistSaveMode& save_mode, const Queue &queue);
```
MaxKellermann, line 63: Pass this by value, not by reference - hey, even if you keep this as struct with 3 bools, it's just 3 bytes (1 bytes for an `enum : uint8_t`) - the reference will take 8 bytes for the pointer, plus the size of the struct!

src/PlaylistSave.cxx
```diff
@@ -83,17 +85,44 @@ playlist_print_uri(BufferedOutputStream &os, const char *uri)
...
+		if (StringIsEqual(arg, "create"))
+			mode.create = true;
+		else if (StringIsEqual(arg, "append"))
+			mode.append = true;
+		else if (StringIsEqual(arg, "replace"))
+			mode.replace = true;
+	}
+
+	if (!mode.is_valid())
+		throw MakeArgError(mode.argValidationError, arg == nullptr ? "" : arg);
```
MaxKellermann, line 103: This whole protocol-specific method living in a non-protocol-specific library is kind of a layering violating - I suggest we leave it that way for now, but please use `std::invalid_argument`; if caught by a command handler, it will be translated to ACK_ERROR_ARG, so that's fine:
https://github.com/MusicPlayerDaemon/MPD/blob/512cd7b0dea47883e54245122171a0178678e178/src/command/CommandError.cxx#L99-L100

src/PlaylistSave.hxx
```diff
@@ -31,32 +31,38 @@ playlist_print_song(BufferedOutputStream &os, const DetachedSong &song);
...
+
+	[[nodiscard]] inline bool is_valid() const noexcept {
+		return create || (append != replace);
+	}
+
+	[[nodiscard]] inline bool is_create_only() const noexcept {
+		return create && !(append || replace);
+	}
+
+	[[nodiscard]] inline bool is_existing_only() const noexcept {
```
MaxKellermann, line 53: More redundant `inline` - `inline` is implied for method bodies inside `struct{}`!

MaxKellermann (changes requested):
A few nitpicks.
The second commit is just a fixup for the first one; please fold them into one.

src/PlaylistSave.cxx
```diff
@@ -84,20 +84,29 @@ playlist_print_uri(BufferedOutputStream &os, const char *uri)
...
+	}
+	else if (!FileExists(path_fs)) {
+		throw PlaylistError(PlaylistResult::NO_SUCH_LIST, "No such playlist");
+	}
 
-	FileOutputStream fos(path_fs);
+	FileOutputStream fos(path_fs,
+						 save_mode == PlaylistSaveMode::APPEND
+										 ? FileOutputStream::Mode::APPEND_EXISTING
+										 : FileOutputStream::Mode::CREATE);
```
MaxKellermann, line 104: Bad indent

src/PlaylistSave.cxx
```diff
@@ -84,20 +84,29 @@ playlist_print_uri(BufferedOutputStream &os, const char *uri)
...
 	BufferedOutputStream bos(fos);
 
-	for (unsigned i = 0; i < queue.GetLength(); i++)
+	for (unsigned i = 0; i < queue.GetLength(); i++) {
 		playlist_print_song(bos, queue.Get(i));
+	}
```
MaxKellermann, line 109: Don't change existing code style (in an unrelated commit); this adds unnecessary noise to the diff.

src/PlaylistSave.cxx
```diff
@@ -106,7 +115,7 @@ spl_save_queue(const char *name_utf8, const Queue &queue)
 }
 
 void
-spl_save_playlist(const char *name_utf8, const playlist &playlist)
+spl_save_playlist(const char *name_utf8, PlaylistSaveMode save_mode, const playlist &playlist)
 {
-	spl_save_queue(name_utf8, playlist.queue);
+    spl_save_queue(name_utf8, save_mode, playlist.queue);
```
MaxKellermann, line 120: Bad indent

src/PlaylistSave.hxx
```diff
@@ -31,16 +31,16 @@ playlist_print_song(BufferedOutputStream &os, const DetachedSong &song);
 void
 playlist_print_uri(BufferedOutputStream &os, const char *uri);
 
-/**
- * Saves a queue object into a stored playlist file.
- */
```
MaxKellermann, line 36: Why delete API documentation?

MaxKellermann (changes requested):
Some coding style nitpicks; I think after that I'll merge it.
Before you re-push a PR branch, please always rebase on the official master branch.

PR #1627, opened 2022-09-27
Add pcre status to config response

jcorporation:
A way for mpd clients to detect pcre support for filters.


src/command/OtherCommands.cxx
```diff
@@ -395,6 +395,10 @@ handle_config(Client &client, [[maybe_unused]] Request args, Response &r)
 	if (const auto spl_path = map_spl_path(); !spl_path.IsNull())
 		r.Fmt(FMT_STRING("playlist_directory: {}\n"), spl_path.ToUTF8());
 
+#ifdef HAVE_PCRE
+	r.Write("pcre: true\n");
```
MaxKellermann, line 399: The protocol currently uses `1` for boolean true (e.g. consume/random/... mode); there is no precedent for the string `true`. I don't think `1` was a good choice, but I think we shouldn't introduce a new keyword here, as this only increases confusion. (Some of the protocol is already inconsistent and confusing enough.)

jcorporation: Is changed.

PR #1628, opened 2022-09-27
Add starts_with to filter expressions

jcorporation:


src/lib/icu/Compare.cxx
```diff
@@ -112,3 +112,32 @@ IcuCompare::IsIn(const char *haystack) const noexcept
...
+				       w_haystack.c_str(), -1,
+				       needle.c_str(), -1,
+				       nullptr,
+				       nullptr, nullptr, 0) == 0;
+	} catch (...) {
+		/* MultiByteToWideChar() has failed */
+		return false;
+	}
+#else
+	return strncmp(haystack, needle.c_str(), strlen(needle.c_str()));
```
MaxKellermann, line 141: That check was inverted; I fixed it.
Yeah, the C library is confusing - better use MPD's wrapper functions like `StringStartsWith()`.

PR #1634, opened 2022-10-01
meson.build: use the pkg-config file when attempting to detect sndio

brad0:


brad0:
ping.

MaxKellermann:
Your PR lacks information about which libsndio version is required for this to work.
From a first glance, your PR appears to break the build on Debian stable, and probably others.

MaxKellermann:
No reply

PR #1639, opened 2022-10-07
Add cache_directory to config command response

jcorporation:


MaxKellermann:
What's the point of that? No client should ever access this directory; the format of its content may change with any minor release.

jcorporation:
The mount commands only works with a configured cache_directory. Therefore it is good to know if it is configured.

MaxKellermann:
So you're not interested in the path of the directory, and not even whether there is one, all you want to know is whether mounting is possible?

jcorporation:
Exact this.

MaxKellermann:
For that, I think exposing the cache path is not a good way; on one hand, it will make clients believe they should mess with this directory, which they should not. On the other hand, future MPD versions may have different preconditions for mounting, and hard-coding this current requirement is more complex and more fragile than necessary.

Then the contract should better be: if `mount` appears in `commands`, it must be usable by the client. It does not currently work that way, but we should improve this. This should be true for all commands that have certain preconditions.

jcorporation:
You are right, this is the better approach.

PR #1640, opened 2022-10-07
add precondition for mount/unmount in commands response

jcorporation:
As discussed in #1639

### DIFF
--- a/src/db/plugins/simple/SimpleDatabasePlugin.hxx
+++ b/src/db/plugins/simple/SimpleDatabasePlugin.hxx
@@ -86,6 +86,10 @@ public:
 		return *root;
 	}
 
+	bool HasCache() const noexcept {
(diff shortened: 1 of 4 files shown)


src/db/plugins/simple/SimpleDatabasePlugin.hxx
```diff
@@ -86,6 +86,10 @@ public:
 		return *root;
 	}
 
+	const AllocatedPath GetCachePath() noexcept {
+		return cache_path;
+	}
```
MaxKellermann, line 91: This allocates a copy of the path; but you don't need that here; all you want is `bool HasCache() const noexcept`.

jcorporation: fixed

PR #1646, opened 2022-10-13
Added connect_time configuration to curl input plugin

geneticdrift:
Enables the user to override the hard coded 10 seconds curl connection timeout.

### DIFF
--- a/doc/plugins.rst
+++ b/doc/plugins.rst
@@ -223,21 +223,31 @@ variables such as ``http_proxy`` or specified in :file:`~/.curlrc`
 will be in effect.
 
 .. list-table::
-   :widths: 20 80
+   :widths: 20 70 10
    :header-rows: 1
 
    * - Setting
      - Description
+     - Default
    * - **proxy**
      - Sets the address of the HTTP proxy server.
+     -
    * - **proxy_user, proxy_password**
      - Configures proxy authentication.
+     -
    * - **verify_peer yes|no**
      - Verify the peer's SSL certificate? `More information <http://curl.haxx.se/libcurl/c/CURLOPT_SSL_VERIFYPEER.html>`_.
+     - yes
    * - **verify_host yes|no**
      - Verify the certificate's name against host? `More information <http://curl.haxx.se/libcurl/c/CURLOPT_SSL_VERIFYHOST.html>`_.
+     - yes
    * - **cacert**
      - Set path to Certificate Authority (CA) bundle `More information <https://curl.se/libcurl/c/CURLOPT_CAINFO.html>`_.
+     -
+   * - **connect_timeout**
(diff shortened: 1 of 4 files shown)


MaxKellermann:
You mixed up a few names. CURL calls it "connect timeout", your commit message says "connection time", and your option is named "connection timeout". What is a **connection** timeout? This is about the "connect" operation.

geneticdrift:
Corrected the name to `connect_timeout`

MaxKellermann:
This PR does two different things:
- allows the user to override CURL's default connect timeout
- if the user does not choose to overide CURL's default, the MPD will override CURL's default

The second part is "hidden" within this PR because your commit message does not mention it. If you suggest overriding CURL's default value even if the user has not expressed his will to do so, this is a different case; for now, please remove this. If you believe CURL's default should be lowered in any case, explain why, and we can discuss it, but should not be changed in the same commit.

geneticdrift:
The 10 seconds default is set here: https://github.com/MusicPlayerDaemon/MPD/blob/dda0dfc1408e137e11055fec43a7022a1c028aae/src/lib/curl/Setup.cxx#L47

geneticdrift:
I also just noticed that "0" is actually the `libcurl`'s default built-in connection timeout - 300 seconds and not "no timeout".
So I'll amend the PR again.

MaxKellermann:
> The 10 seconds default is set here:

You're right - though what I still don't like is that now we have 2 copies of this MPD-specific default override. If somebody later decides the default should be different, chances are low that he'll find both copies. I'll merge your PR for now, because that's only a minor problem that remains, but I care a lot about keeping the code maintainable for future developers.

geneticdrift:
I agree. Can't `Curl::Setup` be in-lined into `CurlRequest::SetupEasy`? I hesitate to make such changes without knowing the reasons behind commits such as https://github.com/MusicPlayerDaemon/MPD/commit/dda0dfc1408e137e11055fec43a7022a1c028aae

MaxKellermann:
> I agree. Can't `Curl::Setup` be in-lined into `CurlRequest::SetupEasy`? I hesitate to make such changes without knowing the reasons behind commits such as [dda0dfc](https://github.com/MusicPlayerDaemon/MPD/commit/dda0dfc1408e137e11055fec43a7022a1c028aae)

The `CurlRequest` class is used in several of my projects (e.g. https://github.com/XCSoar/XCSoar/blob/master/src/lib/curl/Request.cxx and https://github.com/CM4all/libcommon/blob/master/src/lib/curl/Request.cxx), and I moved the project-specific defaults to a separate source file to keep the diffs between those projects small. That way, I can cherry-pick relevant commits between those projects to keep them in sync. This is some kind of compile-time plugging, while also keeping `CurlRequest` generic and free of project-specific settings.

PR #1652, opened 2022-10-20
curl input plugin - added config options: verbose, low_speed_limit, low_speed_time, tcp_keepalive, tcp_keepidle, tcp_keepintvl

geneticdrift:
Use of these options can help diagnose network streaming issues or work around network interruptions and server dropped connections.

### DIFF
--- a/src/input/plugins/CurlInputPlugin.cxx
+++ b/src/input/plugins/CurlInputPlugin.cxx
@@ -154,6 +154,48 @@ static bool verify_peer, verify_host;
 /** Connection settings */
 static long connect_timeout;
 
+/**
+ * CURLOPT_VERBOSE - verbose mode
+ * DEFAULT 0, meaning disabled.
+ */
+static bool verbose = false;
+
+/**
+ * CURLOPT_LOW_SPEED_LIMIT - low speed limit in bytes per second
+ * DEFAULT 0, disabled
+ */
+static const unsigned default_low_speed_limit = 0;
+static long low_speed_limit = default_low_speed_limit;
+
+/**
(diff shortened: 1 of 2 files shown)


src/input/plugins/CurlInputPlugin.cxx
```diff
@@ -151,8 +151,54 @@ static const char *cacert;
...
+ * CURLOPT_VERBOSE - verbose mode
+ * DEFAULT 0, meaning disabled.
+ */
+static bool curl_opt_verbose = false;
+
+/**
+ * CURLOPT_LOW_SPEED_LIMIT - low speed limit in bytes per second
+ * DEFAULT 0, disabled
+ */
+constexpr unsigned default_curl_opt_low_speed_limit = 0;
```
MaxKellermann, line 171: Make those `static` so the compiler doesn't generate exported symbols

src/input/plugins/CurlInputPlugin.cxx
```diff
@@ -163,6 +209,9 @@ CurlInputStream::DoResume()
 {
 	assert(GetEventLoop().IsInside());
 
+	FmtDebug(curl_domain, "Resuming: capacity = {}, space = {}",
+		 GetBufferCapacity(), GetBufferSpace());
```
MaxKellermann, line 213: Is logging such a thing really use full? This happens all the time, and is part of the normal workflow.

MaxKellermann:
This commit does several things:
- add new options
- rename the variable for the existing timeout option
- add debug logging
- add `[[nodiscard]]` to existing methods

Separate commits for separate things, please - it's confusing to mix everything in one commit.

PR #1654, opened 2022-10-25
Minor changes in documentation

geneticdrift:
- suppress warnings.
- improve readability by eliminating unnecessary hyphenation (changed from auto to manual and align to left instead of justify).


MaxKellermann:
> Not sure if that's the right way to override just few styles

I don't know the right way either, but copying a large file from another project isn't acceptable. This adds a maintenance burden for keeping it in sync with upstream.

doc/user.rst
```diff
@@ -717,10 +717,10 @@ MPD enables MixRamp if:
 - Cross-fade is enabled
 - :ref:`mixrampdelay <command_mixrampdelay>` is set to a positive
   value, e.g.::
-    mpc mixrampdelay 1
+  mpc mixrampdelay 1
```
MaxKellermann, line 720: This change breaks the code block

MaxKellermann:
I cherry-picked a few commits to v0.23.x and master

geneticdrift:
I made a new PR just for the custom css. Hopefully the right way this time.

doc/user.rst
geneticdrift: Sorry.

PR #1658, opened 2022-10-30
queue/PlayListState: Setting consume or single to "oneshot" causes exception on server restart

rbackhouse:
e.g
server_socket: bind to '0.0.0.0:6601' failed (continuing anyway, because binding to '[::]:6601' succeeded): Failed to bind socket: Address already in use
output: No 'audio_output' defined in config file
output: Successfully detected a osx audio device
exception: Unrecognized single mode, expected 0, 1, or oneshot

It seems the the state restore code is not using the SingleToString and ConsumeToString functions.


rbackhouse:
Nevermind, fix seems to cause problems with values stored in state

Issue #608, opened 2019-07-26
[Feature request] Abillity to attach sticker data to song metadata instead of URI


JerwuQu:
Currently, song stickers are attached to song paths. For someone that (infrequently) reorganizes their collection, this is fragile. It would be good to have an option in MPD to translate any incoming sticker URIs to either
- A hashsum of the complete audio file
- A hashsum of the file's metadata section
- A composite of `artist`, `title`, and `album`

This is somewhat related to #114 and #484.

chaseadam:
My particular use case uses plugins to play items in the playlist (i.e. http locations) which are not represented in the database. It would be useful to have this concept of (sticker) metadata which can be associated to non-database items.

sp1ff:
Just noting that attaching the sticker to a hash of the file's metadata (or just artist, title & album) would break if the tags were updated.

jcorporation:
As #1659 is merged, can this issue be closed?

JerwuQu:
I haven't been using MPD for the last couple of years but it looks like the PR fulfills the original request, so closing.

Thanks! :)

Issue #114, opened 2017-09-21
[Request] Allow stickers to work with directory (and other) objects


carnager:
Originally stickers were meant to be used on any objects, like tags etc.
Sadly to this day the only supported object is "song" which associates to file uris.

I suggest using Directories as a first addition to the sticker api, since this seems the least work to add support for.

andyduke:
Stickers would be very useful for any objects, for example for storing additional information about playlists, etc.

gilphilbert:
It would be good to add stickers to albums (for ratings, etc) but also to artists - this would allow the sticker database to host artist profiles that can be pulled down from web APIs and stored locally without the need for additional external databases.

jcorporation:
As #1659 is merged, can this issue be closed?

jcorporation:
Implemented with https://github.com/MusicPlayerDaemon/MPD/commit/432675d4c23450c527333ed4a29d13f843e001da

PR #1659, opened 2022-10-30
Stickers on some tag types and playlists

geneticdrift:
This PR adds support for stickers on playlists and on selected Tag types.

The URI for tag stickers can be a filter expression to allow clients to specifically target objects when the tag value alone is ambiguous. For example when multiple albums have identical title (see examples in the doc).

Stickers cleanup for playlists is done from ```Instance::OnPlaylistDeleted```.

Cleanup for tag stickers is called from ```Instance::OnDatabaseModified``` but this will have to be changed because the performance when there are many stickers (filter match on the database for each tag sticker).

Related to issues #608 and #114

### DIFF
--- a/doc/protocol.rst
+++ b/doc/protocol.rst
@@ -1385,6 +1385,41 @@ Objects which may have stickers are addressed by their object
 type ("song" for song objects) and their URI (the path within
 the database for songs).
 
+.. note:: Since :program:`MPD` 0.24 stickers can also be attached to playlists,
+  some tag types, and :ref:`filter expressions <filter_syntax>`.
+  The following tag types are alloed: Title, Album, Artist, AlbumArtist, Genre,
+  Composer, Performer, Conductor, Work, Ensable, Location, and Label.
+
+.. list-table:: Sticker addressing
+   :widths: 10 45 45
+   :header-rows: 2
+
+   * - Type
+     - URI
+     - URI
+
+   * -
+     - get, set, delete, list, find
+     - find only
+
+   * - "song"
+     - File path within the database
+     - Directory path within the database to find a sticker on
+       all songs under this path recursively
+
+   * - "playlist"
+     - The playlist name of a stored playlist
+     - An empty string to find a sticker in all playlists
+
+   * - Tag type e.g. "Album"
+     - The tag value
+     - An empty string to find a sticker in all instances of the tag type
+
+   * - "filter"
+     - A :ref:`filter expression <filter_syntax>`.
+
+     - An empty string to find a sticker in all instances of the filter
+
 .. _command_sticker_get:
 
 :command:`sticker get {TYPE} {URI} {NAME}`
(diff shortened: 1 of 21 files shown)


geneticdrift:
Updated the PR:

1. Moved the stickers cleanup to a thread (deleting stickers that no longer match anything).
2. Separated tag stickers from filter stickers (tag-stickers only match by value and filter-stickers match by filter expression).
3. Normalized the filter-sticker URI by `filter->ToExpression()`

MaxKellermann:
Sorry for the review taking so terribly long. My time was extremely scarce and I avoided this PR because I knew it would take a lot of focus to understand it.

Stickers on playlists - that one is easy to understand, and I could easily merge that - unfortunately, it's not a separate commit, so I can't do that.

Stickers on tag types - what use is that? Say you add a sticker to "Album So" ... and now? Album names are ambiguous, maybe artist names are useful, but even then, what value does it add?

Stickers on filters - is that some kind of workaround for the problem of having multiple copies of the same song (or moving a song file around), and have them all carry the same stickers? If yes, then I might understand this feature - but it seems awfully complicated to use. Will anybody ever write a usable UI for this feature?

On "escape sticker name" - I'm not happy with this one. The escaping syntax for sticker names is not documented, and I'd rather restrict the allowed sticker names than adding some escaping rules. I mean, why would anybody really need `=` in a sticker name? Not restricting sticker names might have been an oversight in the initial design.

"GenParseName: added output message" .... I'd rather check this at compile time and at runtime, maybe do an assert(). Thanks to C++20, that is possible.

geneticdrift:
> Sorry for the review taking so terribly long. My time was extremely scarce and I avoided this PR because I knew it would take a lot of focus to understand it.

Understood and appreciated.

> Stickers on tag types - what use is that? Say you add a sticker to "Album So" ... and now? Album names are ambiguous, maybe artist names are useful, but even then, what value does it add?

The use cases for stickers on tag types include features such as favorites and ratings (e.g. 1-5 starts) which are useful not only on tracks/songs but also on other entities e.g. album, artist, work, directory, etc.
For example:
[Request] Allow stickers to work with directory (and other) objects #114
[Feature request] Abillity to attach sticker data to song metadata instead of URI #608 

Some clients implement these features locally. The downside is no sync between client installations and no interoperability with other clients (potentially with standardized sticker names).

Deciding which tag types this is useful for and for which it doesn't make sense is not obvious. But the implementation is one for all so maybe there's no point to have limited eligibility at all? I included a hard coded list of allowed tags to enable such restriction, but looking at it now I'm not sure is worth it, or at least review again which should be allowed.

> Album names are ambiguous

> Stickers on filters - is that some kind of workaround for the problem of having multiple copies of the same song (or moving a song file around), and have them all carry the same stickers? If yes, then I might understand this feature - but it seems awfully complicated to use. Will anybody ever write a usable UI for this feature?

My motivation for the sticker on a filter was the ```album + album-artist``` disambiguation case. Filters let clients the ability to disambiguate album names and if they do so it would be natural to use the same method to apply stickers to these 'virtual entities'.
Your point about moving a song file around is another use case (maybe by putting a sticker on the TRACK_ID tag).

> but it seems awfully complicated to use. Will anybody ever write a usable UI for this feature?

For GUI clients I don't think it's too complicated. Some already let the user choose to disambiguate albums by albumartist using a filtered queries, so for them it should be straight forward to implement a 'Favorite' feature using a filter sticker.
I think the point is that users and client implementers can come up with useful solutions that were not possible or easy before.


> On "escape sticker name" - I'm not happy with this one. The escaping syntax for sticker names is not documented, and I'd rather restrict the allowed sticker names than adding some escaping rules. I mean, why would anybody really need `=` in a sticker name? Not restricting sticker names might have been an oversight in the initial design.

Agree. Not a significant issue.

> "GenParseName: added output message" .... I'd rather check this at compile time and at runtime, maybe do an assert(). Thanks to C++20, that is possible.

Just something I tripped over and thought an explicit output message would have saved me some time figuring it out.

jcorporation:
I am the developer of the myMPD client. It would be great if this pull request is merged for the 0.24 release.

MaxKellermann:
432675d4c23450c527333ed4a29d13f843e001da
Merged with lots of cleanups.
Sorry for the extreeeeemyl long delay again. But at least it's been somewhat less than a year ;-)

PR #1661, opened 2022-11-03
systemd: use PrivateUsers= in user unit

bluca:
ProtectSystem= and other sandboxing options require a user namespace in order to work as user units (the user manager does not run as root and thus without a user namespace it is unable to perform mounts).

We are looking to enable this implicitly. Very few user units use these options, so want to make sure it is intentional and won't cause regressions. The alternative is to remove ProtectSystem=.

### DIFF
--- a/systemd/user/mpd.service.in
+++ b/systemd/user/mpd.service.in
@@ -19,6 +19,8 @@ LimitRTTIME=infinity
 # for io_uring
 LimitMEMLOCK=64M
 
+# Required in order for ProtectSystem= (and other sandboxing) to work
+PrivateUsers=yes
 # disallow writing to /usr, /bin, /sbin, ...
 ProtectSystem=yes
 


MaxKellermann:
Did you test whether this works on all setups? e.g. setups where user namespaces are forbidden, like Debian by default.

bluca:
Debian enables user namespaces too now:

```
$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
$ systemd-run -t --user --property PrivateUsers=yes --property ProtectSystem=yes ls /usr
Running as unit: run-u55756.service
Press ^] three times within 1s to disconnect TTY.
bin    i686-linux-gnu	 include  lib32  libexec  local  share	x86_64-linux-gnu
games  i686-w64-mingw32  lib	  lib64  libx32   sbin	 src	x86_64-w64-mingw32
```

In case they are not enabled, there is a graceful fallback and it is ignored.

bluca:
@MaxKellermann ping

PR #1664, opened 2022-11-08
Docs: added custom.css to override default hyphenation and text alignment

geneticdrift:
Followed [this sphinx doc](https://docs.readthedocs.io/en/stable/guides/adding-custom-css.html#adding-custom-css-or-javascript-to-sphinx-documentation) to add a custom css.

It overrides text alignment to left instead of justified, and hyphenation to manual instead of auto.

Seems to be more readable this way without words getting broken by an hyphen.


geneticdrift:
I was just adding another commit when you merged.
I hope it did not cause an unwanted commit to be merged.

PR #1667, opened 2022-11-10
Threading issues

geneticdrift:


MaxKellermann:
> There's no guarantee handler is not changed by multiple threads

If this is a problem, then your commit doesn't solve it.

geneticdrift:
> If this is a problem, then your commit doesn't solve it.

You are right. It only avoids a potential null dereferece. Probably not worth it.

MaxKellermann:
A nullptr dereference would actually be better than use-after-free, because it would guarantee to crash instead of doing something undefined. But the nullptr dereference would most likely not occur even without your patch, because the compiler would not load the variable twice from RAM (there's no barrier between the two accesses), and your commit does nothing about the potential use-after-free bug.

geneticdrift:
Understood. Thanks.

Issue #1546, opened 2022-06-02
Improve logging when playlist songs are missing


patagonaa:
## Feature request
I'd prefer if there were log messages when songs are missing when loading a playlist.

I had the issue today that no songs were added when I loaded a playlist.
When I looked at the (verbose) log, everything seemed fine:
```
client: [3] process command "load "music/playlists/something.m3u8""
client: [3] command returned 0
```
A bit of troubleshooting revealed that the playlist used `\` instead of `/` as a path separator which caused that issue.
It would be great if there was some kind of log message like `missing song "music\songs\song.mp3"` or something similar, then I might've caught the issue with the playlist right away.

PR #1671, opened 2022-11-16
Address issue #1546

sp1ff:
This PR updates `playlist_check_load_song()` to log a message (at level warn) on error to address [issue 1546](https://github.com/MusicPlayerDaemon/MPD/issues/1546)

### DIFF
--- a/src/playlist/PlaylistQueue.cxx
+++ b/src/playlist/PlaylistQueue.cxx
@@ -28,6 +28,7 @@
 #include "song/DetachedSong.hxx"
 #include "thread/Mutex.hxx"
 #include "fs/Traits.hxx"
+#include "Log.hxx"
 
 #ifdef ENABLE_DATABASE
 #include "SongLoader.hxx"
@@ -41,12 +42,14 @@ playlist_load_into_queue(const char *uri, SongEnumerator &e,
 			 playlist &dest, PlayerControl &pc,
 			 const SongLoader &loader)
 {
+	const unsigned max_log_msgs = 8;
+
 	const auto base_uri = uri != nullptr
 		? PathTraitsUTF8::GetParent(uri)
 		: ".";
 
 	std::unique_ptr<DetachedSong> song;
-	for (unsigned i = 0;
+	for (unsigned i = 0, failures = 0;
 	     i < end_index && (song = e.NextSong()) != nullptr;
 	     ++i) {
 		if (i < start_index) {
@@ -56,6 +59,12 @@ playlist_load_into_queue(const char *uri, SongEnumerator &e,
 
 		if (!playlist_check_translate_song(*song, base_uri,
 						   loader)) {
+			failures += 1;
+			if (failures < max_log_msgs) {
+				FmtError(playlist_domain, "Failed to load \"{}\".", song->GetURI());
+			} else if (failures == max_log_msgs) {
+				LogError(playlist_domain, "Further errors for this playlist will not be logged.");
+			}
 			continue;
 		}
 


.gitignore
```diff
@@ -9,3 +9,4 @@ __pycache__/
 
 /.clangd/
 /compile_commands.json
+TAGS
```
MaxKellermann, line 12: What is TAGS and why is it in the global .gitignore?

src/playlist/PlaylistSong.cxx
```diff
@@ -17,12 +17,14 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+#include "Log.hxx"
```
MaxKellermann, line 20: A file's own header must be included first.

src/playlist/PlaylistSong.cxx
```diff
@@ -54,18 +56,22 @@ merge_song_metadata(DetachedSong &add, const DetachedSong &base) noexcept
...
+	try {
+		DetachedSong tmp = loader.LoadSong(song.GetURI());
+
+		song.SetURI(tmp.GetURI());
+		if (!song.HasRealURI() && tmp.HasRealURI())
+			song.SetRealURI(tmp.GetRealURI());
+
+		merge_song_metadata(song, tmp);
+		return true;
+	} catch (const std::exception &ex) {
```
MaxKellermann, line 69: If the exception is not derived from `std::exception`, then MPD crashes.

src/playlist/PlaylistSong.cxx
```diff
@@ -54,18 +56,22 @@ merge_song_metadata(DetachedSong &add, const DetachedSong &base) noexcept
...
+		FmtWarning(playlist_domain,
```
MaxKellermann, line 70: If your playlist has a million songs that don't exist, this will produce a million log lines. That's why I never implemented this one - I feared this would spam the logs due to loading a broken playlist from the internet.

MaxKellermann:
Most of your commit messages are bad. For example, `Address issue #1546.` doesn't mean anything - it doesn't say what is being changed. If you mention a GH bug, add a "Closes" tag at the end of the commit message.

.gitignore
sp1ff: Apologies; this PR includes commits I did not intend to (including the one that added `TAGS` to this file). Will update.

src/playlist/PlaylistSong.cxx
```diff
@@ -17,12 +17,14 @@
...
  */
 
+#include "Log.hxx"
 #include "PlaylistSong.hxx"
 #include "SongLoader.hxx"
 #include "tag/Tag.hxx"
 #include "tag/Builder.hxx"
 #include "fs/Traits.hxx"
 #include "song/DetachedSong.hxx"
+#include "src/PlaylistError.hxx"
```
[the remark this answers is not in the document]
sp1ff: To pick up the declaration of `playlist_domain`.

src/playlist/PlaylistSong.cxx
sp1ff: Ah! Is it the case that exceptions _not_ derived from `std::exception` may be thrown in MPD? If so, I'll update.

src/playlist/PlaylistSong.cxx
sp1ff: OK, I see your point. Still, failing silently seems not ideal, either. Would you be open to logging at a lower verbosity? Or, what I think I would prefer, moving the log message further up and including logic to prevent this: log the first, say, 8 missing files, then a message along the lines of "further errors elided..."

sp1ff:
Update:
  - removed extraneous local commits
  - rate-limited logs

PR #1688, opened 2022-12-08
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,6 +26,9 @@ on:
       - master
       - v0.23.x
 
+permissions:
+  contents: read #  to fetch code (actions/checkout)
+
 jobs:
   build-linux:
     strategy:


MaxKellermann:
I don't understand, and I don't understand the documentation you linked to. What does it mean to grant "write" permissions? I don't find any explanation there.
Also please move your text from the PR to the commit message. PR texts get lost in GitHub's proprietary database, only commit messages remain forever to explain what was done and why.

sashashura:
Workflows triggered on any non `pull_request` event, in your case `push` or `workflow_dispatch` have unrestricted permissions to the repository and may modify commits, artifacts, releases, etc. By explicitly setting permissions you can restrict what the workflow can potentially do.
image

MaxKellermann:
Thanks. I did not know that actions can do all that, and it sounds like they shouldn't have those capabilities by default.
Is it enough to do that in the master branch?

sashashura:
Workflows on `push` start from the target branch, while others like on `comment` from the default branch. So it may be a good idea to change the repository defaults - https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions

PR #1690, opened 2022-12-13
Allow titlesort and composersort to fall back to title and composer

faceless2:
Unlike album, artist and albumartist, title and composer were not used as a fallback when titlesort and composersort were specified but unavailable - this patch fixes that.


MaxKellermann:
Please move the PR text to the commit message (to avoid it from getting lost in GH's proprietary database) and add documentation to https://github.com/MusicPlayerDaemon/MPD/blob/master/doc/protocol.rst#filters

faceless2:
Both done - this is beyond my git comfort level so hope I got this correct.

PR #1691, opened 2022-12-13
Add "searchcount" for case-insensitive count

faceless2:
The `count` command isn't case-insensitive. This PR adds `searchcount` which is a case-insensitive version of `count`.

### DIFF
--- a/doc/protocol.rst
+++ b/doc/protocol.rst
@@ -1286,6 +1286,16 @@ The music database
     The ``position`` parameter specifies where the songs will be
     inserted. [#since_0_23_4]_
 
+.. _command_searchcount:
+
+:command:`searchcount {FILTER} [group {GROUPTYPE}]`
+    Count the number of songs and their total playtime in
+    the database matching ``FILTER`` (see
+    :ref:`Filters <filter_syntax>`).
+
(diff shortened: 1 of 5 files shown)


doc/protocol.rst
```diff
@@ -1286,6 +1286,16 @@ The music database
...
     inserted. [#since_0_23_4]_
 
+.. _command_searchcount:
+
+:command:`searchcount {FILTER} [group {GROUPTYPE}]`
+    Count the number of songs and their total playtime in
+    the database matching ``FILTER`` (see
+    :ref:`Filters <filter_syntax>`).
+
+    Parameters have the same meaning as for :ref:`count <command_search>`
```
MaxKellermann, line 1296: Did you mean `command_count`?

MaxKellermann:
I cherry-picked the actual PR commit (ignoring the one that is duplicate from https://github.com/MusicPlayerDaemon/MPD/pull/1690), fixed the commit message (let's have all the information in git commit messages, not just in GitHub!), and added a NEWS line.

Note: I think it's fine to have unrelated commits in a single PR - if you submit many small things, that can save you some time, and saving time is good - but this one looked like it was accidental, because you submitted two separate PRs.

PR #1730, opened 2023-01-30
Fix deprecation warnings by replacing sprintf with snprintf

dhocker:
Fix deprecation warnings


MaxKellermann:
I don't want any new code using printf - let's rather migrate to libfmt, and MPD's wraper `FmtToBuffer` - https://github.com/MusicPlayerDaemon/MPD/blob/master/src/lib/fmt/ToBuffer.hxx

PR #1733, opened 2023-01-31
Decoder/FluidSynth: Allow gain to be adjusted

### LINKED ISSUE #929: Volume/gain adjustment for decoder plugins?
## Feature request

Currently playing tracker music/midis and possibly other non-pcm music gives very quiet sound compared to conventional music. I assume tracker/midi decoders just provide PCM with that volume, however there's no way to adjust that. Currently I have to manually decrease/increase volume in MPD or system when switching between tracks, which isn't ideal and it's easy to forget that you were playing midis which are quiet and need to lower the volume when switching to traditional music to avoid ear being blasted.

Playing Cave Story theme (modplug)
image

Playing GRABBAG.MID (fluidsynth, same results with wildmidi)
image

Playing Deus Ex theme (Title_musix.umx) (modplug)
image

Playing Grabbag.ogg from duke 3d megaton
image

Sean-Der:
Resolves #929

### DIFF
--- a/src/decoder/plugins/FluidsynthDecoderPlugin.cxx
+++ b/src/decoder/plugins/FluidsynthDecoderPlugin.cxx
@@ -21,6 +21,7 @@
 #include "../DecoderAPI.hxx"
 #include "pcm/CheckAudioFormat.hxx"
 #include "fs/Path.hxx"
+#include "lib/fmt/RuntimeError.hxx"
 #include "util/Domain.hxx"
 #include "Log.hxx"
 
@@ -29,6 +30,8 @@
 static constexpr Domain fluidsynth_domain("fluidsynth");
 
 static unsigned sample_rate;
+static double gain = 0.0;
+static bool gain_set = false;
 static const char *soundfont_path;
 
 /**
@@ -83,6 +86,17 @@ fluidsynth_init(const ConfigBlock &block)
 	soundfont_path = block.GetBlockValue("soundfont",
 					     "/usr/share/sounds/sf2/FluidR3_GM.sf2");
 
+	char *endptr;
+	const char *svalue = block.GetBlockValue("gain");
+	if (svalue != nullptr) {
+		gain = strtod(svalue, &endptr);
+		if (svalue == endptr || *endptr != 0) {
+			throw FmtRuntimeError(
+					"fluidsynth decoder gain value not a number: {}", svalue);
+		}
+		gain_set = true;
+	}
+
 	fluid_set_log_function(LAST_LOG_LEVEL,
 			       fluidsynth_mpd_log_function, nullptr);
 
@@ -93,6 +107,7 @@ static void
 fluidsynth_file_decode(DecoderClient &client, Path path_fs)
 {
 	char setting_sample_rate[] = "synth.sample-rate";
+	char setting_gain[] = "synth.gain";
 	/*
 	char setting_verbose[] = "synth.verbose";
 	char setting_yes[] = "yes";
@@ -109,6 +124,9 @@ fluidsynth_file_decode(DecoderClient &client, Path path_fs)
 		return;
 
 	fluid_settings_setnum(settings, setting_sample_rate, sample_rate);
+	if (gain_set) {
+		fluid_settings_setnum(settings, setting_gain, gain);
+	}
 
 	/*
 	fluid_settings_setstr(settings, setting_verbose, setting_yes);


src/decoder/plugins/FluidsynthDecoderPlugin.cxx
```diff
@@ -83,6 +84,14 @@ fluidsynth_init(const ConfigBlock &block)
 	soundfont_path = block.GetBlockValue("soundfont",
 					     "/usr/share/sounds/sf2/FluidR3_GM.sf2");
 
+	char *endptr;
+	const char *svalue = block.GetBlockValue("gain", 0.2);
```
MaxKellermann, line 88: Where does the `0.2` come from?
I'd prefer not to call `fluid_settings_setnum()` if the user did not specify the setting, and leave the default to the library.

Sean-Der: @MaxKellermann That is the default value from `FluidSynth` today https://www.fluidsynth.org/api/settings_synth.html#settings_synth_gain

This *should* have zero impact on users if the value isn't set

Sean-Der:
Sorry for all the pushes @MaxKellermann I am done now

The error handling works as expected `Jan 31 16:06 : exception: Failed to initialize decoder plugin 'fluidsynth'; fluidsynth decoder gain value not a number: foo`

and I am using it locally. thanks for the review!

src/decoder/plugins/FluidsynthDecoderPlugin.cxx
MaxKellermann: If a new library version changes the default, MPD should use the new default, and not stick to some old default value, so please don't hard-code those defaults in MPD.

src/decoder/plugins/FluidsynthDecoderPlugin.cxx
```diff
@@ -83,6 +85,14 @@ fluidsynth_init(const ConfigBlock &block)
 	soundfont_path = block.GetBlockValue("soundfont",
 					     "/usr/share/sounds/sf2/FluidR3_GM.sf2");
 
+	char *endptr;
+	const char *svalue = block.GetBlockValue("gain", "0.2");
+	gain = strtod(svalue, &endptr)
+	if (svalue == endptr || *endptr != 0) {
+		throw FmtRuntimeError(
+				"fluidsynth decoder gain value not a number: %s", svalue);
```
MaxKellermann, line 93: You can't use `%s` with libfmt!

Sean-Der: Fixed, properly uses `{}`

src/decoder/plugins/FluidsynthDecoderPlugin.cxx
Sean-Der: Fixed! Pulls and checks for `nullptr`

Sean-Der:
Mind looking again @MaxKellermann I have addressed both your comments, thank you!

src/decoder/plugins/FluidsynthDecoderPlugin.cxx
```diff
@@ -109,6 +122,9 @@ fluidsynth_file_decode(DecoderClient &client, Path path_fs)
 		return;
 
 	fluid_settings_setnum(settings, setting_sample_rate, sample_rate);
+	if (gain != 0.0) {
```
MaxKellermann, line 125: This will trigger `-Wfloat-equal`, because such a comparison is fragile. Can you use a "magic" value that's robust to compare, such as a negative value or NaN?

Sean-Der: @MaxKellermann What do you think of just doing `gain_set` I pushed a PR for it.

I first tried to do NaN, but I was worried about portability issues. The right thing to do is query `is_iec559` before using NaN, so I don't think I could depend on that.

Then magic values as well have some corner cases. If you would prefer I use a magic value happy to change it (what value should I do?)

MaxKellermann: `gain_set` is fine, it's just slightly more complicated than using `std::optional`.

MaxKellermann:
@Sean-Der, I merged this PR, but please write some documentation (`doc/plugins.rst`).

Sean-Der:
@MaxKellermann ACK! Should I switch to std optional? Sorry my C++ isn’t great :(

PR #1734, opened 2023-02-03
oss: Remove use of the soundcard.h header, OpenBSD no longer uses OSS

brad0:


MaxKellermann:
What does "no longer uses OSS" mean, and since when? Does MPD need to be compatible with older versions?

brad0:
> What does "no longer uses OSS" mean, and since when? Does MPD need to be compatible with older versions?

Sndio was created as a replacement for our former use of OSS and added 14 years ago. For OpenBSD that is 28 releases, approaching 29.

PR #1736, opened 2023-02-05
Resolve deprecation warnings by replacing use of sprintf with FmtBuffer

dhocker:
As I am sure you are aware, there some subtle and some not so subtle differences between sprintf format specs and libfmt format specs. libfmt seems to be better at picking reasonable format specs for cases where you simply use {} for substitution. I have no way to actually test some of the changes because I run on macOS Ventura 13.2. PidFile.hxx is the only change that I am sure was tested. To compensate for that problem, I wrote a small program where I tested the new format specs.

### DIFF
--- a/src/net/Resolver.cxx
+++ b/src/net/Resolver.cxx
@@ -34,6 +34,7 @@
 #include "AddressInfo.hxx"
 #include "HostParser.hxx"
 #include "lib/fmt/RuntimeError.hxx"
+#include "lib/fmt/ToBuffer.hxx"
 #include "util/CharUtil.hxx"
 
 #ifdef _WIN32
@@ -96,8 +97,9 @@ FindAndResolveInterfaceName(char *host, size_t size)
 	const unsigned i = if_nametoindex(interface);
 	if (i == 0)
 		throw FmtRuntimeError("No such interface: {}", interface);
-
-	sprintf(interface, "%u", i);
+	
+	StringBuffer ifx = FmtBuffer<64>("{d}", i);
+	strcpy(interface, ifx);
 }
 
 #endif
--- a/src/net/Resolver.cxx
+++ b/src/net/Resolver.cxx
@@ -98,7 +98,7 @@ FindAndResolveInterfaceName(char *host, size_t size)
 	if (i == 0)
 		throw FmtRuntimeError("No such interface: {}", interface);
 	
-	StringBuffer ifx = FmtBuffer<64>("{d}", i);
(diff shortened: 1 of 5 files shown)


src/net/Resolver.cxx
```diff
@@ -96,8 +97,9 @@ FindAndResolveInterfaceName(char *host, size_t size)
 	const unsigned i = if_nametoindex(interface);
 	if (i == 0)
 		throw FmtRuntimeError("No such interface: {}", interface);
-
-	sprintf(interface, "%u", i);
+	
+	StringBuffer ifx = FmtBuffer<64>("{0:d}", i);
+	strcpy(interface, ifx);
```
MaxKellermann, line 102: This is .... horrible!

src/pcm/AudioFormat.cxx
```diff
@@ -47,30 +48,32 @@ ToString(const AudioFormat af) noexcept
...
 
 	if (af.channels > 0)
-		p += sprintf(p, "%u", af.channels);
+		wp = FmtBuffer<24>("{}", af.channels);
 	else {
-		*p++ = '*';
-		*p = 0;
+		wp = FmtBuffer<24>("{}", "*");
 	}
+	strcpy(p, wp);
```
MaxKellermann, line 76: WTF! `strcpy()` just to avoid `sprintf()`?

src/pcm/MixRampGlue.cxx
```diff
@@ -37,8 +38,7 @@ StartToString(const MixRampArray &a) noexcept
 		if (i.time < FloatDuration{} || i == last)
 			continue;
 
-		char buffer[64];
-		sprintf(buffer, "%.2f %.2f;", i.volume, i.time.count());
+		StringBuffer buffer = FmtBuffer<64>("{0} {1};", i.volume, i.time.count());
```
MaxKellermann, line 41: This omits the precision

src/unix/PidFile.hxx
```diff
@@ -71,10 +72,9 @@ public:
 		if (fd < 0)
 			return;
 
-		char buffer[64];
-		sprintf(buffer, "%lu\n", (unsigned long)pid);
+		StringBuffer fb = FmtBuffer<64>("{}\n", (unsigned long)pid);
```
MaxKellermann, line 75: Why keep this cast?

MaxKellermann:
Cherry-picked, folded and removed bad changes. f248fe2decf2aa859ae3132ffac2a44b75ba2d48

PR #1747, opened 2023-02-23
Adding libsamplerate 0.2.2 to the win32 build & libs.py scripts

lerenwe:
I added the required entry to libs.py for libsamplerate and added it to the build script third party lib list.
That way, Windows user can use libsamplerate in addition to the internal resampler.

Tested the resulting python scripts myself on Debian 11 (bullseye), managed to compile mpd.exe without issue.

### DIFF
--- a/python/build/libs.py
+++ b/python/build/libs.py
@@ -16,6 +16,17 @@ libmpdclient = MesonProject(
     'lib/libmpdclient.a',
 )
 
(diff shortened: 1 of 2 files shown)


lerenwe:
As discussed here : #1746

MaxKellermann:
Cool, thanks!

PR #1760, opened 2023-03-14
Update user.rst

iconoclasthero:
MPD does not recognize file names in .mpdignore files that contain square brackets.  There may be other characters to which this also applies but I haven't found any.  FWIW '>' *does* work which is good because it is in a lot of bootleg song names.


MaxKellermann:
Bad commit message and bad documentation text.

PR #1779, opened 2023-04-11
Fixes hide_playlist_targets not working after server restart

datasone:
Currently, `hide_playlists_targets` works by skipping songs with `in_playlist` value set to true in [`Directory::Walk`](https://github.com/MusicPlayerDaemon/MPD/blob/a57bcd02382947abbd961594cdf00302c0c7866a/src/db/plugins/simple/Directory.cxx#L237). But `in_playlist` is not stored and only updated in [`UpdateWalk::PurgeDanglingFromPlaylists`](https://github.com/MusicPlayerDaemon/MPD/blob/a57bcd02382947abbd961594cdf00302c0c7866a/src/db/update/Playlist.cxx#L139), which will only be executed while updating DB.

This causes the problem that playlist target songs are correctly hidden after database update, but will remain visible after mpd server restarted. This pr solves the problem by storing `in_playlist` value of songs into the `SimpleDatabase` file.

### DIFF
--- a/src/SongSave.cxx
+++ b/src/SongSave.cxx
@@ -48,6 +48,8 @@ song_save(BufferedOutputStream &os, const Song &song)
 	if (song.audio_format.IsDefined())
 		os.Fmt(FMT_STRING("Format: {}\n"), song.audio_format);
 
+	os.Fmt(FMT_STRING("InPlaylist: {}\n"), song.in_playlist ? "yes" : "no");
+
 	if (!IsNegative(song.mtime))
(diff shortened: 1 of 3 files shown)


MaxKellermann:
838b19c2b514e7427449f06a04b743e0c75d9e59
Cherry-picked and amended with two changes:
- the commit message was bad; you wrote text in this PR but it was missing in the commit message; please do not wrote PR texts, because these explanations are lost after merging; all that counts is the commit message
- I changed your code to only emit the `InPlaylist` line if the value is `yes`; writing it always just adds unnecessary clutter if the feature is not used

PR #1787, opened 2023-04-20
decoder/openmpt: Fix build error with libopenmpt < 0.5

nomis:
openmpt_at_end is a string, not an integer


MaxKellermann:
I don't get it. Why did you (conditionally) change the type of the variable to `std::string`? And if you fix a build error, the commit message should at least describe what build error is being fixed.

MaxKellermann:
Simplified: ad8117e7c8dc0b8a4c1f6b252c881ba5f5d2ebef

nomis:
> I don't get it. Why did you (conditionally) change the type of the variable to `std::string`?

At the time I had assumed `ctl_set()` needed a non-temporary object as a reference, but the reference is `const`.

PR #1805, opened 2023-05-11
reorder ffmpeg to be lower priority than gme

fuck-shithub:
This should prevent ffmpeg from taking priority over the gme plugin. The ffmpeg plugin is more buggy than gme.
One of the prominent bugs of preferring ffmpeg over gme is that ffmpeg cannot seek SAP files while gme can. This should prevent that from happening.


MaxKellermann:
This change is good; FFmpeg should always be the last-resort fallback, and specialized plugins should be preferred.

Issue #1813, opened 2023-05-20
Plans for the next release


FtZPetruska:
Since the latest release (v0.23.12) does not build using fmt 10 (#1810, #1807), and since support for fmt 10 is added with large changes (move from custom `StringView` to `std::string_view`, and additional smaller patches), is there an expected timeframe for a new release ?

Alternatively, if v0.24 is not planned to be released soon, would there be an interest for a PR targeting the v0.23.x branch adding fmt 10 support with the goal of making a v0.23.13 release?

Thank you for your understanding.

MaxKellermann:
I wanted to release 0.24 instead of doing another 0.23 bug fix release, but this is taking longer than I wanted (as always), so I'm going to release 0.23.13 very soon.

Issue #1807, opened 2023-05-14
Build failure with libfmt 10


NexAdn:
### Describe the bug

Compile error with libfmt 10. Seems like there was a change in libfmt's API.

## Expected Behavior

Software compiles.

## Actual Behavior

Compile failure. Excerpt from the build output:

```
[5/99] c++ -Impd.p -I. -I.. -Isrc -I../src -I/usr/include/dbus-1.0 -I/usr/lib64/dbus-1.0/include -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Wpedantic -std=c++2a -O0 -g -ffast-math -ftree-vectorize -Wcast-qual -Wdouble-promotion -Wmissing-declarations -Wshadow -Wunused -Wvla -Wwrite-strings -Wno-stringop-overflow -fno-threadsafe-statics -fmerge-all-constants -Wcomma-subscript -Wextra-semi -Wmismatched-tags -Woverloaded-virtual -Wsign-promo -Wvolatile -Wvirtual-inheritance -Wno-non-virtual-dtor -Wsuggest-override -fvisibility=hidden -D_GNU_SOURCE -DPCRE2_CODE_UNIT_WIDTH=0 -MD -MQ mpd.p/src_command_PlayerCommands.cxx.o -MF mpd.p/src_command_PlayerCommands.cxx.o.d -o mpd.p/src_command_PlayerCommands.cxx.o -c ../src/command/PlayerCommands.cxx
FAILED: mpd.p/src_command_PlayerCommands.cxx.o 
[...]

PR #1814, opened 2023-05-20
[v0.23.x] Add fmt 10 support.

FtZPetruska:
This adds support for fmt 10 on the v0.23.x branch (see #1813).

This is done by adding formatter specialisation for the custom `StringView` and `StringBuffer` types.


MaxKellermann:
See https://github.com/MusicPlayerDaemon/MPD/issues/1807#issuecomment-1548490223 for why I don't think your approach is good.

PR #1817, opened 2023-05-21
Fix meson build warning for get_pkgconfig_variable

mbeijen:
Otherwise, building will generate these warnings:

systemd/system/meson.build:5: WARNING: Project targeting '>= 0.56.0' but tried to use feature deprecated since '0.56.0': Dependency.get_pkgconfig_variable. use Dependency.get_variable(pkgconfig : ...) instead
systemd/user/meson.build:5: WARNING: Project targeting '>= 0.56.0' but tried to use feature deprecated since '0.56.0': Dependency.get_pkgconfig_variable. use Dependency.get_variable(pkgconfig : ...) instead

### DIFF
--- a/systemd/system/meson.build
+++ b/systemd/system/meson.build
@@ -2,7 +2,7 @@ systemd_system_unit_dir = get_option('systemd_system_unit_dir')
 if systemd_system_unit_dir == ''
   systemd = dependency('systemd', required: false)
   if systemd.found()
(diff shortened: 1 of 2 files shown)


MaxKellermann:
v0.23.x: f7eb1c9a8324284b224e5adb2653eaf2f6ccf6e7

PR #1836, opened 2023-06-18
fix two compile warnings

critkitten:
fix two compile warnings


MaxKellermann:
Your code changes are bad: they do not only "fix" the warnings, but also change the program's behavior, and introduce two file descriptor leaks.

PR #1885, opened 2023-10-20
Add sort and window parameter for "sticker find"

### LINKED ISSUE #1103: Sort and limit results from sticker database
## Feature request

I'd like to be able to retrieve a set of stickers ordered in a descending list, preferably with a limit in place:

`sticker find {TYPE} {URI} {NAME} sort {SORT_DIRECTION}`
and
`sticker find {TYPE} {URI} {NAME} sort {SORT_DIRECTION} limit {LIMIT}`

For example:
`sticker find song "" playCount sort desc limit 100`

This would allow retrieval of stickers that are numeric or alphabetic and be able to pull the "top 100 played songs". It would also be possible to set a sticker for "date added" and retrieve a list of "50 most recent additions". This allows for dynamic playlists without needing to add extra databases.

I know we can use:
`sticker find {TYPE} {URI} {NAME}`

Then sort, but for large collections (10,000+ songs) this means manually sorting thousands of objects and discarding almost all of them when the SQLite database can do that for us quickly and easily

jcorporation:
First rough approach to implement it. It is only roughly tested, but I would request your feedback early.

closes #1103

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,7 @@ ver 0.24 (not yet released)
   - apply Unicode normalization to case-insensitive filter expressions
   - stickers on playlists and some tag types
   - new command "stickernames"
+  - "sticker find" supports sort and window parameter
 * database
   - attribute "added" shows when each song was added to the database
   - proxy: require MPD 0.21 or later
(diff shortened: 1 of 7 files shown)


MaxKellermann:
This leaks memory because you create new statements, but never free them.

jcorporation:
> This leaks memory because you create new statements, but never free them.

Fixed, replaced the reset and clear binding with finalize.

jcorporation:
#1895

PR #1910, opened 2023-12-10
First implementation of the `visualization` output plugin

sp1ff:
NOTE TO REVIEWERS: this was discussed [previously](https://github.com/MusicPlayerDaemon/MPD/pull/1449), but life got busy.

This commit contains the first implementation of a visualization output plugin. It servers a very simple network protocol, backed by a very simple audio analysis. I'm using it as my daily driver, but it needs review & hardening.

Update: Converting this to a draft while I work through the CI issues. Will squash all commits before marking "ready for review".

Update #2: CI passing, unit test suite considerably beefed-up. Using this build as my daily driver. @MaxKellermann ready for review.

### DIFF
--- a/src/lib/fmt/ThreadIdFormatter.hxx
+++ b/src/lib/fmt/ThreadIdFormatter.hxx
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: BSD-2-Clause
+// author: Max Kellermann <max.kellermann@gmail.com>
+
+#ifndef THREAD_ID_FORMATTER_HXX
+#define THREAD_ID_FORMATTER_HXX
+
+#include <fmt/format.h>
+#include <sstream>
+#include <thread>
+
+template<>
+struct fmt::formatter<std::thread::id> : formatter<string_view>
+{
+	template<typename FormatContext>
+	auto format(std::thread::id id, FormatContext &ctx) {
+		std::stringstream stm;
+		stm << id;
+		return formatter<string_view>::format(stm.str(), ctx);
+	}
+};
+
+#endif // THREAD_ID_FORMATTER_HXX
--- a/src/lib/fmt/ThreadIdFormatter.hxx
+++ b/src/lib/fmt/ThreadIdFormatter.hxx
@@ -12,7 +12,7 @@ template<>
 struct fmt::formatter<std::thread::id> : formatter<string_view>
 {
 	template<typename FormatContext>
-	auto format(std::thread::id id, FormatContext &ctx) {
+	auto format(std::thread::id id, FormatContext &ctx) const {
 		std::stringstream stm;
 		stm << id;
 		return formatter<string_view>::format(stm.str(), ctx);
--- a/src/output/plugins/meson.build
+++ b/src/output/plugins/meson.build
@@ -147,6 +147,28 @@ else
   wasapi_dep = dependency('', required: false)
 endif
 
+libfftw3_dep = dependency('fftw3f', version: '>= 3.3.8', required: get_option('fftw3'))
+output_features.set('ENABLE_FFTW3', libfftw3_dep.found())
+
+enable_visualization_output = get_option('visualization')
+conf.set('ENABLE_VISUALIZATION_OUTPUT', enable_visualization_output)
+
+output_features.set('ENABLE_VISUALIZATION_OUTPUT', get_option('visualization'))
+if get_option('visualization')
+  if not libfftw3_dep.found()
+    error('libfftw3 not available, but is required for the visualization plugin')
+  endif
+  output_plugins_sources += [
+    'visualization/VisualizationOutputPlugin.cxx',
+    'visualization/SoundAnalysis.cxx',
+    'visualization/SoundInfoCache.cxx',
+    'visualization/VisualizationServer.cxx',
+    'visualization/VisualizationClient.cxx',
+    'visualization/Protocol.cxx',
+  ]
+  output_plugins_deps += [ event_dep, net_dep, libfftw3_dep ]
+endif
+
 output_plugins = static_library(
   'output_plugins',
   output_plugins_sources,
--- a/src/output/plugins/visualization/LowLevelProtocol.hxx
+++ b/src/output/plugins/visualization/LowLevelProtocol.hxx
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The Music Player Daemon Project
+
+#ifndef LOW_LEVEL_PROTOCOL_HXX_INCLUDED
+#define LOW_LEVEL_PROTOCOL_HXX_INCLUDED
+
+#include "util/PackedBigEndian.hxx"
+
+#include <fftw3.h>
+
+#include <algorithm>
+#include <cstdint>
+#include <limits>
+
+namespace Visualization {
+
+/*  Write a uint16_t to an output iterator over byte in wire format; return the
+ *  iterator in its new position
+ */
+template <typename OutIter>
+OutIter
+SerializeU16(uint16_t n, OutIter pout) {
+	auto m = PackedBE16(n);
+	auto p = (std::byte*)(&m);
+	return std::copy(p, p + 2, pout);
+}
+
+static_assert(std::numeric_limits<float>::is_iec559);
+
+/* Convert an IEEE 754 single-precision floating-point number to wire format;
+ * write it to an output iterator & return the iterator in its new position
+ */
+template <typename OutIter>
+OutIter
+SerializeFloat(float f, OutIter pout) {
+	auto m = PackedBE32(*(uint32_t*)&f);
+	auto p = (std::byte*)(&m);
+	return std::copy(p, p + 4, pout);
+}
+
+/* Convert an fftwf_complex to wire format; write it to an output iterator &
+ * return the iterator in its new position
+ */
+template <typename OutIter>
+OutIter
+SerializeComplex(const fftwf_complex c, OutIter pout) {
+	auto r = PackedBE32(*(const uint32_t*)&(c[0]));
+	auto i = PackedBE32(*(const uint32_t*)&(c[1]));
+	auto pr = (std::byte*)(&r);
+	auto pi = (std::byte*)(&i);
+	pout = std::copy(pr, pr + 4, pout);
+	return std::copy(pi, pi + 4, pout);
+}
+
+} // namespace Visualization
+
+#endif // LOW_LEVEL_PROTOCOL_HXX_INCLUDED
--- a/src/output/plugins/visualization/Protocol.cxx
+++ b/src/output/plugins/visualization/Protocol.cxx
@@ -11,6 +11,7 @@ Visualization::ParseResult
 Visualization::ParseClihlo(void *data,
 			   size_t length,
 			   ClientHello &clihlo) noexcept {
+
 	// CLIHLO payload is 6 bytes, header & footer are five more.
 	if (length < sizeof(ClientHello) + 5) {
 		return ParseResult::NEED_MORE_DATA;
--- a/src/output/plugins/visualization/Protocol.cxx
+++ b/src/output/plugins/visualization/Protocol.cxx
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The Music Player Daemon Project
+
+#include "Protocol.hxx"
+
+#include "Log.hxx"
+#include "util/ByteOrder.hxx"
+#include "util/Domain.hxx"
+
+Visualization::ParseResult
+Visualization::ParseClihlo(void *data,
+			   size_t length,
+			   ClientHello &clihlo) noexcept {
+	// CLIHLO payload is 6 bytes, header & footer are five more.
+	if (length < sizeof(ClientHello) + 5) {
+		return ParseResult::NEED_MORE_DATA;
+	}
+
+	uint8_t *buf = (uint8_t *)data;
+
+	uint16_t msg_type = FromBE16(*(uint16_t *)buf);
+	if (msg_type != 0) {
+		return ParseResult::ERROR;
+	}
+
+	buf += 2;
+	uint16_t payload_len = FromBE16(*(uint16_t *)buf);
+	if (payload_len != 6) {
+		return ParseResult::ERROR;
+	}
+
+	buf += 2;
+	clihlo.major_version = *buf++;
+	clihlo.minor_version = *buf++;
+
+	clihlo.requested_fps = FromBE16(*(uint16_t *)(buf));
+	buf += 2;
+	clihlo.tau = FromBE16(*(int16_t *)(buf));
+	buf += 2;
+
+	if (*buf != 0) {
+		return ParseResult::ERROR;
+	}
+
+	return ParseResult::OK;
+}
--- a/src/output/plugins/visualization/Protocol.hxx
+++ b/src/output/plugins/visualization/Protocol.hxx
@@ -1,0 +1,138 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The Music Player Daemon Project
+
+#ifndef VISUALIZATION_PROTOCOL_HXX_INCLUDED
+#define VISUALIZATION_PROTOCOL_HXX_INCLUDED
+
+#include "LowLevelProtocol.hxx"
+#include "SoundAnalysis.hxx"
+
+#include <cstddef>
+#include <cstdint>
+
+namespace Visualization {
+
+/**
+ * \brief A parsed CLIHLO message
+ *
+ * \sa ParseCliHlo
+ *
+ *
+ * The visualization \ref vis_out_protocol "protocol" begins with the client
+ * connecting to the server & providing certain paramters of the sound analysis
+ * it would like to receive. That is done through the CLIHLO message (which see
+ * \a ref vis_out_protocol_proto_clihlo "here").
+ *
+ * See \a vis_out_protocol_timing "timing" for details on parameter tau.
+ *
+ *
+ */
+
+struct ClientHello {
+	/// Major protocol version the client would like to speak
+	uint8_t major_version;
+	/// Minor protocol version the client would like to speak
+	uint8_t minor_version;
+	/// The number of sound analyses per second the client would like to
+	/// receive (presumably the rate at which it is rendering frames, hence
+	/// the name "fps")
+	uint16_t requested_fps;
+	/// The desired offset (named "tau" in the documentation) between song
+	/// time and analysis time at each analysis performed
+	int16_t tau;
+};
+
+enum class ParseResult {
+	OK,
+	NEED_MORE_DATA,
+	ERROR,
+};
+
+/**
+ * \brief Attempt to parse a \ref vis_out_protocol_proto_clihlo "CLIHLO" message
+ * from the given buffer
+ *
+ * \param buf [in] An array of octets potentially containing the message
+ *
+ * \param length [in] The length of \a buf, in octets
+ *
+ * \param clihlo [out] A reference to a `client_hello_t` structure to be
+ * filled-in on successful execution
+ *
+ * \return ParseResult::OK if the message was successfully parsed,
+ * NEED_MORE_DATA if the message is incomplete, or ERROR if the message cannot
+ * be ready from \a buf
+ *
+ *
+ * CLIHLO is the first message in the protocol, sent by the client. See
+ * \ref vis_out_protocol_proto_clihlo "the protocol specification" for details,
+ * and \ref vis_out_protocol "Visualization Network Protocol" for discussion
+ * of the protocol altogether.
+ *
+ *
+ */
+
+ParseResult
+ParseClihlo(void *buf, size_t length, ClientHello &clihlo) noexcept;
+
+/// Serialize an SRVHLO message to wire format
+template <typename OutIter>
+void
+SerializeSrvhlo(std::byte major_ver, std::byte minor_ver, OutIter pout) {
+	using std::byte;
+
+	*pout++ = byte{0}; //
+	*pout++ = byte{1}; // message type
+	*pout++ = byte{0}; //
+	*pout++ = byte{2}; // payload length
+	*pout++ = major_ver;
+	*pout++ = minor_ver;
+	*pout++ = byte{0}; // check byte
+}
+
+/// Serialize a FRAME message header to wire format
+template <typename OutIter>
+OutIter
+SerializeSoundInfoFrameHeader(uint8_t num_chan,
+			      size_t num_samp,
+			      size_t num_freq,
+ OutIter pout) {
+
+	using std::byte;
+
+	// Start with the "magic number" allowing clients to "lock on" to the
+	// stream of sound info frames in the event of an error.
+	// See \ref vis_out_protocol_proto_msgs for details.
+	*pout++ = byte{0x63};
+	*pout++ = byte{0xac};
+	*pout++ = byte{0x84};
+	*pout++ = byte{0x03};
+
+	*pout++ = byte{16};
+	*pout++ = byte{0};
+
+	return SerializeU16(17 + 4 * num_chan * (num_samp + 3 * num_freq + 3),
+			    pout);
+}
+
+/// Serialize a FRAME message payload to wire format
+template <typename OutIter>
+void
+SerializeSoundInfoFrameFooter(OutIter pout) {
+	*pout = std::byte{0x00};
+}
+
+/// Serialize a FRAME message to wire format
+template <typename OutIter>
+void
+SerializeSoundInfoFrame(const Visualization::SoundAnalysis &a,
+			OutIter pout) {
+	pout = SerializeSoundInfoFrameHeader(a.NumChan(), a.NumSamp(),
+					     a.NumFreq(), pout);
+	pout = a.SerializeSoundInfoFramePayload(pout);
+	SerializeSoundInfoFrameFooter(pout);
+}
+
+} // namespace Visualization.
+
+#endif // VISUALIZATION_PROTOCOL_HXX_INCLUDED
--- a/src/output/plugins/visualization/SoundAnalysis.cxx
+++ b/src/output/plugins/visualization/SoundAnalysis.cxx
@@ -1,0 +1,424 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The Music Player Daemon Project
+
+#include "SoundAnalysis.hxx"
+
+#include "Log.hxx"
+#include "config/Block.hxx"
+#include "lib/fmt/AudioFormatFormatter.hxx"
+#include "lib/fmt/RuntimeError.hxx"
+#include "lib/fmt/ThreadIdFormatter.hxx"
+#include "pcm/FloatConvert.hxx"
+#include "util/Domain.hxx"
+
+#include <cassert>
+#include <climits>
+
+const Domain d_sound_analysis("sound_analysis");
+
+Visualization::SoundAnalysisParameters::SoundAnalysisParameters() noexcept
+: SoundAnalysisParameters(DEFAULT_NUM_SAMPLES, DEFAULT_LO_CUTOFF, DEFAULT_HI_CUTOFF)
+{ }
+
+Visualization::SoundAnalysisParameters::SoundAnalysisParameters(
+	const ConfigBlock &config_block)
+: SoundAnalysisParameters(
+	config_block.GetPositiveValue("num_samples", DEFAULT_NUM_SAMPLES),
+	config_block.GetPositiveValue("lo_cutoff", DEFAULT_LO_CUTOFF),
+	config_block.GetPositiveValue("hi_cutoff", DEFAULT_HI_CUTOFF))
+{ }
+
+Visualization::SoundAnalysisParameters::SoundAnalysisParameters(
+	size_t num_samples_in,
+	float lo_cutoff_in,
+	float hi_cutoff_in):
+	num_samples(num_samples_in),
+	lo_cutoff(lo_cutoff_in), hi_cutoff(hi_cutoff_in)
+{
+	if (lo_cutoff >= hi_cutoff) {
+		throw FmtRuntimeError(
+			"lo_cutoff ({}) must be less than hi_cutoff ({})",
+			lo_cutoff, hi_cutoff);
+	}
+}
+
+/**
+ * \page vis_out_dft The Discrete Fourier Transform & Frequency Analysis
+ *
+ * \section vis_out_dft_intro Introduction
+ *
+ * This page contains some notes on the Discrete Fourier Transform as applied to
+ * music. They are a combination of dimly-remembered mathematics from
+ * university, source code comments from the milkdrop Winamp visualization
+ * plug-in, and the fftw documentation.
+ *
+ * \section vis_out_dft_basics The Basics
+ *
+ * The first thing to note is that the human ear can perceive sounds in the
+ * range 200 - 20,000Hz. For visualization purposes, implementations tend to
+ * throw away frequency data above 10,000Hz or so since there's not much
+ * activity there (something I've observed myself).
+ *
+ * Perceptually, frequency is not linear, it's logarithmic. A change of one
+ * octave corresponds to a doubling in frequency. Intuitively, this means that
+ * the difference between, say, 200 & 300Hz is \em much greater than the
+ * difference betwen 5000 & 5100Hz, for example.
+ *
+ * \subsection vis_out_dft_dft The Discrete Fourier Transform
+ *
+ * Given \c n audio samples, sampled at a frquency of \c F Hz, the DFT computes
+ * \c n complex numbers, each of which corresponds to the frequency:
+ *
+ \code
+	    k * F
+    freq  = -----,  k=0,...,n-1
+	k     n
+
+ \endcode
+ *
+ * (see
+ * <a href="http://fftw.org/fftw3_doc/The-1d-Discrete-Fourier-Transform-_0028DFT_0029.html">here</a>).
+ *
+ * The DFT library I'm using (<a href="http://fftw.org">fftw</a> AKA "The
+ * Fastest Fourier Transform in the West") takes advantage of the Hermitian
+ * property of the Fourier Transform of real data in which the k-th Fourier
+ * coefficient is the complex conjugate of the the (n-k)-th coefficient and ony
+ * returns the first n/2+1 Fourier coefficients (i.e. indicies 0 to n/2,
+ * inclusive) to save time & space. See
+ * <a href="http://fftw.org/fftw3_doc/The-1d-Real_002ddata-DFT.html">here</a>.
+ *
+ * Therefore, the first Fourier coefficient returned corresponds to 0Hz, and the
+ * last to:
+ *
+ \code
+           n
+           - * F
+           2       F
+    freq = ----- = -
+             n     2
+ \endcode
+ *
+ * or half the sampling frequency.
+ *
+ * \subsection vis_out_dft_buckets How To Bucket Frequencies
+ *
+ * To divide frequency data into \c N bands (whether for a scheme like bass/
+ * mids/trebs, or into a number of bars for visualization purposes), consider
+ * your frequency range & compute the number of octaves therein. If we let \c n
+ * be the number of octaves, then we know:
+ *
+ \code
+      n     freq_hi        log(freq_hi/freq_lo)
+     2  :=  ------- => n = -------------------
+            freq_lo            log(2)
+ \encode
+ *
+ * The \c N bands will then be:
+ *
+ \code
+                          n/N
+    freq_lo...freq_lo * 2
+
+               n/N             2*n/N
+    freq_lo * 2  ...freq_lo * 2
+
+    ...
+               (N-1)*n/N           n
+    freq_lo * 2  ...    freq_lo * 2
+
+ \endcode
+ *
+ * \subsection vis_out_dft_eg Worked Example
+ *
+ * Let the number of samples n be 576. This means our dft will return n/2 + 1 =
+ * 289 complex numbers. Let our sampling frequency F be 44,100Hz. For each k,
+ * k=0...289, the corresponding frequency will be k * 44100/289, giving us a
+ * frequency range of 0Hz to 22,500Hz. Let's clamp that to 200-11,000, compute
+ * the power spectrum, and divide that power up into three bands: bass, mids &
+ * trebs.
+ *
+ * First, we need to find the indicies into the dft corresponding to our
+ * desired frequency range.
+ *
+ \code
+                       f  * n
+         k * F          k
+    f  = ----- ==> k = ------, where f := the frequency of the k-th
+     k     n              F           k   Fourier coefficient
+
+
+         | 200 * 576 |
+    k0 = | --------- | = floor(2.61...) = 2
+         |  44100    |
+          -         -
+
+          -           -
+         | 11000 * 576 |
+    k1 = | ----------- | = ceil(143.67...) = 144
+         |   44100     |
+
+ \endcode
+ *
+ * So the power spectrum will have 144 - 2 + 1 = 143 elements in it. Nb. we're
+ * throwing away roughly the upper-half of our frequency spectrum.
+ *
+ * To divide these frequencies into three bands such that each band contains the
+ * same number of octaves, we compute how many octaves there are in our
+ * frequency range (call this \c n):
+ *
+ \code
+
+     n   11000        log(11000/200)
+    2  = ----- => n = -------------- = 5.7814
+          200            log(2)
+ \endcode
+ *
+ * In other words, there are 5.7814 octaves in our chosen frequency range. We
+ * want to give each band 5.7814/3 = 1.9271 octaves. That means the three
+ * "buckets" will be:
+ *
+ \code
+                           1.9271
+    200 ........... 200 * 2        or   200 -    761Hz
+
+           1.9271         2*1.9271
+    200 * 2 ....... 200 * 2        or   761 -  2,892Hz
+
+           2*1.9271        5.7814
+    200 * 2 ....... 200 * 2        or 2,892 - 11,000Hz
+
+ \endcode
+ *
+ *
+ */
+
+Visualization::SoundAnalysis::SoundAnalysis(
+	const SoundAnalysisParameters &params,
+	std::shared_ptr<SoundInfoCache> pc)
+: num_samples(params.GetNumSamples()),
+  out_samples((num_samples / 2) + 1),
+  pcache(pc),
+  audio_format(pc->GetFormat()),
+  num_channels(audio_format.channels),
+  cbbuf(params.GetNumSamples() * audio_format.GetFrameSize()),
+  buf(new std::byte[cbbuf]),
+  in(fftwf_alloc_real(num_samples * num_channels), fftwf_free),
+  out(fftwf_alloc_complex(out_samples * num_channels), fftwf_free),
+  bass_mids_trebs(new float[3 * num_channels])
+{
+	if (num_samples > INT_MAX) {
+		throw FmtInvalidArgument(
+			"num_samples({}) may not be larger than {}",
+			num_samples, INT_MAX);
+	}
+
+	int n[] = { (int)num_samples };
+
+	/* The input is assumed to be interleaved; this seems convenient from
+	 * the perspective of how it's stored from the AudioOutput... tho if we
+	 * need an additional copy to convert it to `float`, we'd have the
+	 * opportunity to re-arrange it. */
+
+	int dist = num_samples;
+
+	/* Per the FFTW docs:
+	 *
+	 * "`rank` is the rank of the transform (it should be the size of the
+	 * array `*n`) we use the term rank to denote the number of independent
+	 * indices in an array. For example, we say that a 2d transform has rank
+	 * 2, a 3d transform has rank 3, and so on."
+	 *
+
+	 * This is always 1, for us.
+	 *
+	 * layout of `in`:
+	 *
+	 * | 0 ... num_samples-1 | num_samples ... 2*num_samples-1 | 2*num_samples ...
+	 * | data for chan 0	 | data for chan 1		   | data for chan 2 */
+
+	/* `howmany` is the number of transforms to compute. The resulting plan
+	 * computes `howmany` transforms, where the input of the k-th transform
+	 * is at location in+k*idist (in C pointer arithmetic), and its output
+	 * is at location out+k*odist. */
+
+	int odist = (num_samples / 2) + 1;
+
+	plan = fftwf_plan_many_dft_r2c(1,             // rank of the input array-- we have one-dimensional arrays
+				       n,	      // the number of elements in each array
+				       num_channels,  // one array for each channel
+				       in.get(),      // input buffer-- need to copy samples here before executing
+				       NULL,
+				       1,	      // input stride
+				       dist,	      // distance between successive arrays (indexes, not bytes)
+				       out.get(),     // output buffer-- overwritten on each execution
+				       NULL,
+				       1,	      // output stride
+				       odist,	      // distance between successive arrays (indexes, not bytes)
+				       FFTW_ESTIMATE);// should probably be zero (to select a more exhaustive
+	// search), but out of an abundance of caution, tell
+	// FFTW to pick a plan quickly
+	if (NULL == plan) {
+		throw FmtRuntimeError("Failed to generate an FFTW plan: "
+				      "num_samp={},num_chan={}",
+				      num_samples, num_channels);
+	}
+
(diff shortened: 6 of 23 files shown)


MaxKellermann:
Hey, I'm happy you continued that effort. That is certainly a useful feature for many MPD users.
I hope I can take time for a proper review soon, but meanwhile, please check out your coding style:
- indent with tabs! Your code tabs and spaces at random
- return type should be on its own line as the length of C++ types can get out of hand and this would add indentation to all parameters
- for multi-line comments, use multi-line comment syntax (`/* ... */`) and not muliple instances of the single-comment syntax (`//`)
- MPD has meanwhile switched to SPDX copyright headers, please do the same for your new code
- in some comments, you have non-ASCII emojis - I prefer to leave the code plain ASCII, except if there are good reasons to use Unicode, but those emojis don't look particularly useful
- functions/method/structs should be CamelCase, same for source file names
- if a function never throws, decorate it with `noexcept`
- top-level definitions (e.g. structs) have pretty generic names in our code, but they may conflict with structs of the same name in other compilation units, conflicting each other's symbol namespaces; better use prefixed names, or even better: put everything in a C++ namespace

src/output/plugins/visualization/VisualizationServer.cxx
```diff
@@ -0,0 +1,113 @@
...
+{
+	FmtNotice(vis_server_domain, "VisualizationServer::ReapClients({}, "
+			  "{} clients)", gettid(), clients.size());
+
+	for (auto p0 = clients.begin(), p1 = clients.end(); p0 != p1; ) {
+		auto p = p0++;
+		if ((*p)->IsClosed()) {
+			LogInfo(vis_server_domain, "Reaping closed client.");
+			clients.erase(p);
+		}
```
MaxKellermann, line 57: This loops endlessly if at least one client isn't closed, doesn't it?

src/output/plugins/visualization/VisualizationServer.hxx
```diff
@@ -0,0 +1,129 @@
...
+    };
+    /// Present state-- 👇 means closed, 👇 means opened (the plugin, that is)
+    std::variant<std::monostate, HavePcmData> state;
+	/// maximum number of clients permitted; zero => unlimited
+	size_t max_clients;
+
+	/// Clients have both a reference to the PCM cache as well as a
+	/// SoundAnalysis instance while the plugin is opened. We'll create new
+	/// clients with our present state
+	std::list<std::unique_ptr<VisualizationClient>> clients;
```
MaxKellermann, line 104: What's the point of wrapping the item type in `unique_ptr`? All this does is add a layer of indirection, doesn't it?

src/output/plugins/visualization/VisualizationServer.hxx
```diff
@@ -0,0 +1,129 @@
...
+	/// invoked periodically to clean-up dead clients
+	CoarseTimerEvent reaper;
```
MaxKellermann, line 106: Why have a timer, why not remove dead clients right away?

src/output/plugins/visualization/protocol.hxx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/// Serialize a FRAME message header to wire format
+template <typename OutIter>
+OutIter serialize_sound_info_frame_header(uint8_t num_chan, size_t num_samp,
+                                          size_t num_freq, OutIter pout) {
+
+    using std::byte;
+
+    constexpr byte zero = byte{0};
```
MaxKellermann, line 88: This coding style looks clumsy. This is easier:
```
constexpr byte zero{0};
```
Why the redundant `byte`?

src/output/plugins/visualization/protocol.hxx
```diff
@@ -0,0 +1,122 @@
...
+    constexpr byte h1 = byte{0xac};
+    constexpr byte h2 = byte{0x84};
+    constexpr byte h3 = byte{0x03};
+
+    *pout++ = h0;
+    *pout++ = h1;
+    *pout++ = h2;
+    *pout++ = h3;
+
+    *pout++ = sixteen;
```
MaxKellermann, line 101: I don't even get why you declare symbolic constants when all the name does is describe the integer value in numbers. Why not just:
```
*pout++ = std::byte{16};
```

src/output/plugins/visualization/VisualizationServer.cxx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+void VisualizationServer::OnAccept(UniqueSocketDescriptor fd,
+                                   SocketAddress /*address*/,
+                                   int) noexcept
+{
+	FmtInfo(vis_server_domain, "VisualizationServer::OnAccept({})", gettid());
+
+    // Can we allow an additional client?
+	if (max_clients && clients.size() > max_clients) {
```
MaxKellermann, line 98: Off-by-one error? This allows `max_clients+1` clients.

src/output/plugins/visualization/SoundInfoCache.hxx
```diff
@@ -0,0 +1,115 @@
...
+	double secs_per_frame;
+	/// Sample size, in bytes
+	unsigned frame_size;
+	// Mutex guarding the ring buffer since instances will be accessed from
+	// multiple threads
+	mutable Mutex mutex;
+	/// ring buffer size/capacity, in bytes
+	std::size_t cb_ring;
+	/// this is the ring buffer
+	std::unique_ptr<uint8_t[]> ring;
```
MaxKellermann, line 82: MPD has `class AllocatedArray` which wraps these two fields nicely - it could be used to simplify your code.

src/output/plugins/visualization/SoundInfoCache.hxx
```diff
@@ -0,0 +1,115 @@
...
+ * 
+ * In the end I decided to just write an application-specific implementation.
+ *
+ *
+ */
+
+class SoundInfoCache {
+public:
+	typedef std::chrono::steady_clock::duration Duration;
+	typedef std::chrono::microseconds Time;
```
MaxKellermann, line 68: I don't get these two types. One is called "duration" and it represents a duration. The other is called "time" and it represents ..... a time? Huh?

src/output/plugins/visualization/SoundInfoCache.hxx
```diff
@@ -0,0 +1,115 @@
...
+	// multiple threads
+	mutable Mutex mutex;
+	/// ring buffer size/capacity, in bytes
+	std::size_t cb_ring;
+	/// this is the ring buffer
+	std::unique_ptr<uint8_t[]> ring;
+	/// # of bytes currently in the ring buffer (as distinct from capacity)
+	std::size_t cb;
+	/// Valid PCM data exists in buf[p0, p1)
+	ptrdiff_t p0, p1;
```
MaxKellermann, line 86: Why is this a signed integer?

src/output/plugins/visualization/VisualizationServer.cxx
sp1ff: I don't believe so... the iterator `p0` is unconditionally postincremented on line 53, so the iteration will always make progress whether or not anyone is closed.

src/output/plugins/visualization/VisualizationServer.hxx
sp1ff: Yes. I wrapped it because `VisualizationClient`, being a `Buffered Socket` is not copy constructable. Your comment made me realize I could emplace elements in the list rather than push them. I've removed the `unique_ptr`.

src/output/plugins/visualization/VisualizationServer.hxx
sp1ff: In order to remove dead clients right away, the server would need a means of notification. Other plugins use a "back reference" from client to server as this means. This, however, introduces a cyclic dependency between the two, which I prefer to avoid.
I wrote-up my thoughts [here](https://github.com/MusicPlayerDaemon/MPD/pull/1910/files#diff-90c68e20e2a42d1047afc0974fb97e354104c84a775ed9d3bbc039eae021b3f9R415).

src/output/plugins/visualization/protocol.hxx
sp1ff: I try to avoid literals in code (well, other than their declaration). Perhaps I've adhered too literally to a guideline. Changed.

src/output/plugins/visualization/VisualizationServer.cxx
sp1ff: Yes-- fixed.

src/output/plugins/visualization/SoundInfoCache.hxx
sp1ff: IME "time" generally represents a singular moment in time; 07:23:00 December 17, 2023, for instance. A _duration_ is the span of time between two such moments; 10 seconds, e.g.
Do you think I should rename one or both types?

src/output/plugins/visualization/SoundInfoCache.hxx
sp1ff: No good reason; must have been an artifact of a first implementation. Changed, in any event.

src/output/plugins/visualization/SoundInfoCache.hxx
sp1ff: OK-- will look.

MaxKellermann: It is mostly the same as unique_ptr, but also remembers the size of the array allocation. Thus, you have one field instead of two.

src/output/plugins/visualization/SoundInfoCache.hxx
MaxKellermann: In C++-speak, what you call "time" is a "time point", and C++ has a dedicated type for that (called time_point). You however (ab)use the "duration" type to express a time point. But how can you express a time point in milliseconds? What's the reference time point? It's not documented in your code.

MaxKellermann:
When you fix PR commits, please amend them instead of adding fixup commits. I don't want to merge known-bad commits; a PR is "good" when all individual commits are "good".

src/output/plugins/meson.build
```diff
@@ -160,6 +160,21 @@ else
   wasapi_dep = dependency('', required: false)
 endif
 
+libfftw3_dep = dependency('fftw3f', version: '>= 3.3.10', required: get_option('fftw3'))
+output_features.set('ENABLE_FFTW3', libfftw3_dep.found())
+output_features.set('ENABLE_VISUALIZATION_OUTPUT', get_option('visualization'))
+if get_option('visualization') and libfftw3_dep.found()
```
MaxKellermann, line 166: This logic seems somewhat wrong. The user asks Meson to enable `visualization`, but then Meson doesn't find libfftw and disables the plugin, even though the user's request could not be fulfilled.

src/output/plugins/visualization/LowLevelProtocol.hxx
```diff
@@ -0,0 +1,54 @@
...
+
+#ifndef LOW_LEVEL_PROTOCOL_HXX_INCLUDED
+#define LOW_LEVEL_PROTOCOL_HXX_INCLUDED
+
+#include <fftw3.h>
+
+#include <algorithm>
+#include <cstdint>
+
+#include <netinet/in.h>
```
MaxKellermann, line 12: That header isn't portable, is it?
I suggest using MPD's `util/ByteOrder.hxx` header, which is portable and even faster!
It also has a `PackedBE32` class which you can use to write big-endian 32 bit integers to a byte stream (without reverting to `std::copy()`).

src/output/plugins/visualization/LowLevelProtocol.hxx
```diff
@@ -0,0 +1,54 @@
...
+ */
+template <typename OutIter>
+OutIter
+SerializeU16(uint16_t n, OutIter pout) {
+	auto m = htons(n);
+	auto p = (std::byte*)(&m);
+	return std::copy(p, p + 2, pout);
+}
+
+/* Convert an IEEE 754 single-precision floating-point number to wire format;
```
MaxKellermann, line 27: Please add a `static_assert` to ensure that this platform really uses IEEE754 (see https://en.cppreference.com/w/cpp/types/numeric_limits/is_iec559). All commonly used do, but I want this to be safe just in case somebody compiles MPD on some very exotic future CPU architecture.

src/output/plugins/visualization/LowLevelProtocol.hxx
```diff
@@ -0,0 +1,54 @@
...
+ * write it to an output iterator & return the iterator in its new position
+ */
+template <typename OutIter>
+OutIter
+SerializeFloat(float f, OutIter pout) {
+	auto m = htonl(*(uint32_t*)&f);
```
MaxKellermann, line 33: This cast is undefined behavior, isn't it?

src/output/plugins/visualization/Protocol.hxx
```diff
@@ -0,0 +1,103 @@
...
+};
+
+enum class ParseResult {
+	OK,
+	NEED_MORE_DATA,
+	ERROR,
+};
+
+/**
+ * \brief Attempt to parse a CLIHLO message from the given buffer
```
MaxKellermann, line 29: What is CLIHLO?

src/output/plugins/visualization/Protocol.hxx
```diff
@@ -0,0 +1,103 @@
...
+#include <cstddef>
+#include <cstdint>
+
+namespace Visualization {
+
+struct ClientHello {
+	uint8_t major_version;
+	uint8_t minor_version;
+	uint16_t requested_fps;
+	int16_t tau;
```
MaxKellermann, line 19: What do these fields mean?

src/output/plugins/visualization/Protocol.cxx
```diff
@@ -0,0 +1,45 @@
...
+	clihlo.requested_fps = ntohs(*(uint16_t *)(buf));
+	buf += 2;
+	clihlo.tau = ntohs(*(int16_t *)(buf));
+	buf += 2;
+
+	if (*buf != 0) {
+		return ParseResult::ERROR;
+	}
+
+	return ParseResult::OK;
```
MaxKellermann, line 43: All of this fragile could be eliminated if `struct ClientHello` were a big-endian struct, e.g. by using `PackedBE16` instead of `uint16_t`.

src/output/plugins/visualization/Protocol.hxx
```diff
@@ -0,0 +1,103 @@
...
+OutIter
+SerializeSoundInfoFrameHeader(uint8_t num_chan, size_t num_samp,
+							  size_t num_freq, OutIter pout) {
+
+	using std::byte;
+
+	*pout++ = byte{0x63};
+	*pout++ = byte{0xac};
+	*pout++ = byte{0x84};
+	*pout++ = byte{0x03};
```
MaxKellermann, line 75: What are these numbers?

src/output/plugins/visualization/Protocol.hxx
```diff
@@ -0,0 +1,103 @@
...
+	*pout = std::byte{0x00};
+}
+	
+/// Serialize a FRAME message to wire format
+template <typename OutIter>
+void
+SerializeSoundInfoFrame(const Visualization::SoundAnalysis &a,
+								OutIter pout) {
+	pout = SerializeSoundInfoFrameHeader(a.NumChan(), a.NumSamp(),
+											 a.NumFreq(), pout);
```
MaxKellermann, line 96: Indent wrong

src/output/plugins/visualization/Protocol.hxx
```diff
@@ -0,0 +1,103 @@
...
+void
+SerializeSoundInfoFrameFooter(OutIter pout) {
+	*pout = std::byte{0x00};
+}
+	
+/// Serialize a FRAME message to wire format
+template <typename OutIter>
+void
+SerializeSoundInfoFrame(const Visualization::SoundAnalysis &a,
+								OutIter pout) {
```
MaxKellermann, line 94: Indent wrong

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+
+const Domain d_sound_analysis("sound_analysis");
+
+Visualization::SoundAnalysisParameters::SoundAnalysisParameters() noexcept
+: SoundAnalysisParameters(DEF_NUM_SAMPLES, DEF_LO_CUTOFF, DEF_HI_CUTOFF)
+{ }
+
+Visualization::SoundAnalysisParameters::SoundAnalysisParameters(const ConfigBlock &config_block):
+	SoundAnalysisParameters(config_block.GetPositiveValue("num_samples", DEF_NUM_SAMPLES),
+							config_block.GetPositiveValue("lo_cutoff", DEF_LO_CUTOFF),
```
MaxKellermann, line 23: Indent wrong

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+  num_channels(audio_format.channels), 
+  cbbuf(params.GetNumSamples() * audio_format.GetFrameSize()), 
+  buf(new std::byte[cbbuf]),
+  in(fftwf_alloc_real(num_samples * num_channels), fftwf_free),
+  out(fftwf_alloc_complex(out_samples * num_channels), fftwf_free),
+  bass_mids_trebs(new float[3 * num_channels])
+{
+	if (num_samples > INT_MAX) {
+		throw FmtInvalidArgument("num_samples({}) may not be larger than {}",
+								 num_samples, INT_MAX);
```
MaxKellermann, line 204: Indent wrong

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+	}
+	
+	int n[] = { (int)num_samples };
```
MaxKellermann, line 207: Why is this signed?

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+
+	/* The input is assumed to be interleaved; this seems convenient from the
+	 * perspective of how it's stored from the AudioOutput... tho if we need an
+	 * additional copy to convert it to `float`, we'd have the opportunity to
+	 * re-arrange it. */
+
+	int dist = num_samples;
```
MaxKellermann, line 214: Why is this signed?

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+	 * 
+	 * | 0 ... num_samples-1 | num_samples ... 2*num_samples-1 | 2*num_samples ...
+	 * | data for chan 0	 | data for chan 1				   | data for chan 2 */
+ 
+	/* `howmany` is the number of transforms to compute. The resulting plan
+	 * computes `howmany` transforms, where the input of the k-th transform is at
+	 * location in+k*idist (in C pointer arithmetic), and its output is at
+	 * location out+k*odist. */
+
+	int odist = (num_samples / 2) + 1;
```
MaxKellermann, line 235: Why is this signed?

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+								   NULL,
+								   1,			  // input stride
+								   dist,		  // distance between successive arrays (indexes, not bytes)
+								   out.get(),	  // output buffer-- overwritten on each execution
+								   NULL,
+								   1,			  // output stride
+								   odist,		  // distance between successive arrays (indexes, not bytes)
+								   FFTW_ESTIMATE);// should probably be zero (to select a more exhaustive
+												  // search), but out of an abundance of caution, tell
+												  // FFTW to pick a plan quickly
```
MaxKellermann, line 250: Wrong indent

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+	float max_freq = (ns - 1.0f) * samples_per_sec / ns;
+	if (max_freq < params.GetHiCutoff()) {
+		FmtWarning(d_sound_analysis, "Clamping max frequency from {} to {}", freq_hi, max_freq);
+		freq_hi = max_freq;
+	} else {
+		freq_hi = params.GetHiCutoff();
+	}
+
+	idx_lo = (size_t) floor(freq_lo * (float) num_samples / samples_per_sec );
+	idx_hi = (size_t) ceil(freq_hi * (float)num_samples / samples_per_sec);
```
MaxKellermann, line 270: Why use `double` functions when the parameter is `float`? This adds an implicit cast and uses a function that may be slower than necessary.

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+
+	float num_octaves = log(freq_hi/freq_lo) / 0.69314718f;
```
MaxKellermann, line 272: Same here: implicit cast to `double` and another implicit cast back to `float`.

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+
+	float freq_mids	 = freq_lo * pow(2.0f, num_octaves / 3.0f);
+	float freq_trebs = freq_lo * pow(2.0f, 2.0f * num_octaves / 3.0f);
```
MaxKellermann, line 275: More `double`/`float` mixups

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+
+	idx_mids  = ns *  freq_mids / samples_per_sec;
+	idx_trebs = ns * freq_trebs / samples_per_sec;
+}
+
+bool
+Visualization::SoundAnalysis::Update(SoundInfoCache::Time t) noexcept
+{
+	FmtDebug(d_sound_analysis, "SoundAnalysis::update(tid: {}), time {}us, # samp: {}, "
+			 "buffer size: {}", gettid(), t.count(), num_samples, pcache->Size());
```
MaxKellermann, line 285: Wrong indent

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+
+	if (!pcache->GetByTime(num_samples, t, buf.get(), cbbuf)) {
+		FmtWarning(d_sound_analysis, "Failed to get samples by time for sound analysis "
+				   "({} samples requested, at time {}us for buf size {}).", num_samples, 
+				   t.count()/1000, cbbuf);
```
MaxKellermann, line 290: Wrong indent

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+	 * than I store, I expect this to be more efficient (both in terms of time &
+	 * space).
+	 * 
+	 * Since we have to do the copy anyway, let's convert from interleaved to
+	 * sequential (i.e. all samples for the first channel laid-out contiguously,
+	 * followed by all from the second, and so forth). */
+	typedef IntegerToFloatSampleConvert<SampleFormat::S8> s8_cvt_t;
+	typedef IntegerToFloatSampleConvert<SampleFormat::S16> s16_cvt_t;
+	typedef IntegerToFloatSampleConvert<SampleFormat::S32> s32_cvt_t;
+	typedef IntegerToFloatSampleConvert<SampleFormat::S24_P32> s24_p32_t;
```
MaxKellermann, line 307: Naming convention for types in MPD is CamelCase, not foo_bar_t. Also please use `using` instead of the deprecated `typedef` keyword.

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+
+	for (size_t i = 0; i < num_samples; ++i) {
+		for (size_t j = 0; j < num_channels; ++j) {
+			/* `buf` index: i * num_channels + j
+			 * `in` index: j * num_samples + i */
+			float x;
+			switch (audio_format.format) {
+			case SampleFormat::S8:
+				x = s8_cvt_t::Convert(*(char*)buf[i * num_channels + j]);
```
MaxKellermann, line 316: `char` is the wrong type here.

src/output/plugins/visualization/SoundAnalysis.cxx
```diff
@@ -0,0 +1,394 @@
...
+			}
+			in.get()[j * num_samples + i] = x;
+		}
+	}
+
+	fftwf_execute(plan);
+
+	size_t max_coeffs_idx = num_samples/2;
+
+	for (uint8_t c = 0; c < num_channels; ++c) {
```
MaxKellermann, line 344: Using `uint8_t` here instead of `unsigned` has no effect at best, and can add overhead and code bloat. It's a harmful micro-optimization.

src/output/plugins/visualization/SoundAnalysis.hxx
```diff
@@ -0,0 +1,295 @@
...
+	/* Data greater than this frequency (in the frequency domain) shall be
+	 * discarded; must be greater than or equal to zero, and greater than
+	 * lo_cutoff. A typical value would be 10000-12000 (empirically, there's not a
+	 * lot of activity above 10000 in song data). Units Hz. Configuration value
+	 * "hi_cutoff" */
+	float hi_cutoff;
+
+	static constexpr size_t DEF_NUM_SAMPLES = 513;
+	static constexpr size_t DEF_LO_CUTOFF = 200;
+	static constexpr size_t DEF_HI_CUTOFF = 10000;
```
MaxKellermann, line 56: What does `DEF_` mean and why is this a `size_t` when the non-static fields with similar names are `float`?

src/output/plugins/visualization/SoundAnalysis.hxx
```diff
@@ -0,0 +1,295 @@
...
+
+public:
+	SoundAnalysisParameters() noexcept;
+	SoundAnalysisParameters(const ConfigBlock &config_block);
```
MaxKellermann, line 60: Should be `explicit` to avoid implicit unwanted conversions.

src/output/plugins/visualization/SoundAnalysis.hxx
```diff
@@ -0,0 +1,295 @@
...
+		// The # of frequencies stored in `out` per channel
+		size_t total_freq_per_chan = num_samples / 2 + 1;
+
+		// The maximum indexable frequency per channel
+		size_t upper_freq_per_chan = std::min(idx_hi, total_freq_per_chan);
+
+		/* In both `out` & `pout`, the coefficients are laid out as:
+		 * 	   | coeffs for chan #0... | coeffs for chan #1... | ... |
+		 * so outer loop will be on channel. */
+		for (uint8_t chan = 0; chan < num_channels; ++chan) {
```
MaxKellermann, line 262: Another bad micro-optimization by using `uint8_t`

src/output/plugins/visualization/SoundInfoCache.cxx
```diff
@@ -0,0 +1,231 @@
...
+
+inline
+typename std::chrono::microseconds::rep
+NowTicks() {
+	return duration_cast<SoundInfoCache::Time>(
+		std::chrono::system_clock::now().time_since_epoch()).count();
+}
+
+Visualization::SoundInfoCache::SoundInfoCache(const AudioFormat &audio_format,
+											  const Duration &buf_span):
```
MaxKellermann, line 25: Bad indent

MaxKellermann (changes requested):

src/output/plugins/visualization/LowLevelProtocol.hxx
sp1ff: Huh... I had no idea (I only work on Linux). Replaced... but I don't see how to use `PackedBE{16,32}` to copy the raw bytes into a buffer without recourse to `std::copy` (or its equivalent)?

src/output/plugins/visualization/Protocol.hxx
sp1ff: Added references to the message documentation.

src/output/plugins/visualization/SoundAnalysis.cxx
sp1ff: That would be a question for the authors of libfftw :) I'm just conforming to the API.

src/output/plugins/visualization/SoundAnalysis.cxx
sp1ff: Ditto-- libfftw API.

src/output/plugins/visualization/SoundAnalysis.cxx
sp1ff: libfftw API

src/output/plugins/visualization/SoundAnalysis.cxx
sp1ff: fixed.

src/output/plugins/visualization/SoundAnalysis.cxx
sp1ff: Yup, fixed.

src/output/plugins/visualization/SoundAnalysis.cxx
sp1ff: fixed

src/output/plugins/visualization/SoundAnalysis.cxx
sp1ff: Wasn't intended as an optimization or any sort; just chose the type to match that of `num_channels`. Changed, but can you say more about how making `c` a `uint8_t` could add overhead & code bloat?

src/output/plugins/visualization/SoundAnalysis.hxx
sp1ff: Short for "default"-- re-named. They're integral because `ConfigBlock` offers no accessors for floating-point values.

src/output/plugins/visualization/LowLevelProtocol.hxx
sp1ff: How so?

src/output/plugins/meson.build
sp1ff: Good point.

src/output/plugins/visualization/SoundInfoCache.hxx
sp1ff: Changed.

src/output/plugins/visualization/SoundInfoCache.hxx
sp1ff: Ahhhhh... I completely missed that. Thank you. Fixed.

src/output/plugins/visualization/Protocol.cxx
sp1ff: Sorry-- could you say more? If I replaced the `uint16_t` members of `struct ClientHello` with `PackedBE16` equivalents, that would help how, exactly. I mean, I see that the calls to `ntohs` would go away, to be replaced by invocations of their `uint16_t()` methods, I suppose?

src/output/plugins/meson.build
MaxKellermann: Now we have two Meson options: one to enable "FFTW support" - but what does it mean, what does it do?
Second option is "Visualization output plugin", I understand that. So I disable it, but MPD still looks for libfftw - why?
The I enable "visualization", but this still depends on the "fftw" options where I can have contradicting values.
The simples solution would be to have one single option, let's say "visualization", which is used as "required" parameter for looking up libfftw. This would be less obscure and there wouldn't be a way to have contradicting options.

src/output/plugins/visualization/LowLevelProtocol.hxx
MaxKellermann: Since the `Packed` classes are just bytes, you can simply cast a pointer to `PackedBE16*` and access the object from there.

src/output/plugins/visualization/LowLevelProtocol.hxx
MaxKellermann: Casting one object type to another is undefined behavior in C++ (there are a few exceptions).

src/output/plugins/visualization/Protocol.cxx
MaxKellermann: What methods are invoked?

src/output/plugins/visualization/SoundAnalysis.cxx
MaxKellermann: wtf.

src/lib/fmt/ThreadIdFormatter.hxx
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: BSD-2-Clause
+// author: Max Kellermann <max.kellermann@gmail.com>
+
+#ifndef THREAD_ID_FORMATTER_HXX
+#define THREAD_ID_FORMATTER_HXX
+
+#include <fmt/format.h>
+#include <sstream>
```
MaxKellermann, line 8: Please please please no iostreams in MPD!
Why do we even need this whole header??

src/output/plugins/meson.build
```diff
@@ -160,6 +160,28 @@ else
   wasapi_dep = dependency('', required: false)
 endif
 
+libfftw3_dep = dependency('fftw3f', version: '>= 3.3.8', required: get_option('fftw3'))
+output_features.set('ENABLE_FFTW3', libfftw3_dep.found())
```
MaxKellermann, line 164: This macro is never used

src/output/plugins/visualization/LowLevelProtocol.hxx
```diff
@@ -0,0 +1,57 @@
...
+#include <limits>
+
+namespace Visualization {
+
+/*  Write a uint16_t to an output iterator over byte in wire format; return the
+ *  iterator in its new position
+ */
+template <typename OutIter>
+OutIter
+SerializeU16(uint16_t n, OutIter pout) {
```
MaxKellermann, line 22: What's the reason you decided to make this a template instead of just passing a `std::byte*`?

src/output/plugins/visualization/LowLevelProtocol.hxx
```diff
@@ -0,0 +1,57 @@
...
+	auto m = PackedBE16(n);
+	auto p = (std::byte*)(&m);
+	return std::copy(p, p + 2, pout);
```
MaxKellermann, line 25: Using `std::copy` after creating a copy on the stack seems pretty clumsy. Now we have three copies of the value. Can we do better than that?

src/output/plugins/visualization/LowLevelProtocol.hxx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/* Convert an fftwf_complex to wire format; write it to an output iterator &
+ * return the iterator in its new position
+ */
+template <typename OutIter>
+OutIter
+SerializeComplex(const fftwf_complex c, OutIter pout) {
+	auto r = PackedBE32(*(const uint32_t*)&(c[0]));
+	auto i = PackedBE32(*(const uint32_t*)&(c[1]));
```
MaxKellermann, line 48: Yet more undefined behavior here. Why are you not calling `SerializeFloat` here twice? Too simple?

src/output/plugins/visualization/Protocol.cxx
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The Music Player Daemon Project
+
+#include "Protocol.hxx"
+
+#include "Log.hxx"
```
MaxKellermann, line 6: Why?

src/output/plugins/visualization/Protocol.cxx
```diff
@@ -0,0 +1,46 @@
...
+			   size_t length,
+			   ClientHello &clihlo) noexcept {
+	// CLIHLO payload is 6 bytes, header & footer are five more.
+	if (length < sizeof(ClientHello) + 5) {
+		return ParseResult::NEED_MORE_DATA;
+	}
+
+	uint8_t *buf = (uint8_t *)data;
+
+	uint16_t msg_type = FromBE16(*(uint16_t *)buf);
```
MaxKellermann, line 21: Cast is undefined behavior.

src/output/plugins/visualization/Protocol.hxx
```diff
@@ -0,0 +1,138 @@
...
+ */
+
+struct ClientHello {
+	/// Major protocol version the client would like to speak
+	uint8_t major_version;
+	/// Minor protocol version the client would like to speak
+	uint8_t minor_version;
+	/// The number of sound analyses per second the client would like to
+	/// receive (presumably the rate at which it is rendering frames, hence
+	/// the name "fps")
```
MaxKellermann, line 38: Don't use single-line comment syntax for multi-line comments

src/output/plugins/visualization/Protocol.hxx
```diff
@@ -0,0 +1,138 @@
...
+SerializeSrvhlo(std::byte major_ver, std::byte minor_ver, OutIter pout) {
+	using std::byte;
+
+	*pout++ = byte{0}; //
+	*pout++ = byte{1}; // message type
+	*pout++ = byte{0}; //
+	*pout++ = byte{2}; // payload length
+	*pout++ = major_ver;
+	*pout++ = minor_ver;
+	*pout++ = byte{0}; // check byte
```
MaxKellermann, line 90: Can we have a `struct` for this?

src/output/plugins/visualization/SoundAnalysis.hxx
```diff
@@ -0,0 +1,317 @@
...
+	size_t out_samples;
+	std::shared_ptr<SoundInfoCache> pcache;
+	AudioFormat audio_format;
+	/* # of audio channels (e.g. 1 is mono, 2 is stereo--
+	 * # cf. SampleFormat.hxx); should be audio_format.num_channels */
+	uint8_t num_channels;
+	/// Size of `buf`, in bytes
+	size_t cbbuf;
+	/// Pre-allocated buffer for raw PCM data
+	std::unique_ptr<std::byte[]> buf;
```
MaxKellermann, line 109: What happened to my suggestion to replace this with a single `AllocatedArray` field?

src/output/plugins/visualization/SoundAnalysis.hxx
```diff
@@ -0,0 +1,317 @@
...
+	/* Data greater than this frequency (in the frequency domain) shall be
+	 * discarded; must be greater than or equal to zero, and greater than
+	 * lo_cutoff. A typical value would be 10000-12000 (empirically, there's
+	 * not a lot of activity above 10000 in song data). Units
+	 * Hz. Configuration value "hi_cutoff" */
+	float hi_cutoff;
+
+	static constexpr size_t DEFAULT_NUM_SAMPLES = 513;
+	static constexpr size_t DEFAULT_LO_CUTOFF = 200;
+	static constexpr size_t DEFAULT_HI_CUTOFF = 10000;
```
MaxKellermann, line 56: What are these? They're never used.

src/output/plugins/visualization/SoundAnalysis.hxx
```diff
@@ -0,0 +1,317 @@
...
+ * There are any number of parameters governing our analysis of PCM data. Other
+ * than the `AudioFormat`, they are read from configuration at startup and
+ * constant. Rather than force callers to write methods taking many parameters,
+ * this class colects them all in one place, and enforces some constraints on
+ * their values.
+ *
+ *
+ */
+
+class SoundAnalysisParameters
```
MaxKellermann, line 36: Why is this a `class` with private fields and not a `struct`? Your design requires getter methods for something that simple, but what's the point of protecting the fields, for having to add code bloat?

src/output/plugins/visualization/SoundAnalysis.hxx
```diff
@@ -0,0 +1,317 @@
...
+	/// # of samples to be used in each analysis
+	size_t num_samples;
+	/* # of Fourier coefficients computed by FFTW (should be
+	 * (num_samples / 2) + 1) */
+	size_t out_samples;
+	std::shared_ptr<SoundInfoCache> pcache;
+	AudioFormat audio_format;
+	/* # of audio channels (e.g. 1 is mono, 2 is stereo--
+	 * # cf. SampleFormat.hxx); should be audio_format.num_channels */
+	uint8_t num_channels;
```
MaxKellermann, line 105: Are you aware that this `uint8_t` field in fact allocates 8 bytes in this struct?

src/output/plugins/visualization/SoundAnalysis.hxx
```diff
@@ -0,0 +1,317 @@
...
+	/* Bass/mids/trebs, laid-out as [bass, mids, trebs, bass, mids, trebs]
+	 * (i.e. 3 * num_channels floats) */
+	std::unique_ptr<float[]> bass_mids_trebs;
+
+public:
+	SoundAnalysis(const SoundAnalysisParameters &params,
+		      std::shared_ptr<SoundInfoCache> pc);
+
+	uint8_t
+	NumChan() const noexcept {
```
MaxKellermann, line 134: `NumChan` is a very obscure method name. What does it do? Does it numerate Jackie Chan?
Hey, this public method name is shorter than the private field it accesses!
If you really need this method, why not `GetChannelCount()`?

src/output/plugins/visualization/SoundAnalysis.hxx
```diff
@@ -0,0 +1,317 @@
...
+#include "SoundInfoCache.hxx"
+#include "LowLevelProtocol.hxx"
+
+#include <fftw3.h>
+
+#include <algorithm>
+#include <cstddef>
+#include <memory>
+
+#include <math.h>
```
MaxKellermann, line 16: Why?

src/output/plugins/visualization/SoundAnalysis.hxx
```diff
@@ -0,0 +1,317 @@
...
+		pout = SerializePowerSpectra(pout);
+		pout = SerializeBassMidsTrebs(pout);
+		return pout;
+	}
+
+	/* Write the Fourier coefficients in the range `[idx_lo, idx_hi)` to
+	 * \a pout first transforming them by \a op. */
+	template <typename OutIter>
+	OutIter
+	TransformCoeffs(
```
MaxKellermann, line 249: This looks like a huge method. Why is it `inline`?

src/output/plugins/visualization/SoundInfoCache.hxx
```diff
@@ -0,0 +1,104 @@
...
+	unsigned frame_size;
+	/* Mutex guarding the ring buffer since instances will be accessed from
+	   multiple threads */
+	mutable Mutex mutex;
+	/// this is the ring buffer
+        AllocatedArray<uint8_t> ring;
+	/// # of bytes currently in the ring buffer (as distinct from capacity)
+	std::size_t cb;
+	/// Valid PCM data exists in buf[p0, p1)
+	size_t p0, p1;
```
MaxKellermann, line 71: `std::size_t`?

src/output/plugins/visualization/SoundInfoCache.hxx
```diff
@@ -0,0 +1,104 @@
...
+#include "thread/Mutex.hxx"
+#include "util/AllocatedArray.hxx"
+
+#include <chrono>
+#include <memory>
+
+namespace Visualization {
+
+/**
+ * \brief Thread-safe cache for recent PCM data
```
MaxKellermann, line 18: I don't understand the whole point of this class and why it needs to be thread-safe. I also don't understand the documentation text here.
Why "cap the amount of of memory used"? What does that mean?
Why is it called "cache"?

src/output/plugins/visualization/VisualizationClient.hxx
```diff
@@ -0,0 +1,274 @@
...
+	};
+	/// Plugin open/closed state-- cf. PluginIsOpen()
+	std::variant<std::monostate, HavePcmData> pcm_state;
+
+	/// The protocol can be represented as an FSM
+	enum class ProtocolState {
+		/* FSM initial state; the socket has been established, but no
+		 * communication has taken place; we are expecting a CLIHLO
+		 * message to arrive (i.e. a READ/POLLIN notification) */
+		Init,
```
MaxKellermann, line 164: enum choices should be UPPER_CASE

src/output/plugins/visualization/VisualizationClient.cxx
```diff
@@ -0,0 +1,533 @@
...
+
+#include <chrono>
+#include <thread>
+#include <variant>
+
+const Domain d_vis_client("vis_client");
+
+inline
+typename std::chrono::microseconds::rep
+NowTicks() {
```
MaxKellermann, line 19: This function is only used for your excessive debug prints, but those should be removed before I merge this. What's even the point of this function, why do you insert those time stamps in all log messages?

src/output/plugins/visualization/VisualizationServer.cxx
```diff
@@ -0,0 +1,106 @@
...
+		if (p->IsClosed()) {
+			LogInfo(vis_server_domain, "Reaping closed client.");
+			clients.erase(p);
+		}
+	}
+
+	if (!clients.empty()) {
+		LogInfo(vis_server_domain, "Scheduling another reaping in 3 "
+			"seconds.");
+		reaper.Schedule(std::chrono::seconds(3));
```
MaxKellermann, line 50: This periodically calls ReapClients, all the time, while MPD runs! This is bad, because it causes unnecessary CPU wakeups. Never wake up the CPU unless you explicitly have a reason to do so.
I also don't understand the point of this whole method. Why does it exist? If a client is closed, why not delete it immediately?

src/output/plugins/visualization/VisualizationServer.cxx
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+void
+Visualization::VisualizationServer::OnPluginOpened(
+	const std::shared_ptr<SoundInfoCache> &pcache)
+{
+	state = HavePcmData{pcache };
+
+	for (auto p0 = clients.begin(), p1 = clients.end(); p0 != p1; ) {
```
MaxKellermann, line 60: This iteration happens in the OutputThread, but other methods happen in the i/o thread (via EventLoop), thus all methods in this class may crash all the time.

src/output/plugins/visualization/VisualizationServer.hxx
```diff
@@ -0,0 +1,119 @@
...
+ *
+ *
+ */
+
+class VisualizationServer : public ServerSocket {
+
+	/// only valid when the plugin is open
+	struct HavePcmData {
+		// I wish C++ had a `not_null` class
+		std::shared_ptr<SoundInfoCache> pcache;
```
MaxKellermann, line 79: Why is this a shared_ptr? What's the point of adding reference counting overhead here?

MaxKellermann (changes requested):
I read only part of the code, but found so many things that were wrong or more complicated than they should be, I stopped reading further.

sp1ff:
@MaxKellermann This isn't going well. Before I embark on another round of low-level changes (addressing "excessive logging", for instance), I think it might be beneficial to first verify that that basic architecture of the plugin is acceptable (not much use updating code that's going to need to be re-written, after all). I've documented that at some length [here](https://github.com/sp1ff/MPD/blob/soundinfo/src/output/plugins/visualization/VisualizationOutputPlugin.cxx#L391), where I hope you'll find the answers to some of your questions above.

In the interests of time, if you _do_ want changes, it would help me if you could indicate what you'd like to see _instead_ of what I've written. I realize you've done that in some cases, but the review so far contains many open-ended questions (such as "why?"). When reviewing others' code, I've gotten better results with concrete suggestions.

Finally, and in particular, could you tell me a bit about casting being undefined behavior? When, e.g. I cast a sixteen bit value to a `uint16_t` I expect to get the bit pattern. This has worked for me on multiple platforms, in both kernel & user mode, for nearly three decades now; that said, I'm always interesting in honing my craft. Could you say more?

MaxKellermann:
> contains many open-ended questions (such as "why?"). When reviewing others' code, I've gotten better results with concrete suggestions.

When I see weird code, I want to know why you wrote it that way. I assume you have good reasons which are just outside of my grasp. So I ask "why".

Maybe you don't have good reasons, but only after I find that out I can make suggestions.

> could you tell me a bit about casting being undefined behavior?

You mustn't cast an unaligned pointer to a pointer to a 16 bit integer (and then dereference it). This works on most CPUs, but may crash on others (with SIGBUS) - but for certain the C++ language does not allow it. Maybe you got away with that mistake for three decades, but that doesn't mean it's correct and doesn't mean it will never crash for anybody. Let's write the code to be correct, maybe more correct than it is necessary for the CPUs you happen to have been using for the last three decades, hoping that MPD will remain stable for everybody.

MaxKellermann:
Is this going anywhere?

MaxKellermann:
> Is this going anywhere?

Apparently not.

PR #1913, opened 2023-12-13
android: update Android build tools, fix android builds on macOS

DDRBoxman:


MaxKellermann:
The one commit in your PR does many unrelated things, but only one is described by the commit message, none are explained.

DDRBoxman:
Cool, I'll split this up better

PR #1915, opened 2023-12-13
android: Fixes for building on macOS

DDRBoxman:


android/build.py
```diff
@@ -60,8 +61,11 @@
     '-Dandroid_ndk=' + ndk_path,
     '-Dandroid_abi=' + android_abi,
     '-Dandroid_strip=' + toolchain.strip,
+    '-Dopenssl:asm=disabled'
```
DDRBoxman, line 64: This argument is usually needed on m1 based macs since they don't have `asm`.
I wasn't sure if there was a good way to disable building openssl at all with the meson wrap system so we could just drop this.

doc/user.rst
```diff
@@ -197,7 +197,7 @@ Compiling for Android
 You need:
 
 * Android SDK (sdk platform 29, build tools 29.0.3)
-* `Android NDK r25b <https://developer.android.com/ndk/downloads>`_
+* `Android NDK r26b <https://developer.android.com/ndk/downloads>`_
```
DDRBoxman, line 200: Turns out I had used this newer ndk on my recent linux build too

MaxKellermann:
Thanks!
btw. don't document your work in GH comments; those get lost after merging. Only git commit messages (and of course code comments) remain.

PR #1923, opened 2023-12-18
android: migrate app build system to use gradle

DDRBoxman:
Most of the Android specific meson code has been removed and replaced with the grade build system.

The new meson build scripts build and move the libmpd.so binaries into the correct location that gradle expects. After than gradle handles building the rest of the Android app.

Icons and banners have been updated for the modern app packaging expectations.

For reference here was the figma template Google provides that I used to back the png versions for older versions of Android <https://www.figma.com/community/file/1283953738855070149>


MaxKellermann:
Hey, while you're at this, can you add a GH actions job to build the Android app?

DDRBoxman:
Can do!

PR #1925, opened 2023-12-19
android: Add github action

DDRBoxman:
This lives under a separate workflow so it can have different ignore paths.

### DIFF
--- a/.github/workflows/build_android.yml
+++ b/.github/workflows/build_android.yml
@@ -1,0 +1,79 @@
+---
+on:
+  workflow_dispatch:
+  push:
+    paths-ignore:
+      - 'build/**'
+      - 'doc/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/build_android.yml
```diff
@@ -0,0 +1,79 @@
...
+          distribution: 'zulu'
+          java-version: 17
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends \
+            ninja-build \
+            quilt 
+          pip3 install --user meson==0.56.2
```
DDRBoxman, line 54: Meson needed a bump here to work around some include path stuff that llvm was throwing.

https://reviews.llvm.org/D131441

PR #1928, opened 2023-12-20
android: poc control socket

DDRBoxman:
Just a quick PoC showing that we can use unix sockets on android to communicate to MPD from java. This would allow us to reliably have a low overhead way to bind a java MPD client from the Android service.

I figured I would throw this up as a draft PR to get some feedback on the idea as I wasn't sure if you wanted to expose more native bindings over the bridge instead.

### DIFF
--- a/android/app/src/main/java/org/musicpd/Main.java
+++ b/android/app/src/main/java/org/musicpd/Main.java
@@ -15,6 +15,8 @@ import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.ServiceConnection;
 import android.media.AudioManager;
+import android.net.LocalSocket;
+import android.net.LocalSocketAddress;
 import android.os.Build;
 import android.os.IBinder;
 import android.os.PowerManager;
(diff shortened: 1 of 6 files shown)


MaxKellermann:
But what's the point? Do you want to implement a MPD client inside the MPD server app? The whole point of MPD is to separate server from client.

DDRBoxman:
The main goal here is to get the song metadata into the android media session which allows the current song state to show up on things like the lock screen, or a speaker or car connected over bluetooth. Ideally these devices could use their controls to send commands like play, pause, etc. to MPD. 

Plumbing this through the socket seems like a good solution since it's the public API that MPD provides and requires way less work to plumb that information up to the host app.

The other option is just finding ways to grab that information out of the native code and find ways to hook into state changes there and pass it through the native bridge.

MaxKellermann:
I think the simplest and most robust solution is to just invoke a Java method whenever a state change in the Java world is necessary.
We may provide a local socket if this seems useful for local MPD client apps to connect, but for connections from within the MPD process, between different threads, a MPD socket connection seems overkill; it would add complexities and chances for things to go wrong.

DDRBoxman:
👍

PR #1933, opened 2023-12-23
android: Redo settings with Compose and Material UI

DDRBoxman:
This changes lets us use the latest UI design from Google, Material 3.

Google only provides the material UI 3 themes for compose, compose only works with Kotlin.

We end up with a lot of nice improvements where we don't have to worry about the lifecycles of things as closely as before which is nice. And a lot of the new Flow stuff is thread safe.

Screenshot_20231222_224723
Screenshot_20231222_224745


DDRBoxman:
Also now that the different UI sections are in compose it's really to move them into different parts of the UI

PR #1936, opened 2023-12-23
Set minimum meson_version to 0.63.0

jcorporation:
Fixes following warning:
WARNING: Project specifies a minimum meson_version '>= 0.56.0' but uses features which were added in newer versions:
 * 0.63.0: {'Wrap files with diff_files'}

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -97,7 +97,7 @@ jobs:
           directory: output/full
           setup-options: -Ddocumentation=disabled -Dtest=true -Dsystemd=enabled -Dpcre=enabled
           options: --verbose
-          meson-version: 0.56.0
+          meson-version: 0.63.0
 
       - name: Unit Tests
         uses: BSFishy/meson-build@v1.0.3
@@ -106,7 +106,7 @@ jobs:
           directory: output/full
           setup-options: -Ddocumentation=disabled -Dtest=true -Dsystemd=enabled -Dpcre=enabled
           options: --verbose
-          meson-version: 0.56.0
+          meson-version: 0.63.0
 
       - name: Mini Build
         uses: BSFishy/meson-build@v1.0.3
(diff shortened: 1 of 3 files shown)


naglis:
Looks like the build workflow [needs to be updated](https://github.com/MusicPlayerDaemon/MPD/actions/runs/7308878981/job/19915917602?pr=1936#step:5:50):
- https://github.com/MusicPlayerDaemon/MPD/blob/23b4688c444f3a083c479946dc0cca6d1076fc65/.github/workflows/build.yml#L100
- https://github.com/MusicPlayerDaemon/MPD/blob/23b4688c444f3a083c479946dc0cca6d1076fc65/.github/workflows/build.yml#L109
- https://github.com/MusicPlayerDaemon/MPD/blob/23b4688c444f3a083c479946dc0cca6d1076fc65/.github/workflows/build.yml#L118
- https://github.com/MusicPlayerDaemon/MPD/blob/23b4688c444f3a083c479946dc0cca6d1076fc65/.github/workflows/build.yml#L164
- https://github.com/MusicPlayerDaemon/MPD/blob/23b4688c444f3a083c479946dc0cca6d1076fc65/.github/workflows/build.yml#L173

The Meson version requirement is also mentioned in the documentation, so would be good to update it as well.
- https://github.com/MusicPlayerDaemon/MPD/blob/23b4688c444f3a083c479946dc0cca6d1076fc65/doc/user.rst?plain=1#L61
- https://github.com/MusicPlayerDaemon/MPD/blob/23b4688c444f3a083c479946dc0cca6d1076fc65/doc/user.rst?plain=1#L161
- https://github.com/MusicPlayerDaemon/MPD/blob/23b4688c444f3a083c479946dc0cca6d1076fc65/doc/user.rst?plain=1#L201

MaxKellermann:
I'm not sure if we should do this - it means it's not any longer possible to build MPD with Ubuntu LTS (Meson 0.61), doesn't it?
I mean, it's possible to build MPD without any of the Meson wraps, and Meson 0.63 is only necessary to build those wraps. We shouldn't do this just to shut up a Meson warning if it means we break the build for people who don't need that feature.

jcorporation:
I didn’t take this into account.

PR #1941, opened 2023-12-25
meson.build: raise required meson version to 0.63

skidoo23:
Fixes WARNING: Project specifies a minimum meson_version '>= 0.56.0' but uses features which were added in newer versions: 0.63.0: {'Wrap files with diff_files'}


skidoo23:
Already fixed by jcorporation.

PR #1944, opened 2023-12-26
no setenv on Windows

neheb:


src/output/plugins/PulseOutputPlugin.cxx
```diff
@@ -23,6 +23,10 @@
 
 #include <stdlib.h>
 
+#ifdef _WIN32
+#include <windows.h>
```
MaxKellermann, line 27: According to MSDN, the function is provided by `processenv.h`; we should always use the narrowest possible header.

neheb: OTOH, upnp support is broken because of that.

neheb: Fixed.

PR #1945, opened 2023-12-27
android: Set kotlin jvm target to 1.9

DDRBoxman:


MaxKellermann:
I wish your commit messages were slightly more verbose - what you're doing is not my expertise, and it would be useful for me (and others) to know why you are doing this - what is the purpose of this change, what does it enable or improve, what future stuff does it prepare? This PR's commit message is mostly useless - it only describes what is obvious from the diff, but doesn't explain anything. Okay for now, but please keep in mind to always write a tiny bit about the "why".

PR #1947, opened 2023-12-27
mostly libupnp fixes

neheb:


src/lib/xiph/meson.build
```diff
@@ -1,6 +1,6 @@
 libflac_dep = dependency('flac', version: '>= 1.2', required: get_option('flac'))
 
-if is_windows
+if is_windows and libflac_dep.found()
   # Our Windows build generates a static libFLAC build
   libflac_dep = declare_dependency(compile_args: '-DFLAC__NO_DLL',
```
neheb, line 5: Another way to fix is to make libflac_dep a disabler. OTOH, this is simpler.

src/db/plugins/upnp/ContentDirectoryService.cxx
```diff
@@ -77,8 +77,8 @@ ContentDirectoryService::readDirSlice(UpnpClient_Handle hdl,
 		{"BrowseFlag", "BrowseDirectChildren"},
 		{"Filter", "*"},
 		{"SortCriteria", ""},
-		{"StartingIndex", fmt::format_int{offset}.c_str(),
-		{"RequestedCount", fmt::format_int{count).c_str()},
+		{"StartingIndex", fmt::format_int(offset).c_str()},
```
neheb, line 80: actually why is fmt even used here and not std::to_string?

src/lib/xiph/meson.build
MaxKellermann: Disablers are nice, but not applicable here, because this would simply disable the whole MPD executable.

src/db/plugins/upnp/ContentDirectoryService.cxx
MaxKellermann: In this case, a `std::string` gets allocated anyway, so there's no advantage of using fmt's fixed-size stack buffer; but just always using fmt seems like a sane choice; std is inferior more often than not.

MaxKellermann:
I don't understand the patch "upnp: fix compilation with libnpupnp". Your commit message doesn't explain the problem.

MaxKellermann:
"MPD uses strict header includes under Windows, which blows up"
This is just the library's fault and there is no pressure for MPD to work around this. I havn't ever heard of a single user using this library on Windows, and there is no CI test case for it. If anybody wants this fixed, they should contact the library maintainer, not MPD.

neheb:
> I don't understand the patch "upnp: fix compilation with libnpupnp". Your commit message doesn't explain the problem.

Mismatching () and {}

neheb:
> "MPD uses strict header includes under Windows, which blows up" This is just the library's fault and there is no pressure for MPD to work around this. I havn't ever heard of a single user using this library on Windows, and there is no CI test case for it. If anybody wants this fixed, they should contact the library maintainer, not MPD.

It is. I reported to npupnp. It was fixed there. pupnp is left but it seems dead upstream.

src/db/plugins/upnp/ContentDirectoryService.cxx
neheb: I assume in the future MPD will migrate to std::format. Makes sense to reduce dependency on fmt.

neheb:
msys2 CI: https://github.com/neheb/MPD/tree/mm

note the stuff I disabled since it doesn't build (mostly wchar vs char issues).

MaxKellermann:
> msys2 CI: https://github.com/neheb/MPD/tree/mm

Uh, there are tons of errors from system headers. How is that possible? I don't think MPD should aim to support such a target.

neheb:
@MaxKellermann just looked at it. The only thing I see is the iphlapi.h include from lib(np)upnp causing issues because of MPD's lightweight header defines. Unless you mean something else.

neheb:
@MaxKellermann fixed. New commits added.

PR #1949, opened 2023-12-29
sndfile: fix compilation with clang's libc++

neheb:
error: non-constant-expression cannot be narrowed from type 'sf_count_t' (aka 'long long') to 'size_type' (aka 'unsigned int') in initializer list
 [-Wc++11-narrowing]
  215 |                      std::span{buffer, num_frames * frame_size},
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~
note: insert an explicit cast to silence this issue
  215 |                      std::span{buffer, num_frames * frame_size},
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~
      |                                        static_cast<size_type>()

### DIFF
--- a/src/decoder/plugins/SndfileDecoderPlugin.cxx
+++ b/src/decoder/plugins/SndfileDecoderPlugin.cxx
@@ -212,7 +212,7 @@ sndfile_stream_decode(DecoderClient &client, InputStream &is)
 			break;
 
 		cmd = client.SubmitAudio(is,
-					 std::span{buffer, num_frames * frame_size},
+					 std::span{buffer, static_cast<size_t>(num_frames) * frame_size},
 					 0);
 		if (cmd == DecoderCommand::SEEK) {
 			sf_count_t c = client.GetSeekFrame();


src/decoder/plugins/SndfileDecoderPlugin.cxx
```diff
@@ -212,7 +212,7 @@ sndfile_stream_decode(DecoderClient &client, InputStream &is)
 			break;
 
 		cmd = client.SubmitAudio(is,
-					 std::span{buffer, num_frames * frame_size},
+					 std::span{buffer, static_cast<size_t>(num_frames * frame_size)},
```
MaxKellermann, line 215: This is the wrong way to fix this. This should never have been a signed multiplication in the first place.

neheb: hmm? What's the proper way?

MaxKellermann: Your code does `num_frames*frame_size` first (a signed multiply), and the result is a signed integer (which might have overflowed, which is UB), and then you cast the undefined result to unsigned (i.e. `size_t`), giving you a random value.
This is the proper way:
```
std::span{buffer, static_cast<size_t>(num_frames) * frame_size},
```
This ensures that both operands are unsigned, then does an unsigned multiply, resulting in an unsigned integer (`size_t`).

neheb: Oh I thought you meant something else. Fixed.

PR #1957, opened 2024-01-06
android: Support Next and Previous track from media session

DDRBoxman:


DDRBoxman:
Tested with the Media Controller Test App which allows you to control apps with a media session.

https://github.com/googlesamples/android-media-controller

PR #1968, opened 2024-01-17
HTTPD output: Added support for HTTP Basic authentication

Mrkvak:
I want to have a publicly available streaming MPD server available from internet. However I want to protect the output with password - if not for anything else then for legal reasons (the moment someone other than me opens the stream I break the law).

This adds new optional "password" option for "httpd" output in mpd.conf. If it is specified and non-empty, it requires HTTP Basic authentication with arbitrary user and set password.

(It's been some time since I've wrote anything in C++, so please check the code if I haven't made any stupid mistakes. I've tested it and it seems to be working as intended in both cases.)

### DIFF
--- a/doc/mpdconf.example
+++ b/doc/mpdconf.example
@@ -283,6 +283,7 @@ input {
 #	bitrate		"128"			# do not define if quality is defined
 #	format		"44100:16:1"
 #	max_clients	"0"			# optional 0=no limit
+#	password		"secretPassword"	# optional, enables HTTP Basic authentication with set password
 #}
 #
 # An example of a pulseaudio output (streaming to a remote pulseaudio server)
--- a/doc/mpdconf.example
+++ b/doc/mpdconf.example
@@ -283,7 +283,7 @@ input {
 #	bitrate		"128"			# do not define if quality is defined
 #	format		"44100:16:1"
 #	max_clients	"0"			# optional 0=no limit
-#	password		"secretPassword"	# optional, enables HTTP Basic authentication with set password
+#	password		"secretPassword"	# optional, enables HTTP Basic authentication with set password (needs base64 support)
 #}
 #
 # An example of a pulseaudio output (streaming to a remote pulseaudio server)
--- a/src/output/plugins/httpd/HttpdClient.cxx
+++ b/src/output/plugins/httpd/HttpdClient.cxx
@@ -8,7 +8,6 @@
 #include "util/AllocatedString.hxx"
 #include "Page.hxx"
 #include "IcyMetaDataServer.hxx"
-#include "lib/crypto/Base64.hxx"
 #include "net/SocketError.hxx"
 #include "net/UniqueSocketDescriptor.hxx"
 #include "util/SpanCast.hxx"
@@ -24,7 +23,6 @@
 #ifdef HAVE_BASE64
 #include "lib/crypto/Base64.hxx"
 #endif
-
 
 HttpdClient::~HttpdClient() noexcept
 {
@@ -115,14 +113,13 @@ HttpdClient::HandleLine(const char *line) noexcept
 			return true;
 		}
 
-
-		#ifdef HAVE_BASE64
+#ifdef HAVE_BASE64
 		if (const char *b64 = StringAfterPrefixIgnoreCase(line, "Authorization: Basic ")) {
 			auto decodedBytes = DecodeBase64(b64);
 			auto [username, pw] = Split(ToStringView(decodedBytes), ':');
 			provided_password = std::string{pw};
 		}
-		#endif
+#endif
 
 		if (StringEqualsCaseASCII(line, "Icy-MetaData: 1", 15) ||
 		    StringEqualsCaseASCII(line, "Icy-MetaData:1", 14)) {
@@ -200,8 +197,7 @@ HttpdClient::HttpdClient(HttpdOutput &_httpd, UniqueSocketDescriptor _fd,
 			 bool _metadata_supported)
 	:BufferedSocket(_fd.Release(), _loop),
 	 httpd(_httpd),
-	 metadata_supported(_metadata_supported),
-	 provided_password{""}
+	 metadata_supported(_metadata_supported)
 {
 }
 
--- a/src/output/plugins/httpd/HttpdClient.cxx
+++ b/src/output/plugins/httpd/HttpdClient.cxx
@@ -11,7 +11,6 @@
 #include "net/SocketError.hxx"
 #include "net/UniqueSocketDescriptor.hxx"
 #include "util/SpanCast.hxx"
-#include "util/StringCompare.hxx"
 #include "util/StringSplit.hxx"
 #include "Log.hxx"
 
@@ -22,6 +21,7 @@
 
 #ifdef HAVE_BASE64
 #include "lib/crypto/Base64.hxx"
+#include "util/StringCompare.hxx"
 #endif
 
 HttpdClient::~HttpdClient() noexcept
@@ -48,9 +48,11 @@ HttpdClient::BeginResponse() noexcept
 {
 	assert(state != State::RESPONSE);
 
+#ifdef HAVE_BASE64
 	if (!head_method) {
 		should_reject_unauthorized = httpd.password != nullptr && provided_password != httpd.password;
 	}
+#endif
 
 	state = State::RESPONSE;
 	current_page = nullptr;
@@ -151,6 +153,7 @@ HttpdClient::SendResponse() noexcept
 			"Connection: close\r\n"
 			"\r\n"
 			"404 not found";
+#ifdef HAVE_BASE64
 	} else if (should_reject_unauthorized) {
 		allocated = fmt::format(
 			"HTTP/1.1 401 unauthorized\r\n"
@@ -160,6 +163,7 @@ HttpdClient::SendResponse() noexcept
 			"\r\n"
 			"401 unauthorized", httpd.name);
 		response = allocated.c_str();
+#endif
 	} else if (metadata_requested) {
 		allocated =
 			icy_server_metadata_header(httpd.name, httpd.genre,
@@ -442,7 +446,11 @@ HttpdClient::OnSocketInput(void *data, size_t length) noexcept
 		if (!SendResponse())
 			return InputResult::CLOSED;
 
+#ifdef HAVE_BASE64
 		if (head_method || should_reject || should_reject_unauthorized) {
+#else
+		if (head_method || should_reject) {
+#endif
 			LockClose();
 			return InputResult::CLOSED;
 		}
--- a/src/output/plugins/httpd/HttpdClient.cxx
+++ b/src/output/plugins/httpd/HttpdClient.cxx
@@ -1,21 +1,30 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 // Copyright The Music Player Daemon Project
 
+#include "config.h"
 #include "HttpdClient.hxx"
 #include "HttpdInternal.hxx"
 #include "util/ASCII.hxx"
 #include "util/AllocatedString.hxx"
 #include "Page.hxx"
 #include "IcyMetaDataServer.hxx"
+#include "lib/crypto/Base64.hxx"
 #include "net/SocketError.hxx"
 #include "net/UniqueSocketDescriptor.hxx"
 #include "util/SpanCast.hxx"
+#include "util/StringCompare.hxx"
+#include "util/StringSplit.hxx"
 #include "Log.hxx"
 
 #include <fmt/core.h>
 
 #include <cassert>
 #include <cstring>
+
+#ifdef HAVE_BASE64
+#include "lib/crypto/Base64.hxx"
+#endif
+
 
 HttpdClient::~HttpdClient() noexcept
 {
@@ -40,6 +49,10 @@ void
 HttpdClient::BeginResponse() noexcept
 {
 	assert(state != State::RESPONSE);
+
+	if (!head_method) {
+		should_reject_unauthorized = httpd.password != nullptr && provided_password != httpd.password;
+	}
 
 	state = State::RESPONSE;
 	current_page = nullptr;
@@ -102,6 +115,15 @@ HttpdClient::HandleLine(const char *line) noexcept
 			return true;
 		}
 
+
+		#ifdef HAVE_BASE64
+		if (const char *b64 = StringAfterPrefixIgnoreCase(line, "Authorization: Basic ")) {
+			auto decodedBytes = DecodeBase64(b64);
+			auto [username, pw] = Split(ToStringView(decodedBytes), ':');
+			provided_password = std::string{pw};
+		}
+		#endif
+
 		if (StringEqualsCaseASCII(line, "Icy-MetaData: 1", 15) ||
 		    StringEqualsCaseASCII(line, "Icy-MetaData:1", 14)) {
 			/* Send icy metadata */
@@ -132,6 +154,15 @@ HttpdClient::SendResponse() noexcept
 			"Connection: close\r\n"
 			"\r\n"
 			"404 not found";
+	} else if (should_reject_unauthorized) {
+		allocated = fmt::format(
+			"HTTP/1.1 401 unauthorized\r\n"
+			"Content-Type: text/plain\r\n"
+			"WWW-Authenticate: Basic realm=\"{}\"\r\n"
+			"Connection: close\r\n"
+			"\r\n"
+			"401 unauthorized", httpd.name);
+		response = allocated.c_str();
 	} else if (metadata_requested) {
 		allocated =
 			icy_server_metadata_header(httpd.name, httpd.genre,
@@ -169,7 +200,8 @@ HttpdClient::HttpdClient(HttpdOutput &_httpd, UniqueSocketDescriptor _fd,
 			 bool _metadata_supported)
 	:BufferedSocket(_fd.Release(), _loop),
 	 httpd(_httpd),
-	 metadata_supported(_metadata_supported)
+	 metadata_supported(_metadata_supported),
+	 provided_password{""}
 {
 }
 
@@ -414,7 +446,7 @@ HttpdClient::OnSocketInput(void *data, size_t length) noexcept
 		if (!SendResponse())
 			return InputResult::CLOSED;
 
-		if (head_method || should_reject) {
+		if (head_method || should_reject || should_reject_unauthorized) {
 			LockClose();
 			return InputResult::CLOSED;
 		}
--- a/src/output/plugins/httpd/HttpdClient.hxx
+++ b/src/output/plugins/httpd/HttpdClient.hxx
@@ -4,6 +4,7 @@
 #ifndef MPD_OUTPUT_HTTPD_CLIENT_HXX
 #define MPD_OUTPUT_HTTPD_CLIENT_HXX
 
+#include "config.h"
 #include "Page.hxx"
 #include "event/BufferedSocket.hxx"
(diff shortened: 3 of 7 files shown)


src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -7,6 +7,7 @@
 #include "util/AllocatedString.hxx"
 #include "Page.hxx"
 #include "IcyMetaDataServer.hxx"
+#include "lib/crypto/Base64.hxx"
```
MaxKellermann, line 10: This header is only available if the macro `HAVE_BASE64` is defined (don't forget to `#include "config.h"`), and then you need to add a Meson library dependency on `crypto_base64_dep`.

src/output/plugins/httpd/HttpdOutputPlugin.cxx
```diff
@@ -44,6 +44,8 @@ HttpdOutput::HttpdOutput(EventLoop &_loop, const ConfigBlock &block)
 			ServerSocket::SetDscpClass(value);
 		});
 
+	password = block.GetBlockValue("password", "");
```
MaxKellermann, line 47: Why not `nullptr` for "no password necessary"? That seems like a natural choice, and would make your code simpler.

src/output/plugins/httpd/HttpdClient.hxx
```diff
@@ -111,6 +117,11 @@ class HttpdClient final
 	 */
 	unsigned metadata_fill = 0;
 
+	/**
+	 * Provided password (using HTTP basic auth)
+	 */
+	std::string *provided_password;
```
MaxKellermann, line 123: A pointer to a `std::string`? That makes my head hurt.
One way to make the code simpler and safer would be assuming an empty string means no password was provided; but if you really want to know the difference between "no password" and "empty password", then `std::optional<std::string>` would be better. That would be safer *and* simpler. Though I think it would be somewhat overengineered. Empty string is fine for me.

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -102,6 +112,15 @@ HttpdClient::HandleLine(const char *line) noexcept
 			return true;
 		}
 
+		if (StringEqualsCaseASCII(line, "Authorization: Basic ", 21)) {
+			auto encoded_pwd_length = strlen(line) - 21;
+			size_t debase64_size = CalculateBase64OutputSize(encoded_pwd_length);
+			auto debase64_buffer = std::make_unique<char[]>(debase64_size);
+			debase64_size = DecodeBase64({debase64_buffer.get(), debase64_size}, line + 21);
+			std::string auth_str = std::string(debase64_buffer.get(), debase64_size);
+			provided_password = new std::string(auth_str, auth_str.find(':')+1);
```
MaxKellermann, line 121: This crashes if the `auth_str` contains no colon.

Hint:
```
  const auto [username, pw] = Split(auth_str, ':');
  provided_password = std::string{pw};
```
Simpler & safer.

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -102,6 +112,15 @@ HttpdClient::HandleLine(const char *line) noexcept
 			return true;
 		}
 
+		if (StringEqualsCaseASCII(line, "Authorization: Basic ", 21)) {
+			auto encoded_pwd_length = strlen(line) - 21;
+			size_t debase64_size = CalculateBase64OutputSize(encoded_pwd_length);
```
MaxKellermann, line 117: This is redundant code - there's a `DecodeBase64()` overload which returns `AllocatedArray`. You can then cast this buffer to `std::string_view` (e.g. via `ToStringView()` from `util/SpanCast.hxx`) and operate on this buffer.

This is not only simpler and doesn't add redundant code, but is also faster because it reduces the two `strlen()` calls in your code to just one.

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -102,6 +112,15 @@ HttpdClient::HandleLine(const char *line) noexcept
...
+			auto debase64_buffer = std::make_unique<char[]>(debase64_size);
+			debase64_size = DecodeBase64({debase64_buffer.get(), debase64_size}, line + 21);
+			std::string auth_str = std::string(debase64_buffer.get(), debase64_size);
```
MaxKellermann, line 120: Why copy the string into a newly allocated `std::string` here? Operating on a `std::string_view` would eliminate this heap allocation + copy.

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -41,6 +47,10 @@ HttpdClient::BeginResponse() noexcept
 {
 	assert(state != State::RESPONSE);
 
+	if (!head_method) {
+	       should_reject_unauthorized = (strlen(httpd.password) > 0 && (!provided_password || (strcmp(provided_password->c_str(), httpd.password) != 0)));
```
MaxKellermann, line 51: The `strlen()` call is clumsy and adds overhead, because it needs to walk the whole string until the end for no reason. All you want to know is whether the string is empty, and `*http.password != 0` tells you that without walking the string. MPD has a convenience function for that to make it easier to understand: `!StringIsEmpty(http.password)` (does the same, but gives it a name which declares your intention).
But if you change it according to my other hints, you can easily write instead:
```
should_reject_unauthorized = http.password != nullptr && provided_password != http.password;
```
Simple, eh?

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -102,6 +112,15 @@ HttpdClient::HandleLine(const char *line) noexcept
 			return true;
 		}
 
+		if (StringEqualsCaseASCII(line, "Authorization: Basic ", 21)) {
```
MaxKellermann, line 115: Where does the `21` come from? This coding style is fragile, because who really verifies that this string literal has 21 characters?
Try this:
```
if (const char *b64 = StringAfterPrefixIgnoreCase(line, "Authorization: Basic ")) {
```
This lets the compiler figure out the size of the string literal, no need to do it manually, nothing fragile here.

MaxKellermann (changes requested):

src/output/plugins/httpd/HttpdClient.cxx
naglis, line 51: It is [advised](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#compare-password-hashes-using-safe-functions) to use constant-time string comparison for such cases to avoid [timing attacks](https://en.wikipedia.org/wiki/Timing_attack). Has this been considered?

kingosticks:
I'm sorry to be that guy but I really don't understand why you'd bother with the massive potential headache of supporting this "security" in mpd when it would be easier and safer for users to use an existing reverse-proxy server. They'd get something well tested with support for a variety of auth mechanisms. Am I missing something?

MaxKellermann:
> when it would be easier and safer for users to use an existing reverse-proxy server

The whole idea of the `httpd` output plugin is that MPD gives you something with batteries included; it may have fewer features than other solutions (Icecast or nginx), but you can do most things with just MPD and nothing else.

Of course, this may replicate features that are available with more complex solutions, but MPD's `httpd` may be the sweet spot for some users. Others want a fully-featured solution which comes with complexity.

For those people who want to make a stream available but want some protection, like an unencrypted password on an unencrypted HTTP connection - then it's fine to add this feature. It's weak, but for some people, it's just the right amount of security. Sure, somebody could sniff the password from the wire, and then they can listen to the stream. But the MPD user can't be accused of copyright violation by broadcasting copyrighted works - it's not a broadcast if it's protected, even if the protection is weak.

I understand your complaint, and I know this plugin and this PR is very imperfect, but it suits some kind of MPD users. Not me, and not you. But that's okay.

src/output/plugins/httpd/HttpdClient.cxx
MaxKellermann: Given that the password is transmitted without any encryption, it would be a lot easier to just capture the plain password from the net than to do timing-based guesses, so no, constant-time string comparisons would be pointless overengineering here.

kingosticks:
That's fair enough. My only further comment is that it's not obvious how secure this is trying to be. Maybe a note in the appropriate documentation would help. We obviously can't rely on GitHub comments.

MaxKellermann:
> Maybe a note in the appropriate documentation would help.

Agree. The presence of this feature should not pretend it's truly secure, when it cannot possibly be (because the transport is not secure).

bubbleguuum:
Basic authentication is fairly secure when used in conjunction with https.
Otherwise Digest authentication is better, but more complicated to implement.

src/output/plugins/httpd/HttpdOutputPlugin.cxx
skidoo23: > Why not `nullptr` for "no password necessary"? That seems like a natural choice, and would make your code simpler.

Because MPD segfaults at startup if test for nulllptr and mpd.conf lacks requested parameter.

src/playlist/plugins/SoundCloudPlaylistPlugin.cxx (updated to support new SoundCloud API. PR soon)

```
(...)
static std::string token;
(...)

static bool
soundcloud_init(const ConfigBlock &block)
{
        token = block.GetBlockValue("auth_token", nullptr);
        if (token.empty()) {
(...)
```


mpd.conf
```
playlist_plugin {
        name "soundcloud"
        enabled"yes"
        oauth_token "secret"
}
```

Testing for "" results in config: option 'oauth_token' on line X was not recognized) logentry and MPD continues normal operation.

MaxKellermann: > Because MPD segfaults at startup if test for nulllptr and mpd.conf lacks requested parameter.

.... what?? I don't get it. Why does MPD crash? Because MPD is buggy or because this new code is buggy? Or because the code you imagine is buggy?

skidoo23: @MaxKellermann maybe I misunderstood you. Your suggestion was to pass block.GetBlockValue("something", nullptr) as the second argument nullptr. In the event that "something" does not exist in mpd.conf, a segfault occurs at startup. I can't say whether it's an MPD bug.

MaxKellermann: If you just replace `""` with `nullptr`, of course it will crash. That's not how code hacking works, if you just replace something without thinking and without understanding the code.

Mrkvak:
Damn, sorrry, I've totally forgot this is still open.
@MaxKellermann can you please reopen it? I'll look at it later today.

src/output/plugins/httpd/HttpdOutputPlugin.cxx
Mrkvak: Yeah, I agree it is nicer. Implemented.

src/output/plugins/httpd/HttpdClient.cxx
Mrkvak: If you *really* want, I have no problem with implementing it in constant-time, but as the password is stored in config file in plaintext, and transmitted over unencrypted connection, I think it's indeed a bit too much.

doc/mpdconf.example
```diff
@@ -283,6 +283,7 @@ input {
 #	bitrate		"128"			# do not define if quality is defined
 #	format		"44100:16:1"
 #	max_clients	"0"			# optional 0=no limit
+#	password		"secretPassword"	# protect stream with HTTP Basic authentication (user can be anything)
```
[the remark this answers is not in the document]
Mrkvak: I've explicitly written in documentation can contain all printable characters except colon (the code does not check it, though, but I don't think it should be necessary).

src/output/plugins/httpd/HttpdClient.cxx
Mrkvak: Done.

src/output/plugins/httpd/HttpdClient.hxx
Mrkvak: Okay, empty password in config file now behaves exactly as no password.

src/output/plugins/httpd/HttpdClient.cxx
Mrkvak: Fair point. Should be done, hopefully correctly.

src/output/plugins/httpd/HttpdClient.cxx
Mrkvak: Done.

src/output/plugins/httpd/HttpdClient.cxx
Mrkvak: Done.

src/output/plugins/httpd/HttpdClient.cxx
Mrkvak: Yeah, nice. Done.

Mrkvak:
@kingosticks @MaxKellermann basically explained my motivation behind this. If someone sniffs the password, they can also sniff the audio stream. My goal was to have MPD not publicly broadcasting when accessible from network/internet. Nothing more, nothing less. 
I've added clear warning that the password is transmitted over unencrypted connection - it should be obvious, but I think it's still a good idea.

I think all feedback (thanks a lot for the review, by the way) was implemented, I've tested the code and it seems to be working.

MaxKellermann:
Ouch, the PR branch now contains merge commits. This is a big mess that cannot be reviewed. Please fix this.

Mrkvak:
Sigh... I guess that's my fault for trying to use github gui. Should be rebased on MPD/master without any pesky merge commits.

MaxKellermann:
There are now commits which obviously fix mistakes in earlier commits of this PR, e.g. "working version". I don't want to read known-bad commits and find all those mistakes again, only to see later that those mistakes have been fixed already. If you find a mistake in a commit that is not yet merged, please amend the commit instead of adding a fixup commit.

Mrkvak:
Squashed

src/output/plugins/httpd/HttpdClient.cxx
MaxKellermann: No, you havn't done that.

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -1,22 +1,31 @@
...
 #include <fmt/core.h>
 
 #include <cassert>
 #include <cstring>
 
+#ifdef HAVE_BASE64
+#include "lib/crypto/Base64.hxx"
+#endif
+
+
```
MaxKellermann, line 28: Remove second empty line

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -102,6 +115,15 @@ HttpdClient::HandleLine(const char *line) noexcept
 			return true;
 		}
 
+
+		#ifdef HAVE_BASE64
```
MaxKellermann, line 119: Preprocessor directives should not be indented

src/output/plugins/httpd/HttpdClient.hxx
```diff
@@ -69,6 +69,12 @@ class HttpdClient final
 	 */
 	bool should_reject = false;
 
+	/**
+	 * Should we reject this request as unauthorized?
+	 */
+	bool should_reject_unauthorized = false;
```
MaxKellermann, line 75: This field should only exist when the feature is available

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -169,7 +200,8 @@ HttpdClient::HttpdClient(HttpdOutput &_httpd, UniqueSocketDescriptor _fd,
 			 bool _metadata_supported)
 	:BufferedSocket(_fd.Release(), _loop),
 	 httpd(_httpd),
-	 metadata_supported(_metadata_supported)
+	 metadata_supported(_metadata_supported),
+	 provided_password{""}
```
MaxKellermann, line 204: What's the point of this initializer?

src/output/plugins/httpd/HttpdClient.hxx
```diff
@@ -111,6 +117,11 @@ class HttpdClient final
 	 */
 	unsigned metadata_fill = 0;
 
+	/**
+	 * Provided password (using HTTP basic auth)
+	 */
+	std::string provided_password;
```
MaxKellermann, line 123: This field should only exist when the feature is available

src/output/plugins/httpd/HttpdInternal.hxx
```diff
@@ -115,6 +115,12 @@ private:
 	 */
 	char const *const website;
 
+	/**
+	 * The configured password.
+	 */
+	char const *password;
```
MaxKellermann, line 121: This field should only exist when the feature is available and it should be `const`

src/output/plugins/httpd/HttpdInternal.hxx
```diff
@@ -115,6 +115,12 @@ private:
...
+
+
```
MaxKellermann, line 123: Two blank lines?

MaxKellermann (changes requested):

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -1,22 +1,31 @@
...
 #include "HttpdInternal.hxx"
 #include "util/ASCII.hxx"
 #include "util/AllocatedString.hxx"
 #include "Page.hxx"
 #include "IcyMetaDataServer.hxx"
+#include "lib/crypto/Base64.hxx"
 #include "net/SocketError.hxx"
 #include "net/UniqueSocketDescriptor.hxx"
 #include "util/SpanCast.hxx"
+#include "util/StringCompare.hxx"
```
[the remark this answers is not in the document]
Mrkvak: Missed leftover, sorry.

src/output/plugins/httpd/HttpdClient.cxx
Mrkvak: I don't think it's needed after the last simplification, sorry I missed it.

src/output/plugins/httpd/HttpdClient.cxx
Mrkvak: Done.

src/output/plugins/httpd/HttpdClient.cxx
Mrkvak: Ooops.

src/output/plugins/httpd/HttpdClient.cxx
Mrkvak: ok.

src/output/plugins/httpd/HttpdInternal.hxx
Mrkvak: fixed.

src/output/plugins/httpd/HttpdClient.cxx
```diff
@@ -41,6 +50,10 @@ HttpdClient::BeginResponse() noexcept
 {
 	assert(state != State::RESPONSE);
 
+	if (!head_method) {
```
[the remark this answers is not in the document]
Mrkvak: I don't think it's necessary to password protect HEAD requests, is it?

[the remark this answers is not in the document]
MaxKellermann: You wrote extra code to implement a special case to allow HEAD requests. What justifies this extra complexity?
Your answer isn't good because extra complexity should be justified, not the other way round.

MaxKellermann:
"If you find a mistake in a commit that is not yet merged, please amend the commit instead of adding a fixup commit."

PR #1971, opened 2024-01-18
android: Refactor settings UI into screens and add a bottom bar.

DDRBoxman:
This puts Status, logs, and settings all on different tabs. This gives us plenty more room to work to improve these views going forward


DDRBoxman:
Screenshot_20240118_012427
Screenshot_20240118_012435
Screenshot_20240118_012441

PR #1973, opened 2024-01-18
android: Correctly move the libmpd.so output

DDRBoxman:
libmpd.so was getting written as a file called `arm64-v8a` instead of `arm64-v8a/libmpd.so`


MaxKellermann:
Is there really no way to do that natively with Meson, without resorting to a shell script?
(And btw your shell script is unsafe .... it's the kind of awful shell script I want to avoid - shell is the worst of all programming languages.)

MaxKellermann:
No reply.

DDRBoxman:
Looking around I didn't find a way to have meson just move a file, not sure what the right fix is here.

MaxKellermann:
Maybe *moving* the file wasn't a good idea anyway, because moving an output file will cause it to be rebuilt; and copying files is something Meson can do: https://mesonbuild.com/Fs-module.html#copyfile

DDRBoxman:
Ah, nice I'll take a look at that, thanks!

DDRBoxman:
Hmm, looks like fs can't depend on other modules and is used to move stuff around for the build not after.
I can at least redo the script in python so it's not bash.

PR #1976, opened 2024-01-26
config: remove Spotify leftovers

skidoo23:
Delete some configrations options from the old Spotify plugin


MaxKellermann:
I was never sure whether removing these options is a good idea - even if they have been defunct for years, it would prevent MPD from starting if a user happens to still have these options set, making MPD not backwards-compatible. Sure, it would be better if all users would remove those options...

skidoo23:
According to release notes support for (de)spotify was removed from MPD in 2015.

jcorporation:
I vote for removing this old options.

PR #1992, opened 2024-02-08
New command countplaylist

jcorporation:
Count the number of songs and their total playtime in the playlist.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,7 +10,7 @@ ver 0.24 (not yet released)
   - show PCRE support in "config" response
   - apply Unicode normalization to case-insensitive filter expressions
   - stickers on playlists and some tag types
-  - new command "stickernames"
+  - new commands "stickernames" and "playlistlength"
   - new "search"/"find" filter "added-since"
 * database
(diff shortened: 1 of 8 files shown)


MaxKellermann:
The documentation show describe the response, e.g. show an example response.
I don't like the command name, it's confusing - this sounds like counting playlists, but it actually counts the contents of the playlist. Maybe `playlistsize`? Slightly better, but not perfect. Can you come up with a better name? `playlisttotals`? (still ugly)

jcorporation:
`playlistsize` sounds good for me.

jcorporation:
> Another option could be playlistlength, which is also used in [status response](https://mpd.readthedocs.io/en/stable/protocol.html#command-status) to indicate the length of the playlist (queue).

We calculate the length and the duration with this command. `playlistlength` seems to explicit for me. I would vote for playlistsize, but it is the decision of @MaxKellermann.

jcorporation:
> The documentation show describe the response, e.g. show an example response.

Added a example.

jcorporation:
Renamed to playlistlength

PR #1994, opened 2024-02-18
Clarifications on client-to-client and idle commands 

naglis:
- Document client-to-client message limits (client can subscribe to 16 channels max; client can have 64 unread messages max).
- Add note about handling idle events with care when using client-to-client messages to not miss the `message` idle event (see also commit message of 8fe563422fddcc7d41c979d07495dc98b2bdfa06 for a possible scenario).
- Add note about possibility of missing idle events of other subsystems when idling on particular subsystem(s).
- Clarify `message` idle event description.


MaxKellermann:
Oh. Missed idle events. That's not how it should be, but this bug has been present ever since I implemented idle events 15 years ago. Good catch, but documenting this behavior is not how I'd like to solve this.

PR #1997, opened 2024-02-25
Extend robustness of cdda playback

lazypingu:
This PR tries to address some CDDA playback issues.

**Changes:**
1. Moved speed adjutment after `cdio_cddap_open` as done in cd-paranoia (this seems to prevent errors on speed adjustment)
2. Use sector functions from libcdio-paranoia (`cdio_cddap_*`) to detect sectors (this seems to increase the robustness on special cases like hidden tracks or multisession CDs)
3. Added checks for errors on lsns and non audio tracks

**Improvments:**
* playback of hidden audio tracks (track 0) if present
* complete playback of last track on cd-extra CDs (multisession CDs)
* skipping of tracks on errors like no hidden track, out of range (> num-tracks) or non audio tracks

### DIFF
--- a/src/input/plugins/CdioParanoiaInputPlugin.cxx
+++ b/src/input/plugins/CdioParanoiaInputPlugin.cxx
@@ -224,7 +224,10 @@ input_cdio_open(const char *uri,
 	if (speed > 0) {
 		FmtDebug(cdio_domain, "Attempting to set CD speed to {}x",
 			 speed);
-		cdio_cddap_speed_set(drv,speed);
+		/* Negative value indicate error (e.g. -405: not supported) */
+		if (cdio_cddap_speed_set(drv,speed) < 0)
+			FmtDebug(cdio_domain, "Failed to set CD speed to {}x",
+				 speed);
 	}
 
 	bool reverse_endian;
--- a/src/input/plugins/CdioParanoiaInputPlugin.cxx
+++ b/src/input/plugins/CdioParanoiaInputPlugin.cxx
@@ -257,11 +257,11 @@ input_cdio_open(const char *uri,
 
 	lsn_t lsn_from, lsn_to;
 	if (parsed_uri.track >= 0) {
-		lsn_from = cdio_get_track_lsn(cdio, parsed_uri.track);
-		lsn_to = cdio_get_track_last_lsn(cdio, parsed_uri.track);
+		lsn_from = cdio_cddap_track_firstsector(drv, parsed_uri.track);
+		lsn_to = cdio_cddap_track_lastsector(drv, parsed_uri.track);
 	} else {
-		lsn_from = 0;
-		lsn_to = cdio_get_disc_last_lsn(cdio);
+		lsn_from = cdio_cddap_disc_firstsector(drv);
+		lsn_to = cdio_cddap_disc_lastsector(drv);
 	}
 
 	return std::make_unique<CdioParanoiaInputStream>(uri, mutex,
--- a/src/input/plugins/CdioParanoiaInputPlugin.cxx
+++ b/src/input/plugins/CdioParanoiaInputPlugin.cxx
@@ -264,6 +264,16 @@ input_cdio_open(const char *uri,
 		lsn_to = cdio_cddap_disc_lastsector(drv);
 	}
 
+	/* LSNs < 0 indicate errors (e.g. -401: Invaid track, -402: no pregap) */
+	if(lsn_from < 0 || lsn_to < 0)
+		throw FmtRuntimeError("Error {} on track {}",
+				      lsn_from < 0 ? lsn_from : lsn_to, parsed_uri.track);
+
+	/* Only check for audio track if not pregap or whole CD */
+	if (!cdio_cddap_track_audiop(drv, parsed_uri.track) && parsed_uri.track > 0)
+		throw FmtRuntimeError("No audio track: {}",
+				      parsed_uri.track);
+
 	return std::make_unique<CdioParanoiaInputStream>(uri, mutex,
 							 drv, cdio,
 							 reverse_endian,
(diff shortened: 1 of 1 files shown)


src/input/plugins/CdioParanoiaInputPlugin.cxx
```diff
@@ -251,14 +254,25 @@ input_cdio_open(const char *uri,
...
 
-	lsn_t lsn_from, lsn_to;
-	if (parsed_uri.track >= 0) {
-		lsn_from = cdio_get_track_lsn(cdio, parsed_uri.track);
-		lsn_to = cdio_get_track_last_lsn(cdio, parsed_uri.track);
-	} else {
-		lsn_from = 0;
-		lsn_to = cdio_get_disc_last_lsn(cdio);
-	}
+	lsn_t lsn_to, lsn_from;
```
MaxKellermann, line 258: All this change does is reorder the declaration, but why are you doing this? This makes the diff harder to read.

src/input/plugins/CdioParanoiaInputPlugin.cxx
```diff
@@ -214,18 +214,21 @@ input_cdio_open(const char *uri,
...
-		cdio_cddap_speed_set(drv,speed);
-	}
 
 	if (0 != cdio_cddap_open(drv)) {
 		cdio_cddap_close_no_free_cdio(drv);
 		cdio_destroy(cdio);
 		throw std::runtime_error("Unable to open disc.");
 	}
 
+	if (speed > 0) {
```
MaxKellermann, line 224: I don't quite get why you move this block, and your commit message doesn't explain it. Also you add a check to the `cdio_cddap_speed_set()` return value, but since you do this in the same commit that moves the code, it's hard to see.

src/input/plugins/CdioParanoiaInputPlugin.cxx
```diff
@@ -251,14 +254,25 @@ input_cdio_open(const char *uri,
...
+		lsn_to = cdio_cddap_disc_lastsector(drv);
+	} else{
+		lsn_from = cdio_cddap_track_firstsector(drv, parsed_uri.track);
+		lsn_to = cdio_cddap_track_lastsector(drv, parsed_uri.track);
+	} 
+
+	/* LSNs < 0 indicate errors (e.g. -401: Invaid track, -402: no pregap) */
+	if(lsn_from < 0 || lsn_to < 0)
+		throw FmtRuntimeError("Error {} on track {}",
+							lsn_from < 0 ? lsn_from : lsn_to, parsed_uri.track);
```
MaxKellermann: If you find a mistake like this wrong indent in your commit before your PR gets merged, please amend the commit instead of adding a fixup commit.
I tried to review your comment, found this bad indentation, and thought about adding a comment; but then I saw you had a fixup. If you avoid this noise by amending known-bad commits, you save me some confusion.

src/input/plugins/CdioParanoiaInputPlugin.cxx
```diff
@@ -251,14 +254,25 @@ input_cdio_open(const char *uri,
...
-	if (parsed_uri.track >= 0) {
-		lsn_from = cdio_get_track_lsn(cdio, parsed_uri.track);
-		lsn_to = cdio_get_track_last_lsn(cdio, parsed_uri.track);
-	} else {
-		lsn_from = 0;
-		lsn_to = cdio_get_disc_last_lsn(cdio);
-	}
+	lsn_t lsn_to, lsn_from;
+
+	if(parsed_uri.track < 0){
```
MaxKellermann, line 260: This reorders the if/else branches. Why?
Also you change the coding style by removing the space between `if(` - why?
Both just adds noise and makes your diff hard to review.

src/input/plugins/CdioParanoiaInputPlugin.cxx
lazypingu: This was probably just a testing remnant. Fixed!

src/input/plugins/CdioParanoiaInputPlugin.cxx
lazypingu: Your right! I've missed to clean this up after testing. Sorry! Fixed!

src/input/plugins/CdioParanoiaInputPlugin.cxx
lazypingu: I've just noticed that setting the drive speed seemed to fail on my drive.
On [cd-paranoia](https://github.com/rocky/libcdio-paranoia/blob/c93c72b91585d7d06dfbd868a3ada8e69fcc164b/src/cd-paranoia.c#L1061) `speed_set` is called after `cdda_open`, so I just tried it here and it seemed to fix the issue on my side.

Then I added the check for debugging.

MaxKellermann:
Now there are two more fixup commits for known-bad PR commits.

src/input/plugins/CdioParanoiaInputPlugin.cxx
MaxKellermann: Your commit message should explain this.

lazypingu:
@MaxKellermann Got your point. Thanks! I've tried to rewrite the history with all suggestions.

MaxKellermann:
I find your commit messages very lacking still.
- "Move cddap_speed_set below cddap_open as cd-paranoia does to avoid errors on setting speed" - what errors are avoided, and why are they avoided if ordered that way?
- "Use sector functions from libcdio-paranoia (cdio_cddap_*) to detect sectors" - I have NO CLUE what this means and why you propose changing this and what the consequences are.
- "Add checks for errors on lsns and non audio tracks" - what is "errors on lsns"?

lazypingu:
@MaxKellermann I've tried to specify the commit messages for more clarity.

PR #2003, opened 2024-03-04
db/update/Walk:FindAncestorLoop: uint64_t inode & device

altblue:


MaxKellermann:
Not a good commit message. It only describes what is obvious, but doesn't explain why.

MaxKellermann:
Since you don't want to fix this, I edited the commit message for you: 7b6909f2c0dc82e39ba03ed0bf7d06f9fa62d5bf

PR #2018, opened 2024-04-07
android: Button to jump to bottom of log view

DDRBoxman:
This adds a button that can jump to the bottom of the log view. If the user scrolls up we now disable the auto scroll down and show the down button. When the down button is clicked the auto scroll resumes and the button is removed.


DDRBoxman:
Screenshot_20240406_200926

PR #2020, opened 2024-04-07
event/TimerList: Add custom compare for android

DDRBoxman:
NDK doesn't seem to support using std::compare_three_way for time_point so this implements that.

### DIFF
--- a/src/event/TimerList.hxx
+++ b/src/event/TimerList.hxx
@@ -4,11 +4,22 @@
 
 #pragma once
 
+#include <compare>
+
 #include "Chrono.hxx"
 #include "event/Features.h"
 #include "util/IntrusiveTreeSet.hxx"
 
 class FineTimerEvent;
+
+struct custom_compare {
+  auto operator()(const std::chrono::steady_clock::time_point& lhs,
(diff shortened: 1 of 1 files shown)


DDRBoxman:
Interesting looks like macOS is failing on std::compare_three_way missing too 🤔

MaxKellermann:
> Interesting looks like macOS is failing on std::compare_three_way missing too 🤔

It's because both Android NDK and macOS ship with an outdated libc++ version which didn't implement this part of C++20.

I don't like this PR because it adds code noise to all platforms, even those that are standards-compliant. At least this should be `#if`ed and come with a proper explanatory comment. But I guess we could just as well wait for NDK r27 which should be here any day now.

(I don't care for macOS. It's a broken platform with a broken toolchain.)

DDRBoxman:
I'm fine with waiting for r27 instead

MaxKellermann:
r27-beta1 is out: https://dl.google.com/android/repository/android-ndk-r27-beta1-linux.zip
(I havn't yet tested whether this release fixes the libc++ bug.)

MaxKellermann:
Now tested. NDK r27-beta1 has the fix for this C++20 bug.

PR #2028, opened 2024-04-21
android: Update to NDK 27

DDRBoxman:


DDRBoxman:
Wanted to put this up with the changes I needed to compile with the NDK 27 beta. Probably makes sense to wait for the stable release before we actually merge anything.

MaxKellermann:
> Probably makes sense to wait for the stable release before we actually merge anything.

NDK r26 is broken; it can't compile MPD at all. Everything is better than that.

I'd like to merge this PR, but I don't understand the `-D_LIBCPP_PROVIDES_DEFAULT_RUNE_TABLE` line. Can you add an explanatory comment to that line?

DDRBoxman:
Added a comment for `-D_LIBCPP_PROVIDES_DEFAULT_RUNE_TABLE`

PR #2031, opened 2024-04-23
android: Install the libmpd.so output on android

DDRBoxman:
libmpd.so was getting written as a file called `arm64-v8a` instead of `arm64-v8a/libmpd.so`


MaxKellermann:
This time you're adding a Python script instead of a shell script, which is better... but this still looks more complicated than it should be. Isn't there another, simpler way to do this?
Also your old commit 906d58a918d looks pretty wrong. It does not properly declare the target file; it pretends to output to `OUTPUT/android/apk/libmpd.so` (via `output:` parameter) but it really copies to `meson.current_source_dir()/....` - due to this mistake, Ninja is unable to find the output file and thus copies the file again even if it already exists.

DDRBoxman:
Looks like `output` is just for names and can't take a path. Setting up the install path and calling install might be an option?

DDRBoxman:
Okay, switched this over to use install

PR #2033, opened 2024-04-23
Remove deprecation flag for volume command

jcorporation:


MaxKellermann:
NEWS, please!

jcorporation:
I thought this was not worth a NEWS entry...added

MaxKellermann:
It's a user-visible change that should be announced IMO, just like the deprecation was announced.

PR #2064, opened 2024-06-29
Added explicitly case sensitive/insensitive filter operators.

geneticdrift:
The default case sensitivity is hard coded for each command. These operators allow to override this default case sensitivity.

### DIFF
--- a/doc/protocol.rst
+++ b/doc/protocol.rst
@@ -243,6 +243,43 @@ applies `Unicode normalization <https://unicode.org/reports/tr15/>`__
 and converts all punctuation to ASCII equivalents
 if MPD was compiled with `ICU <https://icu.unicode.org/>`__ support.
 
+Explicit case-sensitivity [#since_0_24]_
+----------------------------------------
+
+.. note:: The following variants of filter operators override the default case sensitivity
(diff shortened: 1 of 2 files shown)


geneticdrift:
Related to #2062 and #494

MaxKellermann:
This is probably a good idea, but the code doesn't look so good.
- `tryParseStringFilter()` (badly named, shouldn't start with lower case) allocates memory for no reason
- `ParseStringFilter()` is being refactored AND new code is being added in one single commit; this makes understanding your code very hard because I can't see what is really being changed
- every `ParseStringFilter()` call allocates memory on the stack for the `std::array` and initializes it for no reason; if you really wanted to refactor the function to use an array, then initalize it statically (`static constexpr`). But since you mixed refactoring and the new feature, I don't understand why you need this array at all
- don't concatenate `std::string`; this will create many unnecessary allocations. If you really want to format exception messages, use libfmt (e.g. via the `FmtRuntimeError()` wrapper function)

MaxKellermann:
The array is still not `constexpr`, thus it's not guaranteed to initialize at compile time; and the function still allocates heap memory.

PR #2082, opened 2024-07-16
add -k alias for --kill

spencermp:
In my opinion, this flag make sense and is present in other daemons' kill functions. The largest example that comes to mind is pulse. 
Plus it's shorter to type -k than --kill, which is something I type more often than --verbose or --no-daemon for example.


MaxKellermann:
No, I don't want anybody to use this option at all, therefore I don't want to extend this feature. People should be running MPD with a service manager. PID files and self-killing of daemons is obsolete technology.

PR #2085, opened 2024-07-21
feat: Add ShowMovement Tag

mayanez:
This PR adds an additional tag as part of the effort originally started with https://github.com/MusicPlayerDaemon/MPD/pull/858. 

It adds the `SHOWMOVEMENT` [tag](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#show-work-movement-4). Historically, this tag originates from iTunes' MP4, but has since become widely used. It is created by Picard's Classical music [plugins](https://picard-docs.musicbrainz.org/en/variables/variables_classical.html) (such as "Classical Extras" or "Work & Movement"). 

The reasoning behind this tag is to display Work & Movement titles without redundant information and in a more uniform way. Moreover, it additionally serves as an implicit marker denoting classical music tracks (genre tags aren't sufficient).

The screenshot below provides an example of its use-case. 

My tracks are titled as follows: `%composer% • %work%: %movement%`. If the client so chooses to support it, they can display `Work` and `Movement` instead of the track title allowing for cleaner display. Other clients can continue to display  the `%title%` as before without any fuss.
 
**Without `ShowMovement`:**

image

**With `ShowMovement`:**

image

Some might wonder, why not just change the `%title%` to be the` %movement%`. Unfortunately, this breaks compatibility with software that simply doesn't know how to display classical music (to which the unfortunate answer is close to 99%). Classical music metadata is hard :crying_cat_face: 

Please do let me know if I missed anything (in particular to the `libmpdclient` version as I wasn't too sure about that) or would like me to elaborate further on the use-case. 

P.S. I have the changes for `libmpdclient` should this be merged.


MaxKellermann:
This PR contains text that you omitted in the commit message. PR text only exists in GitHub's proprietary database, but we should preserve all explanations.

mayanez:
I'd be happy to add the explanation to the commit. 

Are there any specific formatting rules I should be aware of?

mayanez:
I added summary text to the commit message as suggested.

PR #2089, opened 2024-07-30
Dynamically update remote stream tags by merge instead of replace - i…

geneticdrift:
…ssue #1400

Clients can use addtagid to set tags for remote streams. For example a radio station genre. Then when the song title is updated from the remote stream, only the title is updated but other tags are preserved.

### DIFF
--- a/doc/protocol.rst
+++ b/doc/protocol.rst
@@ -957,6 +957,10 @@ Whenever possible, ids should be used.
     the server, and the data is gone when the song gets
     removed from the queue.
 
+    Tags received from the server are merged [#since_0_24]_, i.e.
+    only received tags are overwritten, but other tags are not affected
+    (previously they were removed).
+
 .. _command_cleartagid:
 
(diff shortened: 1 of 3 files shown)


MaxKellermann:
But what if the stream switches to a new song which doesn't have a certain tag type; e.g. it has no "Composer"; then the Composer from the previous song remains forever, doesn't it?

geneticdrift:
You are right.
It's a bad idea to try to shoehorn a "Radio station" entity into a model that only defines a "Song". It doesn't fit.
Unless the model directly supported entities such as Radio station, Podcast, Audio book etc. it will just be a hack.

ghaspias:
> But what if the stream switches to a new song which doesn't have a certain tag type; e.g. it has no "Composer"; then the Composer from the previous song remains forever, doesn't it?

It is the user's choice and determination to add tags. The user should be respected, not the stream. If the user does want some tag to be updated from the stream, he just won't set it in first place.

PR #2090, opened 2024-07-30
New command stickernamestypes

jcorporation:
Lists sticker names with their types and optionally filters by type


src/command/StickerCommands.cxx
```diff
@@ -122,6 +122,25 @@ class DomainHandler {
 		return CommandResult::OK;
 	}
 
+	virtual CommandResult NamesTypes(const char *type) {
```
MaxKellermann, line 125: Why is this method virtual?

Issue #938, opened 2020-08-08
Loaded playlist name ?


XpamAmAdEuS:
Can we see name of the last loaded playlist?
I have some kind of plugin that load playlists by date/code/name every day.
I cannot know name of playlist currently loaded into mpd.

MaxKellermann:
No.
How is that supposed to work? Let's say I send `load a`, `load b` and add a few songs to the queue and then remove a few - what will MPD tell you which playlist is loaded? Because it's a combination of "a", "b" and some manual changes.

geneticdrift:
Just an idea: add a new property "queue_title" to the player queue. Set this to the playlist name when loaded and clear it when the queue is changed. The "queue_title" will be included in the status response. Optionally add commands to set and clear the queue title so clients can set it not only when loading a playlist.

XpamAmAdEuS:
> Just an idea: add a new property "queue_title" to the player queue. Set this to the playlist name when loaded and clear it when the queue is changed. The "queue_title" will be included in the status response. Optionally add commands to set and clear the queue title so clients can set it not only when loading a playlist.

Great idea.
For my needs its ok if it returns last loaded playlist name becouse i must always start with clear queue.
But command for set name its great idea.
Even better if we can set additional infos like datetime added.

PR #2101, opened 2024-08-18
Store last loaded playlist

jcorporation:
Stores the last loaded playlist uri in the queue struct. Last loaded playlist is set on load command and cleared with the clear command. The last loaded playlist is printed in the status command and saved / restored from the partition state file.

Resolves issue #938

### DIFF
--- a/src/queue/Playlist.hxx
+++ b/src/queue/Playlist.hxx
@@ -102,6 +102,14 @@ struct playlist {
 	[[gnu::pure]]
 	int GetNextPosition() const noexcept;
 
+	const std::string_view GetLastLoadedPlaylist() const noexcept {
+		return queue.last_loaded_playlist;
+	}
+
+	void SetLastLoadedPlaylist(const char *playlist_name) noexcept {
+		queue.last_loaded_playlist = playlist_name;
+	}
+
 	/**
 	 * Returns the song object which is currently queued.  Returns
 	 * none if there is none (yet?) or if MPD isn't playing.
--- a/src/queue/Queue.cxx
+++ b/src/queue/Queue.cxx
@@ -270,6 +270,7 @@ Queue::Clear() noexcept
 	}
 
 	length = 0;
+	last_loaded_playlist.clear();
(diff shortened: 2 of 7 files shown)


src/queue/Playlist.hxx
```diff
@@ -102,6 +102,9 @@ struct playlist {
 	[[gnu::pure]]
 	int GetNextPosition() const noexcept;
 
+	const std::string getLastLoadedPlaylist() const noexcept;
```
MaxKellermann, line 105: Bad method name, but the worst thing is that this allocates a copy of the string on the heap.

src/queue/Queue.cxx
```diff
@@ -270,6 +270,7 @@ Queue::Clear() noexcept
 	}
 
 	length = 0;
+	last_loaded_playlist = "";
```
MaxKellermann, line 273: This has a smell - `last_loaded_playlist.clear();` is better and faster

src/queue/Queue.hxx
```diff
@@ -85,6 +85,9 @@ struct Queue {
 	/** play back songs in random order? */
 	bool random = false;
 
+	/** Last loaded playlist */
+	std::string last_loaded_playlist = "";
```
MaxKellermann, line 89: This initializer has no effect (except for adding runtime overhead)

jcorporation: Removed

src/queue/Queue.cxx
jcorporation: Replaced

src/queue/Playlist.hxx
jcorporation: Renamed to starting uppercase and converted to std::string_view. I hope this is correct!

MaxKellermann:
There's no protocol documentation; and the `playlist` get/set methods should be inlined - they are trivial, after all.

jcorporation:
Added documentation and inlined the functions.

PR #2102, opened 2024-08-19
Add config option to disable the listing of playlists in the root folder for the lsinfo command.

jcorporation:
Default is enabled to keep backward compatibility.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -67,6 +67,7 @@ ver 0.24 (not yet released)
 * require libfmt 9 or later
 * documentation: switch to sphinx-rtd-theme
 * require Meson 1.0
+* option to disable listing of playlists with lsinfo in the root folder
 
 ver 0.23.15 (2023/12/20)
 * decoder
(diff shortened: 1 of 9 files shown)


MaxKellermann:
I can imagine 4 ways dealing with this deprecation:
1. leave it as it is, forever
2. remove it completely
3. make it a per-connection setting
4. make it a global config option

This is 4, but I fear it's the worst way. No user knows about this protocol-level detail and no user should have to know about this. This means no client can ever take advantage of this.
1 is bad because nobody likes to support deprecated legacy stuff, and it may require client-side kludges to suppress it.
2 is bad because some random client may rely on this, and if somebody complains, they're right.

3 seems like the best way to deal with this: every client can decide; newer clients can explicitly opt into new protocol semantics, easily. We just need a protocol command for such an opt-in. I had been thinking about this a few times, but never came up with a design.

We have something similar with tag types: the "old" way to enable/disable tag types is the `metadata_to_use` setting, but this discards tags while scanning; it's a fragile design that doesn't allow clients to opt-in and opt-out, meaning either they have too much protocol overhead or not enough information. Then I added a `tagtypes` command which allowed clients to choose exactly which tag types they were interested in. This is fully backwards compatible, but allows modern clients to reduce overhead.

jcorporation:
Thanks for your detailed reply. I close this pull request and open an issue to discuss the protocol design for this. I will gladly implement this feature. myMPD implements a client-side kludges to suppress it and I am not happy with it.

PR #2109, opened 2024-09-01
Print positions in the search playlist function

jcorporation:
For simple listing this can be done by the client, but for the search case the client can not determine the correct position.

I hope this does not break existing client implementations.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -16,6 +16,7 @@ ver 0.24 (not yet released)
   - "sticker find" supports sort and window parameter and new sticker compare operators "eq", "lt" and "gt"
   - consume only idle flags that were subscribed to
   - volume command is no longer deprecated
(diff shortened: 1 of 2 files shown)


MaxKellermann:
If no detail was requested, MPD shouldn't ever print anything other than the URI.
If detail was requested, the position should only be printed if it is not redundant information, for example if this is a search result - the thing you're aiming at. Adding overhead to all response is useless.
(I consider this a mistake in the "playlistinfo" response - the "Pos" is always printed for no reason; but we can't remove it now.)

jcorporation:
Reworked to print the Pos only for search playlist command.

MaxKellermann:
But it's still printed if no detail was requested

jcorporation:
Fixed

PR #2116, opened 2024-09-14
Fix integer overflow calculating length of very large playlists

jcorporation:


MaxKellermann:
This loses type safety by downgrading from std::chrono to raw integers, but what do these raw integers mean? You make the core more obscure and less safe.

jcorporation:
Ok, migrated it to std::chrono::duration api.

src/playlist/Length.cxx
```diff
@@ -36,15 +36,16 @@ playlist_provider_length(Response &r,
...
 	while ((song = e.NextSong()) != nullptr) {
 		if (playlist_check_translate_song(*song, base_uri,
 						  loader))
 			playtime += get_duration(*song);
 		i++;
 	}
 	r.Fmt(FMT_STRING("songs: {}\n"), i);
-	r.Fmt(FMT_STRING("playtime: {}\n"), playtime.RoundS());
+	const auto seconds = std::chrono::duration_cast<std::chrono::seconds>(playtime);
+	r.Fmt(FMT_STRING("playtime: {}\n"), seconds.count());
```
MaxKellermann, line 48: This removes the rounding which was very explicit previously

src/playlist/Length.cxx
```diff
@@ -21,7 +21,7 @@
 
 static SignedSongTime get_duration(const DetachedSong &song) {
 	const auto duration = song.GetDuration();
-	return duration.IsNegative() ? (SignedSongTime)0 : song.GetDuration();
+	return duration.IsNegative() ? (SignedSongTime)0 : duration;
```
MaxKellermann, line 24: ?

jcorporation: As I get the song duration in the local variable there is no need to call the GetDuration function a second time.

src/playlist/Length.cxx
jcorporation: Replaced with std::chrono::round

src/playlist/Length.cxx
MaxKellermann: True, but it doesn't fix an integer overflow

PR #2137, opened 2024-10-26
Fix order of commands

jcorporation:


MaxKellermann:
I wonder if there's some fancy `constexpr` trick to sort the list at compile time....

PR #2140, opened 2024-10-27
meson.build: support building with ICU 76

cho-m:
ICU 76 decided to reduce overlinking[^1] thus `icu-i18n` will no longer add `icu-uc` when linking to shared libraries. This results in failure:
```
src/lib/icu/libicu.a.p/Converter.cxx.o: undefined reference to symbol 'ucnv_fromUnicode_76'
```

---

ICU 76:

```console
$ objdump -p libicui18n.so | grep icuuc
  NEEDED               libicuuc.so.76

$ pkg-config --libs-only-l icu-i18n
-licui18n

$ pkg-config --libs-only-l --static icu-i18n
-licui18n -licuuc -licudata -lpthread -lm

$ pkg-config --libs-only-l icu-i18n icu-uc
-licui18n -licuuc
```

ICU 75

```console
$ objdump -p libicui18n.so | grep icuuc
  NEEDED               libicuuc.so.75

$ pkg-config --libs-only-l icu-i18n
-licui18n -licuuc -licudata

$ pkg-config --libs-only-l --static icu-i18n
-licui18n -licuuc -licudata -lpthread -lm

$ pkg-config --libs-only-l icu-i18n icu-uc
-licui18n -licuuc -licudata
```

[^1]: https://github.com/unicode-org/icu/commit/199bc827021ffdb43b6579d68e5eecf54c7f6f56

### DIFF
--- a/src/lib/icu/meson.build
+++ b/src/lib/icu/meson.build
@@ -1,5 +1,7 @@
-icu_dep = dependency('icu-i18n', version: '>= 50', required: get_option('icu'))
-conf.set('HAVE_ICU', icu_dep.found())
+icu_i18n_dep = dependency('icu-i18n', version: '>= 50', required: get_option('icu'))
+icu_uc_dep = dependency('icu-uc', version: '>= 50', required: get_option('icu'))
+have_icu = icu_i18n_dep.found() and icu_uc_dep.found()
+conf.set('HAVE_ICU', have_icu)
 
 icu_sources = [
   'Canonicalize.cxx',
@@ -13,7 +15,7 @@ if is_windows
 endif
 
 iconv_dep = []
-if icu_dep.found()
+if have_icu
   icu_sources += [
     'Init.cxx',
     'Util.cxx',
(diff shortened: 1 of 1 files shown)


src/lib/icu/meson.build
```diff
@@ -1,5 +1,6 @@
-icu_dep = dependency('icu-i18n', version: '>= 50', required: get_option('icu'))
-conf.set('HAVE_ICU', icu_dep.found())
+icu_i18n_dep = dependency('icu-i18n', version: '>= 50', required: get_option('icu'))
+icu_uc_dep = dependency('icu-uc', version: '>= 50', required: get_option('icu'))
+conf.set('HAVE_ICU', icu_i18n_dep.found())
```
MaxKellermann, line 3: What if icu-i18n was found but not icu_uc?

cho-m: I can update to check both.

In theory, should be an unlikely scenario as `icu-i18n` has always used `icu-uc`. Though I guess there is no guarantee on what ICU will do in future.

MaxKellermann: True, this *should* be both available, but since this is out of our control, and users may do stupid things (and some *will* do), let's be sure, and it's so simple to check both...

Issue #2173, opened 2024-12-27
Failed compile with libfmt-11.1.0


heitbaum:
### Describe the bug
Compile with libfmt 1.11.0 fails (api changes)

## Expected Behavior
- Successful compile

## Actual Behavior
- error

## Version

0.23.16

## Log

Majority of the errors being `error: cannot convert`
```

[2/81] Compiling C++ object mpd.p/src_decoder_DecoderPrint.cxx.o
FAILED: mpd.p/src_decoder_DecoderPrint.cxx.o 
[...]

PR #2181, opened 2025-01-11
lib/fmt: drop use of FMT_STRING

heitbaum:
Fixes build with fmt-11.1.0
- https://github.com/fmtlib/fmt/issues/4304#issuecomment-2585356850
- #2173

### DIFF
--- a/src/SongPrint.cxx
+++ b/src/SongPrint.cxx
@@ -46,14 +46,14 @@ song_print_uri(Response &r, const char *uri, bool base) noexcept
 			uri = allocated.c_str();
 	}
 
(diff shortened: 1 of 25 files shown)


MaxKellermann:
Bad commit message - it only states the obvious, but has no explanation

heitbaum:
> Bad commit message - it only states the obvious, but has no explanation

Will do, I’ll write it up in detail tomorrow (RL got in the way today) based on the documentation shared by the fmt team.

heitbaum:
Update commit message based on information shared by the libfmt team (thanks) and the error being addressed.

MaxKellermann:
Thanks, that's much better.

PR #2183, opened 2025-01-14
Add `bind_to_address` parameter to docs

jbdyn:
Hi there!

While enjoying my favourite tunes with MPD, I wondered how to fix this error message on startup:

> server_socket: bind to '0.0.0.0:6600' failed (continuing anyway, because binding to '[::]:6600' succeeded): Failed to bind socket: Address already in use

The solution for that is to specify the `bind_to_address` parameter in the config file `mpd.conf`, as stated for example in [this Arch Linux Wiki section](https://wiki.archlinux.org/title/Music_Player_Daemon/Troubleshooting#Binding_to_IPV6_before_IPV4).

I noticed that this parameter is missing in the docs I read with `man 5 mpd.conf`. So, I decided to add it.

Do you agree with the description?

### DIFF
--- a/doc/mpd.conf.5.rst
+++ b/doc/mpd.conf.5.rst
@@ -89,6 +89,9 @@ user <username>
    initialization. Do not use this option if you start MPD as an unprivileged
    user.
 
+bind_to_address <address>
+   Set the interface address that mpd listens on. The default is 0.0.0.0 (all interfaces).
+
 port <port>
    This specifies the port that mpd listens on. The default is 6600.
 


MaxKellermann:
Your documentation is somewhat misleading; it talks about "interface addresses" and later about "interfaces", but that misleads users to believe that this is about (network) interfaces, but it's really about addresses. Addresses are bound to interfaces, yes, but writing this documentation causes confusion. This is already difficult enough to understand, but mixing up the words doesn't make it better.
And don't forget that this setting is already documented: https://mpd.readthedocs.io/en/stable/user.html#listeners
Just not in the abbreviated manpage. That's why the manpage says:
> This manual is not complete, it lists only the most important options.
> Please read the MPD user manual for a complete configuration guide:
> http://www.musicpd.org/doc/user/

I appreciate your effort to add documentation, but in this case, it's not necessary and not worded well.

jbdyn:
Ok, thanks for clarifying.

PR #2188, opened 2025-01-19
protocol: command 'binarylimit' without an argument returns the curre…

geneticdrift:
…nt binary limit and it's min, max, and default values

### DIFF
--- a/doc/protocol.rst
+++ b/doc/protocol.rst
@@ -1646,6 +1646,21 @@ Connection settings
 
 .. _command_binarylimit:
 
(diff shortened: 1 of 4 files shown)


MaxKellermann:
Your commit message contains no explanation. This seems like a useless even harmful feature: the existence of such a feature suggests using it is a good idea to use it, but all it does is add overhead to the MPD communication for no reason. If a client wants to set a certain limit, just set it, but what's the point of querying it?

geneticdrift:
OK

PR #2190, opened 2025-01-21
fix: output of command status broken by a misplace comma in the forma…

geneticdrift:
…t arguments


sp1ff:
@geneticdrift Thank you for catching this (I believe it was my merge that introduced the bug).

PR #2191, opened 2025-01-22
Log datetime

geneticdrift:


src/command/PlayerCommands.cxx
```diff
@@ -129,7 +129,7 @@ handle_status(Client &client, [[maybe_unused]] Request args, Response &r)
 	      COMMAND_STATUS_PLAYLIST ": {}\n"
 	      COMMAND_STATUS_PLAYLIST_LENGTH ": {}\n"
 	      COMMAND_STATUS_MIXRAMPDB ": {}\n"
-	      COMMAND_STATUS_STATE ": {}\n",
```
MaxKellermann, line 132: Which commit introduced this bug? This should be mentioned in the commit message so we know whether a backport is necessary.

MaxKellermann:
Your commit message doesn't contain an explanation for why you believe this is necessary.
I don't like `thread_local`. Your commit message doesn't explain why this is necessary.

geneticdrift:
Thanks for the review.

I've added more explanations in the commit messages as well as the commit number.

The logging timestamp format can be made optional by configuration if you prefer.
Personally I find the higher resolution timestamp more helpful in catching issues even when not looking for one specifically. And in cases of user postmortem logs when it's hard to reproduce an issue, so enabling higher resolution after the fact is not always helpful.

geneticdrift:
> I think forcing time stamps on stdout is a bad thing because then you have two time stamps on each line when logging to syslog/journald.

As far as I tell, enabling the timestamp in `log_init` only if `use_stdout` is `true` doesn't affect syslog or systemd, unless the command in the systemd unit includes `--stderr`.
Does this use case, of both `--stderr` and `systemd`, need to be supported? If so then I can add an opt-in in the configuration.
(I tested with syslog and systemd user unit, and both did not have double timestamp. The timestamp was only enabled when using `--stderr` command line option, or when using a log file).


> when two threads want to format a time stamp at exactly the same time

Yes, the chance of something bad happening is negligible. But to me it still feels wrong leaving it like than.
I've added  passing the buffer from the caller which I think is the better option. You decide.

BTW, why "try to avoid TLS whenever possible"?

geneticdrift:
> But what is "something bad"? Is your bad (subjective) feeling the worst that can happen, or do you have something else in mind? Is there a concrete bad consequence for this piece of not-100%-correct code?

Nothing concrete just subjective. Except maybe more CPU cache efficient? But again probably negligible.

> don't think millisecond resolution is necessary or even useful (and it costs screen space)

I've added a configurable timestamp that can be set to none, minutes, seconds, or milliseconds.

src/LogBackend.cxx
```diff
@@ -9,11 +9,12 @@
...
 #include "config.h"
 
+#include <fmt/chrono.h>
 #include <cassert>
-
-#include <stdio.h>
-#include <string.h>
-#include <time.h>
+#include <cstdio>
+#include <cstring>
```
MaxKellermann, line 15: Why change this line? It's not related to the rest of this commit, is it? Sneaking in unrelated changes makes a PR very hard to read, and there isn't even a good reason to do so.

src/LogBackend.cxx
```diff
@@ -9,11 +9,12 @@
...
+#include <chrono>
+#include <functional>
```
MaxKellermann, line 17: Why this header?

src/LogBackend.cxx
```diff
@@ -9,11 +9,12 @@
 #include "Version.h"
 #include "config.h"
 
+#include <fmt/chrono.h>
```
MaxKellermann, line 12: This commit not only adds configurable time formats, but also migrates to fmt. Please don't do two things in one commit! If you believe fmt is better, you can suggest doing that.

src/LogBackend.cxx
```diff
@@ -62,28 +71,56 @@ SetLogThreshold(LogLevel _threshold) noexcept
...
 
+template <typename DurationType>
+inline constexpr const char *date_format_of = "";
+
+template<>
+inline constexpr const char *date_format_of<std::chrono::minutes> = "{:%b %d %H:%M} : ";
+template<>
+inline constexpr const char *date_format_of<std::chrono::seconds> = "{:%b %d %H:%M:%S} : ";
+template<>
+inline constexpr const char *date_format_of<std::chrono::milliseconds> = "{:%b %d %H:%M:%S} : ";
```
MaxKellermann, line 82: Wow, that must be the most complicated way to do this. What's the reason for making this so complicated.

src/LogBackend.cxx
```diff
@@ -62,28 +71,56 @@ SetLogThreshold(LogLevel _threshold) noexcept
 	log_threshold = _threshold;
 }
 
+template <typename DurationType>
+inline constexpr const char *date_format_of = "";
```
MaxKellermann, line 75: `inline` is redundant here, but why is it not `static`? Without `static`, it's being exported and may conflict with symbols from other compilation units.

src/LogBackend.cxx
```diff
@@ -62,28 +71,56 @@ SetLogThreshold(LogLevel _threshold) noexcept
...
+template<typename DurationType>
+static const char *
+log_date(char buf[LOG_DATE_BUF_SIZE])
+{
+	using namespace std::chrono;
+
+	/*
+	 * NOTE: fmt of chrono::time_point with %S writes
+	 *       a floating point number of seconds if the
+	 *       duration resolution is less than second
```
MaxKellermann, line 93: Why do I care?
(Also: bad comment coding style)

src/LogBackend.cxx
```diff
@@ -62,28 +71,56 @@ SetLogThreshold(LogLevel _threshold) noexcept
...
-static const char *
-log_date(char buf[LOG_DATE_BUF_SIZE]) noexcept
-{
-	timespec t;
-	timespec_get(&t, TIME_UTC);
-	strftime(buf, LOG_DATE_BUF_SIZE, "%b %d %H:%M:%S : ", localtime(&t.tv_sec));
-	snprintf(&buf[LOG_DATE_STR_LEN], LOG_DATE_BUF_SIZE - LOG_DATE_STR_LEN, ".%03d : ", int(t.tv_nsec / 1e6));
-	return buf;
+	switch (_log_time_stamp) {
+		case LogTimestamp::MINUTES: log_time_formatter = log_date<std::chrono::minutes>; break;
```
MaxKellermann, line 119: Bad indent

src/LogBackend.cxx
```diff
@@ -62,28 +71,56 @@ SetLogThreshold(LogLevel _threshold) noexcept
...
-
-static const char *
-log_date(char buf[LOG_DATE_BUF_SIZE]) noexcept
-{
-	timespec t;
-	timespec_get(&t, TIME_UTC);
-	strftime(buf, LOG_DATE_BUF_SIZE, "%b %d %H:%M:%S : ", localtime(&t.tv_sec));
-	snprintf(&buf[LOG_DATE_STR_LEN], LOG_DATE_BUF_SIZE - LOG_DATE_STR_LEN, ".%03d : ", int(t.tv_nsec / 1e6));
-	return buf;
+	switch (_log_time_stamp) {
```
MaxKellermann, line 118: This `switch/case` block seems more complicated than it needs to be. You could better make this an array.

src/LogBackend.cxx
```diff
@@ -62,28 +71,56 @@ SetLogThreshold(LogLevel _threshold) noexcept
...
+	 *       duration resolution is less than second
+	 */
+	auto [p, n] = fmt::format_to_n(buf, LOG_DATE_BUF_SIZE, date_format_of<DurationType>,
+				       floor<DurationType>(system_clock::now()));
+	assert(n < LOG_DATE_BUF_SIZE);
+	*p = 0;
+	return buf;
+}
+
+static auto
```
MaxKellermann, line 102: Why `auto`? Having `auto` here and `-> const char*` later is more complicated than just writing `const char *` here.

src/LogBackend.cxx
```diff
@@ -62,28 +71,56 @@ SetLogThreshold(LogLevel _threshold) noexcept
...
+null_log_time_formatter = [](char buf[LOG_DATE_BUF_SIZE]) -> const char* {
+	buf[0] = 0;
+	return buf;
```
MaxKellermann, line 105: Too complicated. Why not `return "";` ?

geneticdrift:
OK. I don't know how to fix it all. So closing.
Thanks.

Issue #2165, opened 2024-12-13
about libnfs api v2 change


chenrui333:
👋 I saw there is [libnfs v6 API compatibility change](https://github.com/MusicPlayerDaemon/MPD/commit/31e583e9f8d14b9e67eab2581be8e21cd5712b47), but did not see it is part of the recent release, any reason why? Thanks!

relates to https://github.com/Homebrew/homebrew-core/pull/200785

heitbaum:
I tested ann update to Kodi last night. 
- https://github.com/LibreELEC/LibreELEC.tv/pull/9568
- https://github.com/sahlberg/libnfs/commit/5e8f7ce273308eb77f94248f4501e574a703c1a5
- https://github.com/LibreELEC/LibreELEC.tv/pull/9568/commits/5303bd9dcf6d8e5d507d2d9b965731501a629e62

and was looking at this this evening - as mpd is a LE addon.

```
[2/5] Compiling C++ object src/lib/nfs/libnfs.a.p/Connection.cxx.o
FAILED: src/lib/nfs/libnfs.a.p/Connection.cxx.o 
[...]

PR #2198, opened 2025-01-29
Minimal backport to support libnfs 6 on v0.23

BlackIkeEagle:
This is based on the following commits in master:
- https://github.com/MusicPlayerDaemon/MPD/commit/31e583e9f8d14b9e67eab2581be8e21cd5712b47
- https://github.com/MusicPlayerDaemon/MPD/commit/58e3b830e9a6826fef8950bccd7b173bf56de8b2


BlackIkeEagle:
Initially I did not want to even bother you with this because of https://github.com/MusicPlayerDaemon/MPD/issues/2165#issuecomment-2541941889. But the initial backport I took from homebrew failed hard, so thats where this PR comes from

loqs:
> This PR is mainly there to see if the backport looks fine to apply for use in Arch Linux for example. Usually "stable" distro's will also not yet have the libnfs v6 package in their repo's.

Was the intent to only raise the C++ standard if libnfs v6 is being used? As that currently is not the case.

@MaxKellermann if the use of std::span was removed so the C++ standard does not need to be increased you would then not be against this pull request on principal?

MaxKellermann:
I fear making C++20 conditional based on a library version check will be complicated. Does GCC/libstdc++ allow using std::span in C++17 mode? If yes, then nobody's MPD would break: people with libnfs<6 (with old GCC) see no change, and people who use libnfs6 need a GCC version that is recent enough.

If the code was changed to not use C++20 features, it could be merged. In MPD 0.23, we have the `ConstBuffer`/`WritableBuffer` template classes which is effectively the same concept (which were removed in 0.24 in favor of std::span).

loqs:
> Does GCC/libstdc++ allow using std::span in C++17 mode? 

Sadly no it produces undeclared errors as expected:
```
../mpd/src/lib/nfs/Connection.hxx:76:27: error: ‘std::span’ has not been declared
   76 |                           std::span<std::byte> dest
      |                           ^~~
../mpd/src/lib/nfs/Connection.hxx:76:36: error: expected ‘,’ or ‘...’ before ‘<’ token
   76 |                           std::span<std::byte> dest
      |                                    ^
../mpd/src/lib/nfs/Connection.hxx:204:19: error: ‘std::span’ has not been declared
  204 |                   std::span<std::byte> dest,
      |                   ^~~
../mpd/src/lib/nfs/Connection.hxx:204:28: error: expected ‘,’ or ‘...’ before ‘<’ token
  204 |                   std::span<std::byte> dest,
      |                            ^
``` 

> If the code was changed to not use C++20 features, it could be merged. In MPD 0.23, we have the `ConstBuffer`/`WritableBuffer` template classes which is effectively the same concept (which were removed in 0.24 in favor of std::span).

This is probably very wrong as it is only compile tested after converting to `WriteableBuffer`
```patch
diff --git a/src/lib/nfs/Connection.cxx b/src/lib/nfs/Connection.cxx
index c9d7985f7..71612500b 100644
--- a/src/lib/nfs/Connection.cxx
+++ b/src/lib/nfs/Connection.cxx
@@ -103,11 +103,25 @@ NfsConnection::CancellableCallback::Stat(nfs_context *ctx,
 
 inline void
 NfsConnection::CancellableCallback::Read(nfs_context *ctx, struct nfsfh *fh,
-					 uint64_t offset, size_t size)
+					 uint64_t offset,
+#ifdef LIBNFS_API_2
+					 WritableBuffer<std::byte> dest
+#else
+					 std::size_t size
+#endif
+	)
 {
 	assert(connection.GetEventLoop().IsInside());
 
-	int result = nfs_pread_async(ctx, fh, offset, size, Callback, this);
+	int result = nfs_pread_async(ctx, fh,
+#ifdef LIBNFS_API_2
+				     dest.data, dest.size,
+#endif
+				     offset,
+#ifndef LIBNFS_API_2
+				     size,
+#endif
+				     Callback, this);
 	if (result < 0)
 		throw FormatRuntimeError("nfs_pread_async() failed: %s",
 					 nfs_get_error(ctx));
@@ -329,7 +343,12 @@ NfsConnection::Stat(struct nfsfh *fh, NfsCallback &callback)
 }
 
 void
-NfsConnection::Read(struct nfsfh *fh, uint64_t offset, size_t size,
+NfsConnection::Read(struct nfsfh *fh, uint64_t offset,
+#ifdef LIBNFS_API_2
+		    WritableBuffer<std::byte> dest,
+#else
+		    std::size_t size,
+#endif
 		    NfsCallback &callback)
 {
 	assert(GetEventLoop().IsInside());
@@ -337,7 +356,13 @@ NfsConnection::Read(struct nfsfh *fh, uint64_t offset, size_t size,
 
 	auto &c = callbacks.Add(callback, *this, false);
 	try {
-		c.Read(context, fh, offset, size);
+		c.Read(context, fh, offset,
+#ifdef LIBNFS_API_2
+		       dest
+#else
+		       size
+#endif
+			);
 	} catch (...) {
 		callbacks.Remove(c);
 		throw;
diff --git a/src/lib/nfs/Connection.hxx b/src/lib/nfs/Connection.hxx
index 49987e2c3..4c5cb9c05 100644
--- a/src/lib/nfs/Connection.hxx
+++ b/src/lib/nfs/Connection.hxx
@@ -24,6 +24,7 @@
 #include "event/SocketEvent.hxx"
 #include "event/CoarseTimerEvent.hxx"
 #include "event/DeferEvent.hxx"
+#include "util/WritableBuffer.hxx"
 
 #include <string>
 #include <list>
@@ -71,7 +72,13 @@ class NfsConnection {
 		void Open(nfs_context *context, const char *path, int flags);
 		void Stat(nfs_context *context, struct nfsfh *fh);
 		void Read(nfs_context *context, struct nfsfh *fh,
-			  uint64_t offset, size_t size);
+			  uint64_t offset,
+#ifdef LIBNFS_API_2
+			  WritableBuffer<std::byte> dest
+#else
+			  std::size_t size
+#endif
+			);
 
 		/**
 		 * Cancel the operation and schedule a call to
@@ -193,7 +200,12 @@ public:
 	/**
 	 * Throws std::runtime_error on error.
 	 */
-	void Read(struct nfsfh *fh, uint64_t offset, size_t size,
+	void Read(struct nfsfh *fh, uint64_t offset,
+#ifdef LIBNFS_API_2
+		  WritableBuffer<std::byte> dest,
+#else
+		  std::size_t size,
+#endif
 		  NfsCallback &callback);
 
 	void Cancel(NfsCallback &callback) noexcept;
diff --git a/src/lib/nfs/FileReader.cxx b/src/lib/nfs/FileReader.cxx
index 6e9457d79..6bd75e9ea 100644
--- a/src/lib/nfs/FileReader.cxx
+++ b/src/lib/nfs/FileReader.cxx
@@ -129,7 +129,15 @@ NfsFileReader::Read(uint64_t offset, size_t size)
 {
 	assert(state == State::IDLE);
 
+#ifdef LIBNFS_API_2
+	assert(!read_buffer);
+	// TOOD read into caller-provided buffer
+	read_buffer = std::make_unique<std::byte[]>(size);
+	connection->Read(fh, offset, {read_buffer.get(), size}, *this);
+#else
 	connection->Read(fh, offset, size, *this);
+#endif
+
 	state = State::READ;
 }
 
@@ -137,7 +145,12 @@ void
 NfsFileReader::CancelRead() noexcept
 {
 	if (state == State::READ) {
+#ifdef LIBNFS_API_2
+		assert(read_buffer);
+		read_buffer.release();
+#endif
 		connection->Cancel(*this);
+
 		state = State::IDLE;
 	}
 }
diff --git a/src/lib/nfs/FileReader.hxx b/src/lib/nfs/FileReader.hxx
index 8d257efdd..a2b9e9c60 100644
--- a/src/lib/nfs/FileReader.hxx
+++ b/src/lib/nfs/FileReader.hxx
@@ -32,6 +32,10 @@
 
 #include <sys/stat.h>
 
+#ifdef LIBNFS_API_2
+#include <memory>
+#endif
+
 struct nfsfh;
 class NfsConnection;
 
@@ -68,6 +72,10 @@ class NfsFileReader : NfsLease, NfsCallback {
 	 */
 	InjectEvent defer_open;
 
+#ifdef LIBNFS_API_2
+	std::unique_ptr<std::byte[]> read_buffer;
+#endif
+
 public:
 	NfsFileReader() noexcept;
 	~NfsFileReader() noexcept;
diff --git a/src/lib/nfs/meson.build b/src/lib/nfs/meson.build
index 274ee7b47..586a71391 100644
--- a/src/lib/nfs/meson.build
+++ b/src/lib/nfs/meson.build
@@ -1,9 +1,16 @@
-nfs_dep = dependency('libnfs', version: ['>= 4', '< 6'], required: get_option('nfs'))
+nfs_dep = dependency('libnfs', version: '>= 4', required: get_option('nfs'))
 conf.set('ENABLE_NFS', nfs_dep.found())
 if not nfs_dep.found()
   subdir_done()
 endif
 
+if nfs_dep.version().version_compare('>=6')
+  # libnfs has no version macro therefore we must detect the API
+  # version 2 at configure time
+  nfs_dep = declare_dependency(compile_args: '-DLIBNFS_API_2',
+                               dependencies: nfs_dep)
+endif
+
 nfs = static_library(
   'nfs',
   'Connection.cxx',
```

Issue #1879, opened 2023-10-10
OSXAudio causes SEGV only when daemonized 


Sean-Der:
### Describe the bug
When launching as a daemon MPD crashes. 

```
0   libsystem_trace.dylib         	       0x18d2e323c _os_log_preferences_refresh + 68
1   libsystem_trace.dylib         	       0x18d2e3c9c os_log_type_enabled + 712
2   CoreFoundation                	       0x18d646728 _CFBundleDlfcnCheckLoaded + 172
3   CoreFoundation                	       0x18d6f1d0c _CFBundleLoadExecutableAndReturnError + 136
4   AudioToolboxCore              	       0x18f75ff8c APComponent_FromBundle_Loadable::ResolveFactoryFunction() + 68
5   AudioToolboxCore              	       0x18f75ef6c APComponent::newInstance(unsigned int, bool, void (OpaqueAudioComponentInstance*, int) block_pointer) + 324
6   AudioToolboxCore              	       0x18f8942fc instantiate(OpaqueAudioComponent*, unsigned int, bool, void (OpaqueAudioComponentInstance*, int) block_pointer) + 428
7   AudioToolboxCore              	       0x18f89481c __AudioComponentInstanceNew_block_invoke + 120
8   AudioToolboxCore              	       0x18f6d1058 Synchronously + 132
9   AudioToolboxCore              	       0x18f8945a0 AudioComponentInstanceNew + 248
10  mpd                           	       0x1048fbeb8 OSXOutput::Enable() + 112
11  mpd                           	       0x1048eeae8 FilteredAudioOutput::Enable() + 36
12  mpd                           	       0x1048f1508 AudioOutputControl::InternalEnable() + 88
13  mpd                           	       0x1048f117c AudioOutputControl::Task() + 228
14  mpd                           	       0x1048cd9dc Thread::ThreadProc(void*) + 32
15  libsystem_pthread.dylib       	       0x18d57f034 _pthread_start + 136
16  libsystem_pthread.dylib       	       0x18d579e3c thread_start + 8
```

If I launch with `mpd --no-daemon` it works.

I am running macOS Sonoma. I can't think of anything else unique about my setup yet. This didn't happen before.

## Expected Behavior
Playback

## Actual Behavior
SEGV

## Version
```
Music Player Daemon 0.23.14 (0.23.14)
Copyright 2003-2007 Warren Dukes <warren.dukes@gmail.com>
[...]

Issue #1901, opened 2023-11-16
Daemon mode does not work with MacOS


laiti:
### Describe the bug
Daemon mode does not work with MacOS. Starting mpd without `--no-daemon` makes it quit immediately without any error messages.

## Expected Behavior
Starting mpd without `--no-daemon` should make it run in background.

## Actual Behavior
Starting mpd without `--no-daemon` makes mpd shut down immediately without any errors in stderr or logs even when `--verbose` is set. There is no mpd in the process list.

## Version
$ uname -a
Darwin nark 23.1.0 Darwin Kernel Version 23.1.0: Mon Oct  9 21:27:24 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T6000 arm64

MacOS version is Sonoma 14.1.

$ mpd --version
Music Player Daemon 0.23.14 (0.23.14)
Copyright 2003-2007 Warren Dukes <warren.dukes@gmail.com>
Copyright 2008-2021 Max Kellermann <max.kellermann@gmail.com>
This is free software; see the source for copying conditions.  There is NO
warranty; not even MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Database plugins:
 simple proxy upnp

Storage plugins:
 local nfs curl

Neighbor plugins:
 upnp

Decoders plugins:
 [mpg123] mp3
 [vorbis] ogg oga
 [oggflac] ogg oga
 [flac] flac
 [opus] opus ogg oga
 [sndfile] wav aiff aif au snd paf iff svx sf voc w64 pvf xi htk caf sd2
 [dsdiff] dff
 [dsf] dsf
 [hybrid_dsd] m4a
 [faad] aac
 [wavpack] wv
 [fluidsynth] mid
[...]

PR #2201, opened 2025-01-30
macOS support and fixes

CamilleScholtz:
I saw that macOS will be dropped in the next version of mpd because there are no maintainers for it. Considering I currently work on an mpd client for macOS, and in addition to that, I am a very happy macOS user of mpd, I was quite disappointed!

I only have basic knowledge of cpp, but I am willing to maintain macOS specific code for mpd. This PR, which might be extended in the coming days, currently:

- fixes the macOS GitHub workflow (it builds now!)
- fixes issue #1879 / #1901. The problem here seems to be that when `fork()` is called, macOS's Objective-C runtime may detect that a class initialization was in progress on another thread. This triggers a crash to prevent potential deadlocks or data corruption, as fork() in a multi-threaded context is unsafe. The solution is to call `fork()` before initializing any macOS frameworks or threads, I've done this for now by creating an apple specific main (like how there is a win32 specific main), that handles forking at a very early stage. An alternative way would be to launch mpd with the `OBJC_DISABLE_INITIALIZE_FORK_SAFETY` environment variable.

As I've said before, my knowledge of cpp is quite basic, so if you see any room for improvement or if I'm doing things wrong, I'd love to hear it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,6 +150,56 @@ jobs:
       - name: Unit Tests Mini
         run: meson test -C output/mini
 
+  build-macos:
+    runs-on: macos-latest
+    steps:
+      - id: checkout
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.x
+
+      - name: Install dependencies
+        run: |
+          brew install \
+            meson ninja \
+            ccache \
+            fmt \
+            googletest \
+            icu4c \
+            ffmpeg \
+            libnfs \
+            libupnp \
+            libid3tag \
+            chromaprint \
+            libsamplerate \
+            libsoxr \
+            flac \
+            opus \
+            libvorbis \
+            faad2 \
+            wavpack \
+            libmpdclient
+
+      - id: cache-ccache
+        uses: hendrikmuhs/ccache-action@v1
+        with:
+          key: macos
+
+      - name: Configure
+        run: |
+          meson setup \
+            -Ddocumentation=disabled \
+            -Dtest=true \
+            output
+
+      - name: Build
+        run: meson compile -C output --verbose
+
+      - name: Unit Tests
+        run: meson test -C output
+
   build-msys2:
     strategy:
       matrix:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -161,7 +161,9 @@ jobs:
             cxx: clang++
             brew_packages: >
               meson ninja
+              pkgconf
               ccache
+              expat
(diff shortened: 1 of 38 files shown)


MaxKellermann:
Hey, thanks for stepping in. I'll check your changes another day, but please don't put merge commits in a PR branch. If you want to update your PR branch with the latest upstream changes, rebase instead of merge.

.gitignore
```diff
@@ -9,3 +9,5 @@ __pycache__/
 
 /.clangd/
 /compile_commands.json
+
+.DS_store
```
MaxKellermann, line 13: No explanatio given in the commit message. I don't know what this is. If this is a temporary directory of some application you're using, it should probably be better in your personal global gitignore.

.github/workflows/build.yml
```diff
@@ -161,15 +161,11 @@ jobs:
             cxx: clang++
             brew_packages: >
               meson ninja
-              pkgconf
               ccache
               expat
               fmt
               googletest
-              pcre2
```
MaxKellermann: What's the rationale for removing all these packages? This means MPD will build without these features, and these features will not be verified by the CI.

CamilleScholtz: The macos-latest workflow ships with all of these packages preinstalled, except for libnfs.

.gitignore
CamilleScholtz: It's a file macOS creates when opening a directory in finder, I agree it should be in my personal gitignore.

.github/workflows/build.yml
MaxKellermann: I added them because I thought it doesn't hurt to mention these, even if they are preinstalled; if some future image happens to remove them, we're still safe.

MaxKellermann:
Your last commit is a revert commit. That means you suggest to merge a commit plus a revert of that commit. That does not make much sense, I'd rather have the commit removed from the PR and only merge known-good commits.
I cherry-picked the commits manually, omitting the gitignore commit, the redundant dependency removal commit and the merge commit.
Thanks so far, hope to get more code from you!

CamilleScholtz:
Thanks, commit messages and PR was a bit of a mess; I was thinking about squashing things or recreating a new proper PR, but that's not needed anymore! I'll keep my eye out for osx issues.

Badsheep:
Hello @CamilleScholtz,

I'm also a mac user and a very happy user of mpd...
I just supgraded to Sonoma and I have the daemon bug.
I was wandering : did you do a binary of mpd with your fixes ? I'd be very happy and lucky to use it :)
Thank you

PR #2205, opened 2025-02-01
doc/user: update windows build meson version

allixx:
Since 0.23.17 it's no longer possible to cross compile for Windows on Linux using win32/build.py with meson 0.56.0 due to subprojects like id3tag utilizing meson features not available in 0.56.0.

### DIFF
--- a/doc/user.rst
+++ b/doc/user.rst
@@ -160,7 +160,7 @@ This section is about the latter.
 You need:
 
 * `mingw-w64 <http://mingw-w64.org/doku.php>`__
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Curious question: why is there text in this PR but not in the commit message?

allixx:
I suppose this pull request was more like a gentle nudge to push meson version requirements up all around the v0.23.x branch, including docs and project's meson.build. I just wasn't sure what version would've been appropriate, therefore half baked commit.

PR #2208, opened 2025-02-03
Fix macOS build with nfs enabled

CamilleScholtz:
I don't really know why this does compile on Linux but not on macOS, but regardless, this fixes it. Needed for the `nfsc/libnfs-raw-nfs.h` include in `input/Error.cxx`.


MaxKellermann:
Next time, please write explanatory text in commit messages, not in the GitHub PR. PR texts are lost in GitHub's proprietary database.

PR #2213, opened 2025-02-21
config/File: support resetting repeatable params

jspricke:
This allows resetting bind_to_address to override the default value in a
included config.


jspricke:
Note that this is the same feature systemd and others use.

Issue #476, opened 2019-02-03
Support XDG Base Directories


shibumi:
Hi,
It would be great if you could support `XDG Base Directories` in your `mpd.conf` file.
I have the following configuration:
```
pid_file "$XDG_RUNTIME_DIR/1000/mpd/pid"
db_file "$XDG_DATA_DIR/mpd/mpd.db"
state_file "$XDG_DATA_DIR/mpd/mpdstate"
playlist_directory "$XDG_DATA_DIR/mpd/playlists"
music_directory "$HOME/rips/"
```

but when I try to use it I get:
```
exception: Invalid path at line 3:
exception: nested: not an absolute path: $XDG_RUNTIME_DIR/1000/mpd/pid
```

MaxKellermann:
Do you want MPD to expand environment variable references, or do you want MPD to support XDG?

shibumi:
It should expand XDG environment variables

Issue #263, opened 2018-04-09
Environment Variables in config file


ForTheReallys:
Would you consider adding the expansion of environment variables in mpv.conf?  I would like mpd to listen on $XDG_RUNTIME_DIR/mpd-socket but cannot do this without expanding the variable by hand.  
Thank you!

ttytyper:
MPD listens on `$XDG_RUNTIME_DIR/mpd/socket` if no `bind_to_address` is set. So maybe expansion of `$XDG_RUNTIME_DIR` is no longer needed?

Source: https://www.musicpd.org/doc/html/user.html#listeners

Works with MPD 0.21.5 on Debian buster. Commented out all `bind_to_address` lines in my `~/.mpdconf`, killed and restarted mpd. Now I can connect to it using the socket:
```
$ MPD_HOST="$XDG_RUNTIME_DIR/mpd/socket" mpc
volume: n/a   repeat: off   random: off   single: off   consume: off
```

However, even when `bind_to_address` is unset, mpd still listens on "[::]:6600" (world reachable) in addition to the XDG socket. If you _do_ set `bind_to_address` to e.g. localhost, the XDG socket is gone. Tilde is expanded, so you can set `bind_to_address` to e.g. `~/.mpd/socket` but that kinda defeats the point of this issue. You can still control which port mpd listens on, using the `port` setting, so at least you can avoid port conflicts.

meliurwen:
Another use case is `music_directory` with `$XDG_MUSIC_DIR`.

If we want to reach maximum versatility of the config file probably [Parameter Expansion](https://pubs.opengroup.org/onlinepubs/9699919799.2016edition/utilities/V3_chap02.html#tag_18_06_02) with subshell support would be ideal; I have zero knowledge of this codebase but I suspect that this would be not so straightforward to implement.

In my case:
```conf
music_directory  "${XDG_MUSIC_DIR:-~/Music}"

...

bind_to_address  "${XDG_RUNTIME_DIR:-/tmp/$(id -u)-runtime}/mpd/main.sock"
```

ifohancroft:
I'm for this as well. I'd like to be able to set the playlist directory or the log file directory using $XDG_CONFIG_HOME or $XDG_CACHE_HOME or $XDG_DATA_HOME, etc.. you get the idea.

nickjj:
[...]

PR #2220, opened 2025-02-28
Initial support for $HOME and some XDG variables inside the configuration file

JoanVC100:
This pull request tries to address issues #263 and #476.

With these changes you can use $HOME and some of the $XDG variables directly in the MPD configuration file. I used the functions in "fs/glue/StandardDirectory.hxx" to extract the XDG variables, so currently it's limited to:
- XDG_CONFIG_HOME
- XDG_MUSIC_DIR
- XDG_CACHE_HOME
- XDG_RUNTIME_DIR

### DIFF
--- a/doc/mpd.conf.5.rst
+++ b/doc/mpd.conf.5.rst
@@ -17,9 +17,18 @@ Each line in the configuration file contains a setting name and its value, e.g.:
 
 Lines starting with ``#`` are treated as comments and ignored.
 
-For settings which specify a filesystem path, the tilde is expanded:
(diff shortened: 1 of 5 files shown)


MaxKellermann:
Your commit message is misleading - this is not about environment variables.
Also documentation is missing.

JoanVC100:
Sorry for the premature closure, I am still learning about GitHub pull requests.

Anyway, I rewrote the commit message to make it more accurate. And I changed the documentation in both the man page and the HTML, as well as the example mpd.conf.

MaxKellermann:
Thanks, that's good!

PR #2227, opened 2025-03-11
src/fs: Enable USE_XDG support on macOS

CamilleScholtz:
Since macOS does not have a `user-dirs.dirs` file, update GetUserDir to use the corresponding XDG environment variable via GetExistingEnvDirectory when compiled on __APPLE__. This change provides a macOS-specific fallbacks.

### DIFF
--- a/src/fs/XDG.hxx
+++ b/src/fs/XDG.hxx
@@ -5,7 +5,7 @@
 #define MPD_FS_XDG_HXX
 
 // Use X Desktop guidelines where applicable
-#if !defined(__APPLE__) && !defined(_WIN32) && !defined(ANDROID)
+#if !defined(_WIN32) && !defined(ANDROID)
 #define USE_XDG
 #endif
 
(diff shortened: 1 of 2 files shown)


MaxKellermann:
If you find a mistake in an unmerged commit, please amend the existing commit instead of adding a fixup commit to the PR branch.

I don't think the approach in this PR is good - you pretend to have XDG support, but this is fake. Instead of adding stuff to `GetUserDir()`, implementing several variable names as special cases, you should add these cases to `GetUserMusicDir()` etc., just like we do it for Android and Windows - these don't have XDG support, but they have a system-defined music directory.

I think you mean to provide support for the new `$XDG_...` variables, don't you? For that, we could remove the `#ifdef USE_XDG` from `GetVariable()`, because these don't really use XDG functions (just one of the implementations is XDG).

CamilleScholtz:
I agree, my motivation was indeed support for the `$XDG...` variables. I'll rework this and open a new PR.

PR #2228, opened 2025-03-12
src/fs: Support use of GetVariable on macOS

CamilleScholtz:
This will allow for the use of `$XDG...` variables in the config file on macOS. Checks for directories based on the environment variable first and falls back to some macOS specific fallback directories when needed.

### DIFF
--- a/src/fs/glue/StandardDirectory.cxx
+++ b/src/fs/glue/StandardDirectory.cxx
@@ -23,7 +23,6 @@
 #endif
 
 #ifdef USE_XDG
-#include "util/StringSplit.hxx"
 #include "util/StringStrip.hxx"
 #include "util/StringCompare.hxx"
 #include "io/FileLineReader.hxx"
@@ -39,7 +38,7 @@
 #include "Main.hxx"
 #endif
 
-#ifdef USE_XDG
+#if defined(USE_XDG) || defined(__APPLE__)
 #include "Version.h" // for PACKAGE_NAME
 #define APP_FILENAME PATH_LITERAL(PACKAGE_NAME)
 static constexpr Path app_filename = Path::FromFS(APP_FILENAME);
@@ -248,13 +247,11 @@ GetUserConfigDir() noexcept
 {
 #if defined(_WIN32)
 	return GetStandardDir(CSIDL_LOCAL_APPDATA);
-#elif defined(USE_XDG)
-	// Check for $XDG_CONFIG_HOME
+#elif defined(USE_XDG) || defined(__APPLE__)
 	if (const auto path = GetExistingEnvDirectory("XDG_CONFIG_HOME");
 	    path != nullptr)
 		return AllocatedPath{path};
 
-	// Check for $HOME/.config
 	if (const auto home = GetHomeDir(); !home.IsNull()) {
 		auto fallback = home / Path::FromFS(".config");
 		if (IsValidDir(fallback))
@@ -274,6 +271,18 @@ GetUserMusicDir() noexcept
 	return GetStandardDir(CSIDL_MYMUSIC);
 #elif defined(USE_XDG)
 	return GetUserDir("XDG_MUSIC_DIR");
+#elif defined(__APPLE__)
+	if (const auto path = GetExistingEnvDirectory("XDG_MUSIC_DIR");
+	    path != nullptr)
+		return AllocatedPath{path};
+
+	if (const auto home = GetHomeDir(); !home.IsNull()) {
+		auto fallback = home / Path::FromFS("Music");
+		if (IsValidDir(fallback))
+			return fallback;
+	}
+
+	return nullptr;
 #elif defined(ANDROID)
 	return Environment::getExternalStoragePublicDirectory(Java::GetEnv(),
 							      "Music");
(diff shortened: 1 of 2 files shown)


MaxKellermann:
Does it make sense to support XDG environment variables on macOS? Does anybody else do this, or are you suggesting MPD does something new here?

Your commit mixes at least two different things:
1. allowing XDG expansion on non-Linux (some of these functions already have Windows/Android implementations, so this would be a useful feature for OSes other than macOS already)
2. implementing some of these functions for macOS; with two implementations: XDG environment variables (which I questioned) and hard-coded home sub directories (which I guess is a macOS standard?)

src/fs/glue/StandardDirectory.cxx
```diff
@@ -23,7 +23,6 @@
 #endif
 
 #ifdef USE_XDG
-#include "util/StringSplit.hxx"
```
MaxKellermann, line 26: Why remove this?

src/fs/glue/StandardDirectory.cxx
```diff
@@ -274,6 +271,18 @@ GetUserMusicDir() noexcept
 	return GetStandardDir(CSIDL_MYMUSIC);
 #elif defined(USE_XDG)
 	return GetUserDir("XDG_MUSIC_DIR");
+#elif defined(__APPLE__)
+	if (const auto path = GetExistingEnvDirectory("XDG_MUSIC_DIR");
```
MaxKellermann, line 275: This pretends it implements XDG on macOS, but it doesn't actually; it only implements the environment variable, so it copies only half of the code. You shouldn't copy code at all; instead, either implement it fully (by calling existing code) or not at all.

MaxKellermann:
If you believe XDG has a place on macOS, I trust your decision, but I believe your implementation is flawed.
When you fix this, please also split this into two commits, one for each step.

src/fs/glue/StandardDirectory.cxx
CamilleScholtz: But that is what I did in the previous PR? The whole point of this PR is to support said environment variable only. I could create another environment variable, something like `MPD_MUSIC_DIR`, but I think that would harm cross compatibility; you now have to think about both `MPD_MUSIC_DIR` or `XDG_MUSIC_DIR` depending on the system.

src/fs/glue/StandardDirectory.cxx
MaxKellermann: It is confusing to hide such a change in an unrelated commit.

src/fs/glue/StandardDirectory.cxx
MaxKellermann: So you explicitly do *not* want to implement the XDG base directory spec, but *only* some of the environment variables?

MaxKellermann: I mean XDG base directory + XDG user-dirs.

CamilleScholtz: Isn't that what you suggested in my previous PR:

>I don't think the approach in this PR is good - you pretend to have XDG support, but this is fake. Instead of adding stuff to GetUserDir(), implementing several variable names as special cases, you should add these cases to GetUserMusicDir() etc., just like we do it for Android and Windows - these don't have XDG support, but they have a system-defined music directory.

>I think you mean to provide support for the new $XDG_... variables, don't you? For that, we could remove the #ifdef USE_XDG from GetVariable(), because these don't really use XDG functions (just one of the implementations is XDG).

MaxKellermann: No, Idid not suggest to implement the XDG base directory / user-dirs on macOS. In the first comment here, I even questioned whether that would eben be useful on macOS, and you said yes, it is useful.

I only suggested removing the `#ifdef` and implementing the underlying functions like `GetUserMusicDir()` on macOS.
You did that (having mixed three steps in one commit), but additionally you implemented some of the XDG environment variables, without however considering the xdg-user-dirs. You could have had both by just calling the existing code, but instead of duplicated half of the existing code.

CamilleScholtz: I might be misunderstanding something here, but I did the following:

>For that, we could remove the #ifdef USE_XDG from GetVariable()

I did that, but without actually implementing the environment variables in the functions `GetVariable()` calls, the results would always just be `nullptr`' for any of the variables besides `HOME`.

I basically just want to set an `XDG_...` environment variable and make MPD respect that, like dozens of other programs do on macOS. Or is your idea that only on Linux the directories can be modified using an environment variable and on other cases there are only hardcoded predefined locations? In that case why even suggest removing the `#ifdef` from `GetVariable()` in the first place?

MaxKellermann: > I did that, but without actually implementing the environment variables in the functions `GetVariable()` calls, the results would always just be `nullptr`' for any of the variables besides `HOME`.

True, that's because my suggestion consisted of two parts, of which you cited only the first.

> I basically just want to set an `XDG_...` environment variable and make MPD respect that, like dozens of other program on macOS do.

Such code already exists in MPD, but you duplicated (part of) that code. Duplicating code isn't good.

> Or is your idea that only on Linux the locations can be modified using an environment variable and on other systems there are only hardcoded predefined locations?

If an operating system has a canonical location for music files, `GetUserMusicDir()` should return that. This is the function that provides the default `music_directory` value, if the configuration file doesn't contain one. And additionally (since 0.24), `$XDG_MUSIC_DIR` expands to this value.

If you decide (unrelated to what I wrote above) that macOS should follow the XDG standard, then you can use the existing XDG code on macOS. But not reimplement it. Yet you did reimplement it, in a way that omits features that MPD has on other operating systems.

That is bad for two reasons:
1. duplicate code
2. inconsistent feature set across operating systems, leading to confused users

> In that case why even suggest removing the `#ifdef` from `GetVariable()` in the first place?

Because my suggestion had a second part.

Issue #2232, opened 2025-03-14
0.24: segfault when last song in playlist finished


dvzrv:
### Describe the bug

When using ncmpcpp after playback of an album and trying to access a directory in the artist browser, mpd segfaulted.
(I am not sure if it crashed already at the end of playback or during browsing using the artist browser).
I am using the httpd plugin, so maybe this is similar to #2176.

I've attached a gdb backtrace at the end.

```bash
coredumpctl info
           PID: 176923 (mpd)
           UID: 1000 (dave)
           GID: 1000 (dave)
        Signal: 11 (SEGV)
     Timestamp: Fri 2025-03-14 11:37:27 CET (30min ago)
  Command Line: /usr/bin/mpd --no-daemon /home/dave/.config/mpd/hmbx-pipewire.conf
    Executable: /usr/bin/mpd
 Control Group: /user.slice/user-1000.slice/user@1000.service/app.slice/app-mpd.slice/mpd@hmbx-pipewire.service
          Unit: user@1000.service
     User Unit: mpd@hmbx-pipewire.service
         Slice: user-1000.slice
     Owner UID: 1000 (dave)
       Boot ID: 765b91fdc9a848e2aeb8c537e5b10634
    Machine ID: efe2daa6378d41c0a5c311fb29c5b583
      Hostname: hmbx
       Storage: /var/lib/systemd/coredump/core.mpd.1000.765b91fdc9a848e2aeb8c537e5b10634.176923.1741948647000000.zst (present)
  Size on Disk: 10.2M
       Message: Process 176923 (mpd) of user 1000 dumped core.

                Stack trace of thread 177012:
                #0  0x00005d75d99a53a4 n/a (n/a + 0x0)
                #1  0x00005d75d9996b3d n/a (n/a + 0x0)
                #2  0x00005d75d9967861 n/a (n/a + 0x0)
                #3  0x00007ada5dfab70a n/a (n/a + 0x0)
                #4  0x00007ada5e02f8b4 n/a (n/a + 0x0)
                ELF object binary architecture: AMD x86-64
```

## Version

```
Music Player Daemon 0.24 (v0.24)
Copyright 2003-2007 Warren Dukes <warren.dukes@gmail.com>
Copyright 2008-2021 Max Kellermann <max.kellermann@gmail.com>
This is free software; see the source for copying conditions.  There is NO
warranty; not even MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

[...]

PR #2254, opened 2025-03-24
Fix crash in TwoFilters::Reset due to missing null check

Kibodi:
When there is only one song in the playlist, consume mode is enabled and the song ends, MPD crashes. This fixes it.

### DIFF
--- a/src/filter/plugins/TwoFilters.cxx
+++ b/src/filter/plugins/TwoFilters.cxx
@@ -10,6 +10,9 @@
 std::span<const std::byte>
 TwoFilters::FilterPCM(std::span<const std::byte> src)
 {
+	if (!first)
+		throw std::runtime_error("First filter is null.");
+
 	if (const auto dest = first->FilterPCM(src); dest.empty()) [[unlikely]]
 		/* no output from the first filter; pass the empty
                    buffer on, do not call the second filter */
--- a/src/filter/plugins/TwoFilters.hxx
+++ b/src/filter/plugins/TwoFilters.hxx
@@ -24,7 +24,8 @@ public:
 		 second(std::forward<S>(_second)) {}
 
 	void Reset() noexcept override {
-		first->Reset();
+		if (first)
+			first->Reset();
 		second->Reset();
 	}
 


MaxKellermann:
Bad indentation, unnecessary braces and the commit message should explain the bug being fixed. There is no explanation at all currently.

MaxKellermann:
Please amend known-bad commits instead of adding fixup commits.

MaxKellermann:
Why does MPD crash? If you know why, the commit message should explain it.

MaxKellermann:
> If the first filter is null

But under what circumstances is the first filter null? How is that possible?

Kibodi:
You're right, the second filter never becomes null.
However, the first filter is set null when the last song in a playlist finishes and no further tracks are available. This happens, for example, when playing a full album or a single track. If playback reaches the end of the playlist (or you manually skip to the end), `first` becomes null.
My change in [TwoFilters.hxx](https://github.com/MusicPlayerDaemon/MPD/pull/2254/commits/8b45628d3f88bd552e65c660f997137a6832913f#diff-2fb8c8076f0b8d4e9b3cb80506b3fff9a04e2ed6047d3bb6bf510132ddebe6ce) ensures that `Reset()` only calls `first->Reset()` if first is not null. However, I noticed that `first` could also be null when selecting a new song after reaching the end of a playlist, which caused a crash. This is what my fix in [TwoFilters.cxx](https://github.com/MusicPlayerDaemon/MPD/pull/2254/commits/8b45628d3f88bd552e65c660f997137a6832913f#diff-9eadb9dd21a6d9d2c5c2b91348eb096d6145351d5f6c4441502e93ef65bb474e) addresses.
While this may not be the ultimate solution, it prevents MPD from crashing when reaching the end of the playlist.

MaxKellermann:
I had a closer look and found that while your code change works around the crash #2232 crash, it's not correct. The problem is when due to a race TwoFilters::Reset() gets called after TwoFilters::Flush(), but according to the Filter API documentation, doing this call sequence is illegal. Thanks anyway for your effort, it was useful input, even if it wasn't the proper solution.

MaxKellermann:
https://github.com/MusicPlayerDaemon/MPD/commit/4ed86148234e71db08fbdbd009cd83fdc7adbc0f is the proper fix for the bug.

PR #2257, opened 2025-03-24
macOS: Fix command parsing when using options with daemon mode

CamilleScholtz:
This fixes an issue where running MPD with command line options on macOS in daemon mode would cause command parsing exceptions. The problem occurred because ParseCommandLine was called twice (once in apple_main and once in mpd_main), which altered the argv array.

Solution:
1. Modify apple_main to call `MainConfigured()` with already parsed options
2. Added `MainConfigured` to the header file

This maintains the early fork for macOS (necessary to avoid Objective-C runtime issues) while ensuring command line arguments are only parsed once.

Fixes: #2235


src/apple/AppleMain.cxx
```diff
@@ -32,7 +32,9 @@ int apple_main(int argc, char *argv[])
 			_exit(0);
 		}
 	}
-#endif
 
+	MainConfigured(options, raw_config);
```
MaxKellermann, line 36: Return value missing - does this even compile?

CamilleScholtz: It does, check the tasks

MaxKellermann:
I merged this to v0.24.x with a few fixups (bad indentation, bad commit message).
Now can we please remove this AppleMain.cxx completely?

PR #2258, opened 2025-03-25
Main: Remove apple_main

CamilleScholtz:
Turns out things could all be done in `MainConfigured` 😅.


MaxKellermann:
The most important part of removing apple_main would be to remove the duplicate (and badly implemented) daemonization code. This is missing here.

CamilleScholtz:
I lack the skills to do that, from what I've read the proper solution would be to use `posix_spawn()` instead of `fork()`, but I don't really know enough about the inner workings of either to know what that would entail.

MaxKellermann:
I have cherry-picked to v0.24.x: e668a72ec0872e5c551612547c1ffc973a4e0cfa

PR #2260, opened 2025-03-30
net/DscpParser: fix building dscp parser code on FreeBSD / OpenBSD.

brad0:
```
/usr/include/netinet/ip.h:67:19: error: field has incomplete type 'struct in_addr'
        struct    in_addr ip_src, ip_dst; /* source and dest address */
                          ^
```

### DIFF
--- a/src/net/DscpParser.cxx
+++ b/src/net/DscpParser.cxx
@@ -9,6 +9,7 @@
 #ifdef _WIN32
 #include <ws2tcpip.h>
 #else
+#include <netinet/in.h>
 #include <netinet/ip.h>
 #endif
 


MaxKellermann:
This awfully looks like broken FreeBSD system headers. If `/usr/include/netinet/ip.h` depends on the definition of `struct in_addr`, then `/usr/include/netinet/ip.h` must include a header that defines it. Convince me that this is a MPD problem.

MaxKellermann:
See https://github.com/MusicPlayerDaemon/MPD/pull/2261#discussion_r2020275680 for more FreeBSD bugs in the same header.

MaxKellermann:
Bug was reported to the FreeBSD project: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=285924
I suggest you join the discussion over there to get this fixed at the root cause.

PR #2269, opened 2025-04-07
Use redundant call to io_uring_queue_init()

aeolio:
Commit e309941 [1] introduced IORING_SETUP_SINGLE_ISSUER in the call to io_uring_setup(). This flag is available since kernel version 6.0 [2]. Within the system call, the kernel checks supplied against defined flags, and returns -EINVAL if extra flags were supplied.

The initialization in io/uring/Ring.cxx fails with a low priority warning, that is not even printed in the log file (using default settings) and continues without io_uring support.

The Linux kernel supports io_uring since version 5.1, MPD currently has a minimum kernel requirement of 5.6. Ideally this feature should work with all supported kernels.

Use redundant call to io_uring_queue_init(): If the first function call fails, repeat with the flags masked to the allowed range (0x3Fu) of kernel version 5.6 (the current MPD minimum requirement).

[1] https://github.com/MusicPlayerDaemon/MPD/commit/e309941
[2] https://www.man7.org/linux/man-pages/man2/io_uring_setup.2.html

### DIFF
--- a/doc/developer.rst
+++ b/doc/developer.rst
@@ -91,13 +91,13 @@ Always write your code against the latest git:
 
 .. code-block:: none
 
-    git clone git://github.com/MusicPlayerDaemon/MPD
+    git clone https://github.com/MusicPlayerDaemon/MPD.git
 
 If you already have a clone, update it:
 
 .. code-block:: none
 
-    git pull --rebase git://github.com/MusicPlayerDaemon/MPD master
+    git pull --rebase https://github.com/MusicPlayerDaemon/MPD.git master
 
 You can do without :code:`--rebase`, but we recommend that you rebase
 your repository on the "master" repository all the time.


aeolio:
Something like this (needs to be tested): 

```
if (int error = io_uring_queue_init(entries, &ring, flags);
    error < 0)
	std_flags = 0x3Fu; // minimum supported since kernel 5.6
	if (error = -EINVAL && flags & ~std_flags) {
	    flags &= std_flags;
		if (int error = io_uring_queue_init(entries, &ring, flags);
		    error < 0)
			throw MakeErrno(-error, "io_uring_queue_init() failed (fallback)");
		}
	else
		throw MakeErrno(-error, "io_uring_queue_init() failed");
```

MaxKellermann:
Please amend unmerged commits instead of adding fixup commits.
And I believe this `0x3Fu` deserves some less obscurity; what does it mean, where does it come from? Maybe use symbolic flags?

MaxKellermann:
Rebase, not merge.

PR #2270, opened 2025-04-07
Recommend the more secure https:// instead of git://

dseomn:
From https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols
> Finally, we have the Git protocol. This is a special daemon that comes
> packaged with Git; it listens on a dedicated port (9418) that provides
> a service similar to the SSH protocol, but with absolutely no
> authentication or cryptography.

### DIFF
--- a/doc/developer.rst
+++ b/doc/developer.rst
@@ -91,13 +91,13 @@ Always write your code against the latest git:
 
 .. code-block:: none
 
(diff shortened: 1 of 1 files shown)


MaxKellermann:
6624efc8ef4de6794572b6cdc991f4d3bafea501
Cherry-picked to v0.24.x

PR #2272, opened 2025-04-09
Use redundant call to io_uring_queue_init()

aeolio:
Commit e309941 [1] introduced IORING_SETUP_SINGLE_ISSUER in the call to io_uring_setup(). This flag is available since kernel version 6.0 [2]. Within the system call, the kernel checks supplied against defined flags, and returns -EINVAL if extra flags were supplied.

The initialization in io/uring/Ring.cxx fails with a low priority warning, that is not even printed in the log file (using default settings) and continues without io_uring support.

The Linux kernel supports io_uring since version 5.1, MPD currently has a minimum kernel requirement of 5.6. Ideally this feature should work with all supported kernels.

Introduce a bit mask for setup flags, initialized to the values known to the minimum supported Linux kernel [3]. If the first attempt to setup a uring queue fails with -EINVAL and extra flags were specified, repeat the setup with flags masked to the minimum values.

[1] https://github.com/MusicPlayerDaemon/MPD/commit/e309941
[2] https://www.man7.org/linux/man-pages/man2/io_uring_setup.2.html
[3] https://elixir.bootlin.com/linux/v5.6/source/include/uapi/linux/io_uring.h#L78-L83

### DIFF
--- a/src/io/uring/Ring.cxx
+++ b/src/io/uring/Ring.cxx
@@ -10,8 +10,16 @@ namespace Uring {
 Ring::Ring(unsigned entries, unsigned flags)
 {
 	if (int error = io_uring_queue_init(entries, &ring, flags);
-	    error < 0)
-		throw MakeErrno(-error, "io_uring_queue_init() failed");
+	    error < 0) {
+		if (error == -EINVAL && (flags & ~iouring_setup_mask)) {
+		    flags &= iouring_setup_mask;
+			if (error = io_uring_queue_init(entries, &ring, flags);
+			    error < 0)
+				throw MakeErrno(-error, "io_uring_queue_init() failed (fallback)");
+			}
+		else
+			throw MakeErrno(-error, "io_uring_queue_init() failed");
+		}
 }
 
 Ring::Ring(unsigned entries, struct io_uring_params &params)
--- a/src/io/uring/Ring.hxx
+++ b/src/io/uring/Ring.hxx
@@ -18,6 +18,16 @@ namespace Uring {
 class Ring {
 	struct io_uring ring;
 
+// io_uring_setup() flags supported by Linux kernel 5.6
+unsigned iouring_setup_mask = (
+	IORING_SETUP_IOPOLL |
+	IORING_SETUP_SQPOLL |
+	IORING_SETUP_SQ_AFF |
+	IORING_SETUP_CQSIZE |
+	IORING_SETUP_CLAMP |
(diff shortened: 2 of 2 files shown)


aeolio:
Second attempt: 

- bit mask initialized to the proper IOURING_SETUP_* flags
- second call is executed only if first call resulted in -EINVAL and more than the minimum flags have been supplied

src/io/uring/Ring.hxx
```diff
@@ -18,6 +18,16 @@ namespace Uring {
 class Ring {
 	struct io_uring ring;
 
+// io_uring_setup() flags supported by Linux kernel 5.6
+unsigned iouring_setup_mask = (
```
MaxKellermann, line 22: - bad indentation
- please use doxygen-style comments for API documentation
- the parentheses are useless
- the variable should be `static constexpr`; a regular class field adds unnecessary overhead
- the variable should have a better name; this is *not* the general io_uring setup mask, after all, but the name suggests tha

src/io/uring/Ring.cxx
```diff
@@ -10,8 +10,16 @@ namespace Uring {
 Ring::Ring(unsigned entries, unsigned flags)
 {
 	if (int error = io_uring_queue_init(entries, &ring, flags);
-	    error < 0)
-		throw MakeErrno(-error, "io_uring_queue_init() failed");
+	    error < 0) {
+		if (error == -EINVAL && (flags & ~iouring_setup_mask)) {
+		    flags &= iouring_setup_mask;
```
MaxKellermann, line 15: Bad indent.
And there should be a code comment explaining why this is being done; it's not obvious.

src/io/uring/Ring.cxx
```diff
@@ -10,8 +10,16 @@ namespace Uring {
...
+			if (error = io_uring_queue_init(entries, &ring, flags);
+			    error < 0)
+				throw MakeErrno(-error, "io_uring_queue_init() failed (fallback)");
```
MaxKellermann, line 18: I'm not sure about the `(fallback)` suffix in the error message. I don't think anybody will understand what it means (and the error message is already difficult enough)

MaxKellermann (changes requested):

MaxKellermann:
I think it's a lot simpler to just repeat the call in EventThread::Run() instead. And I'd prefer to have it there because low-level could shouldn't silently disable options that were explicitly specified by the caller.

PR #2273, opened 2025-04-09
Fix hang when activating shout output

bitboulder:
std::span.empty() returns true or 1 on empty span.
In this case, the function should return and not the other way around.


MaxKellermann:
96992617dddaf2b16efa1efa0ff65a1caa16a8f0
cherry-picked to v0.24.x

PR #2296, opened 2025-05-10
Adds audio CD tags, fetched from Musicbrainz server.

tinoucas:
Enabled by default (musicbrainz_tags_enabled), can be disabled with "musicbrainz_tags" bool option (cdio_paranoia plugin).

RemoteTagCache is disabled for this: changing the CD would fail to re-fetch the same looking "cdda:///x" tracks.
Tags are fetched only once per CD change.

For reference, a Musicbrainz URL used here can look like this:
[https://musicbrainz.org/ws/2/discid/iDvvaSuc6SAP8p_2FuI7V3dabAg-?inc=artist-credits+recordings+genres](https://musicbrainz.org/ws/2/discid/iDvvaSuc6SAP8p_2FuI7V3dabAg-?inc=artist-credits+recordings+genres)

### DIFF
--- a/src/input/RemoteTagScanner.hxx
+++ b/src/input/RemoteTagScanner.hxx
@@ -41,6 +41,11 @@ class RemoteTagScanner {
 public:
 	virtual ~RemoteTagScanner() noexcept = default;
 	virtual void Start() = 0;
+
+	virtual bool DisableTagCaching ()
+	{
+		return false;
+	}
 };
 
 #endif
--- a/src/input/plugins/CdioParanoiaCDID.cxx
+++ b/src/input/plugins/CdioParanoiaCDID.cxx
@@ -3,10 +3,42 @@
 
 #include <vector>
 
+/***********************************************/
+/*                                             */
+/* A significant portion of the sources below  */
+/*	has been copied from cd-discid:            */
+/*                                             */
+/* https://github.com/taem/cd-discid           */
+/*                                             */
+/***********************************************/
+
+// Original disclaimer:
+
+/*
+ * Copyright (c) 1999-2003 Robert Woodcock <rcw@debian.org>
+ * Copyright (c) 2009-2013 Timur Birsh <taem@linukz.org>
+ * This code is hereby licensed for public consumption under either the
+ * GNU GPL v2 or greater, or Larry Wall's Artistic license - your choice.
+
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Porting credits:
+ * Solaris: David Champion <dgc@uchicago.edu>
+ * FreeBSD: Niels Bakker <niels@bakker.net>
+ * OpenBSD: Marcus Daniel <danielm@uni-muenster.de>
+ * NetBSD: Chris Gilbert <chris@NetBSD.org>
+ * MacOSX: Evan Jones <ejones@uwaterloo.ca> http://www.eng.uwaterloo.ca/~ejones/
+ *         Thomas Klausner <tk@giga.or.at>
+ * DragonFly: Thomas Klausner <tk@giga.or.at>, http://pkgsrc.se/audio/cd-discid
+ * IRIX: https://github.com/canavan, https://github.com/taem/cd-discid/issues/4
+ */
+
 extern "C" {
 #include <libavutil/sha.h>
 #include <libavutil/base64.h>
-
+#include <libavutil/mem.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h> /* close() */
@@ -24,16 +56,6 @@ extern "C" {
 #include <sys/cdrom.h>
 #define cdte_track_address      cdte_addr.lba
 #define DEVICE_NAME             "/dev/cd0"
-
-#elif defined(sun) && defined(unix) && defined(__SVR4)
-#include <sys/cdio.h>
-#define CD_MSF_OFFSET   150
-#define CD_FRAMES       75
-/* According to David Schweikert <dws@ee.ethz.ch>, cd-discid needs this
- * to compile on Solaris */
-#define cdte_track_address      cdte_addr.lba
-#define DEVICE_NAME             "/dev/vol/aliases/cdrom0"
-
 /* __FreeBSD_kernel__ is needed for properly compiling on Debian GNU/kFreeBSD
    Look at http://glibc-bsd.alioth.debian.org/porting/PORTING for more info */
 #elif defined(__FreeBSD__) || defined(__FreeBSD_kernel__) || defined(__DragonFly__)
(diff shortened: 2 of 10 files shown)


src/input/plugins/CdioParanoiaCDID.cxx
```diff
@@ -0,0 +1,357 @@
...
+
+#elif defined(__sgi)
+#include <dmedia/cdaudio.h>
+#define CD_FRAMES               75               /* per second */
+#define CD_MSF_OFFSET           150              /* MSF offset of first frame */
+#define cdrom_tochdr            CDSTATUS
+#define cdth_trk0               first
+#define cdth_trk1               last
+#define close                   CDclose
+struct cdrom_tocentry
```
MaxKellermann, line 96: Sun? Hurd? SGI? Really?
This awfully looks like code you have copied from somewhere else.

src/input/plugins/CdioParanoiaCDID.cxx
```diff
@@ -0,0 +1,357 @@
...
+
+static char*
+makeMusicBrainzIdWith(int firstTrack, int lastTrack, int leadIn, std::vector<int> frameOffsets)
+{
+	struct AVSHA* sha1 = av_sha_alloc();
+	const int numBitsSha1 = 160;
+	const int digestSize = numBitsSha1 / 8;
+	uint8_t digest[digestSize];
+	const int tempSize = 32;
+	char temp[tempSize];
```
MaxKellermann, line 114: This is a variable-length array which is illegal in C++ (and even deprecated in C).

src/input/plugins/CdioParanoiaCDID.cxx
```diff
@@ -0,0 +1,357 @@
...
+#else
+#error "Your OS isn't supported yet."
+#endif  /* os selection */
+}
+
+static char*
+makeMusicBrainzIdWith(int firstTrack, int lastTrack, int leadIn, std::vector<int> frameOffsets)
+{
+	struct AVSHA* sha1 = av_sha_alloc();
+	const int numBitsSha1 = 160;
```
MaxKellermann, line 110: A signed integer for the number of bits of something??

src/input/plugins/CdioParanoiaCDID.cxx
```diff
@@ -0,0 +1,357 @@
...
+	const int digestSize = numBitsSha1 / 8;
```
MaxKellermann, line 111: Another signed integer??

src/input/plugins/CdioParanoiaCDID.cxx
```diff
@@ -0,0 +1,357 @@
...
+
+	av_sha_init(sha1, numBitsSha1);
+
+	snprintf(temp, tempSize, "%02X", firstTrack);
+	av_sha_update(sha1, (const uint8_t*)temp, strlen(temp));
+
+	snprintf(temp, tempSize, "%02X", lastTrack);
+	av_sha_update(sha1, (const uint8_t*)temp, strlen(temp));
+
+	for (size_t i = 0; i < 100; ++i)
```
MaxKellermann, line 124: What is this "100"?

src/input/RemoteTagScanner.hxx
```diff
@@ -41,6 +41,11 @@ class RemoteTagScanner {
 public:
 	virtual ~RemoteTagScanner() noexcept = default;
 	virtual void Start() = 0;
+
+	virtual bool DisableTagCaching ()
```
MaxKellermann, line 45: What does this method do?

src/input/plugins/CdioParanoiaCDID.cxx
tinoucas: Yes, it is copied. I didn't try to hide it: the original github link is there.

src/input/plugins/CdioParanoiaCDID.cxx
```diff
@@ -0,0 +1,357 @@
...
+	{
+		int offset = 0;
+
+		if (i < frameOffsets.size())
+			offset = frameOffsets[i] + leadIn;
+		snprintf(temp, tempSize, "%08X", offset);
+		av_sha_update(sha1, (const uint8_t*)temp, strlen(temp));
+	}
+	av_sha_final(sha1, digest);
+	free(sha1);
```
MaxKellermann, line 134: Is this call legal? This looks like it will crash because the memory was not allocated by malloc(). Also this isn't exception-safe; this will leak if somebody in the middle throws an exception. All MPD code must be exception-safe, therefore please never use bare allocations.

src/input/plugins/CdioParanoiaCDID.cxx
```diff
@@ -0,0 +1,357 @@
...
+
+	const int outputSize = 128;
+	char *output = new char[outputSize];
```
MaxKellermann, line 137: Another variable-length array.

src/input/plugins/CdioParanoiaCDID.cxx
```diff
@@ -0,0 +1,357 @@
...
+
+		if (i < frameOffsets.size())
+			offset = frameOffsets[i] + leadIn;
+		snprintf(temp, tempSize, "%08X", offset);
+		av_sha_update(sha1, (const uint8_t*)temp, strlen(temp));
+	}
+	av_sha_final(sha1, digest);
+	free(sha1);
+
+	const int outputSize = 128;
```
MaxKellermann, line 136: A signed integer for a buffer size??

src/input/plugins/CdioParanoiaCDID.cxx
tinoucas: That 100 is from Musicbrainz API.
I'll name it to make it clear.

src/input/plugins/CdioParanoiaMusicBrainzTags.hxx
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	static MusicBrainzCDTagCache* getInstance ()
+	{
+		return instance;
+	}
+
+private:
+	std::string lastCdId;
+	std::map<int, TrackInfo> tracks;
```
MaxKellermann, line 77: Why a std::map for tracks? This seems like a poor choice.

src/input/plugins/CdioParanoiaMusicBrainzTags.hxx
```diff
@@ -0,0 +1,170 @@
...
+
+	static MusicBrainzCDTagCache *instance;
+
+	std::map<int, std::set<Listener*> > listeners;
```
MaxKellermann, line 81: What's this, why is it a std::map, what's the index and why is there a space between the two `>`?

src/input/plugins/CdioParanoiaMusicBrainzTags.hxx
```diff
@@ -0,0 +1,170 @@
...
+	Mutex mutex;
+	CurlInit curl;
```
MaxKellermann, line 83: Why initialize the CURL library here (again)?

src/input/plugins/CdioParanoiaMusicBrainzTags.hxx
```diff
@@ -0,0 +1,170 @@
...
+	std::unique_ptr<CurlRequest> request;
+	bool dataReady = false;
+	std::string device;
+
+	class ResponseHandler
+	: public StringCurlResponseHandler
```
MaxKellermann, line 89: Bad indent

src/input/RemoteTagScanner.hxx
tinoucas: It's for the RemoteTagScanner created here. It isn't compatible with RemoteTagCache

src/input/plugins/CdioParanoiaMusicBrainzTags.hxx
```diff
@@ -0,0 +1,170 @@
...
+		void OnEnd() override
+		{
+			auto resp = StringCurlResponseHandler::GetResponse();
+
+			if (MusicBrainzCDTagCache::getInstance()->makeTrackInfoFromXml(resp.body))
+				MusicBrainzCDTagCache::getInstance()->callListeners();
+		}
+
+		void OnError(std::exception_ptr ) noexcept override
+		{
```
MaxKellermann, line 102: No error handling?

src/input/plugins/CdioParanoiaMusicBrainzXMLParser.hxx
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include "CdioParanoiaMusicBrainzTags.hxx"
+
+class MusicBrainzXMLParser
```
MaxKellermann, line 5: What, a class that has no fields and only one method - that design doesn't appear to make any sense.

src/input/plugins/CdioParanoiaMusicBrainzTags.hxx
tinoucas: It made it easier for the listener system created here. What else do you have in mind?

MaxKellermann:
I had only a very rough look, and the code doesn't look so good. Much of it looks like you've "stolen" it from somewhere else (even though you implicitly claim being author by submitting the code here without attribution).

src/input/plugins/CdioParanoiaMusicBrainzTags.hxx
tinoucas: I could log the error I guess

src/input/plugins/CdioParanoiaCDID.cxx
MaxKellermann: But this is horrible code with large swathes of definitions that are not relevant for MPD.
And that GitHub link hidden somewhere in the middle isn't enough. This is a copyright violation!

tinoucas:
Yes, I stole code from cd-discid [https://github.com/taem/cd-discid](https://github.com/taem/cd-discid)
I included the link in the source, I don't know how to make it more obvious. It wasn't my intent to make it "mine".

How can I fix this?

src/input/RemoteTagScanner.hxx
MaxKellermann: You havn't answered my question. But if I have to ask such a question, it means there should be API documentation.

src/input/plugins/CdioParanoiaMusicBrainzTags.hxx
MaxKellermann: "Easier for the listener system", what does that mean? Doesn't answer my question.
Usually, you have contiguous track numbering on CDs, therefore std::vector is better in any case. But I havn't had a closer look; maybe std::vector isn't the best choice, I only assume it's better than std::map, and I question why anybody would ever choose std::map.

MaxKellermann:
You must preserve the whole copyright notice including all author names.
But please don't copy that piece of code. It's bad code and not acceptable for inclusion in MPD.

tinoucas:
That piece of code has the advantage of getting all the info needed for generating the CD id without spinning the drive. The same with cdio would spin the drive for every tag scanner created. As I would have to call cdio_cdda_open each time to get the values. And that's bad for two reasons: first it slows the process a lot and second, it's bad if another track is already playing (and the disc is already reading something else).

Also, though low level, it's already crossplatform (eventhough, yes, sgi is overkill here).

I can try to clean it up, and add a disclaimer on top, but given your reaction, I fear I may be wasting my time (talking about the cleanup, not the disclaimer; I'm adding that ASAP). I don't know of any other solution to achieve this while still compiling on other platforms than Linux.

(I will apply changes for all other remarks regardless, as I value your input even if I get rejected)

src/input/plugins/CdioParanoiaMusicBrainzTags.hxx
tinoucas: I copied code from QobuzClient for this.

src/input/RemoteTagScanner.hxx
tinoucas: Agreed. Though I will need a bit of guidance for that.

src/input/plugins/CdioParanoiaCDID.cxx
tinoucas: Now that you mention it, I'm also surprise it doesn't crash.
Replaced it with a more appropriate av_free

tinoucas:
I pushed the cleanups requested.
How do I update this? Do I make another pull request?

src/input/plugins/CdioParanoiaMusicBrainzTags.hxx
tinoucas: You can sometimes have a "hidden" track 0 on a CD.

src/input/RemoteTagScanner.hxx
tinoucas: Or an example to follow. I don't know where or how to document this.

tinoucas:
I asked for permission for using cd-discid's code in this and got an approval: [https://github.com/taem/cd-discid/issues/7](https://github.com/taem/cd-discid/issues/7)

src/input/RemoteTagScanner.hxx
MaxKellermann: Add Doxygen-style comments to the function.

MaxKellermann:
> I asked for permission for using cd-discid's code in this and got an approval: [taem/cd-discid#7](https://github.com/taem/cd-discid/issues/7)

That permissions is pointless. The license already implies this permission, and the permission "granted" there (by replying with "No problem") doesn't specify any other licensing conditions.

And, as I said, that code is bad and I don't want to see such bad code in MPD.

tinoucas:
Why do call this code "bad"? Can you be more specific?
From what I can read, it's just calls to ioctl, in various platform styles, between an open and a close of the device.

MaxKellermann:
- uses many CPP macros
- declares all variables at the beginning of each function
- too many ifdefs
- uses signed integers for stuff that cannot ever be negative
- uses memset() instead of designated initializers
- uses malloc() and other unmanaged resources (such as file descriptors)
- not exception-safe
- ASCII art in comments
- no documentation
- no proper error handling
- should throw exceptions on error

There may be much more, but I don't feel a deeper analysis is worth my time.

If you submit code specific for certain operating systems, I expect you to test them all (and of course you must be able to explain each and every code line). Do not submit any line of code that you have not tested. Only you are responsible for all of the code you submit, even if it was written by somebody else.

PR #2301, opened 2025-05-18
Add: test program to fetch and parse audio CD info from Musicbrainz.

tinoucas:
Generates a Curl request to Musicbrainz from an audio CD id given as parameter.

CD id must be calculated as decribed here: [https://musicbrainz.org/doc/Disc_ID_Calculation](https://musicbrainz.org/doc/Disc_ID_Calculation)

2 ids are given in test script.
The XML received is parsed and a digest of the tags is printed out to stderr.

(another test program to output id from a loaded CD will come next)

### DIFF
--- a/test/MusicbrainzCDID.cxx
+++ b/test/MusicbrainzCDID.cxx
@@ -1,0 +1,234 @@
+#include "fs/AllocatedPath.hxx"
+#include "lib/cdio/Paranoia.hxx"
+#include "util/StringCompare.hxx"
+#include "util/StringSplit.hxx"
+#include "util/NumberParser.hxx"
+#include "util/ScopeExit.hxx"
+#include "util/PrintException.hxx"
+#include <cdio/cd_types.h>
+#include <cstdint>
+#include <vector>
+
+extern "C" {
+#include <libavutil/sha.h>
+#include <libavutil/base64.h>
+#include <libavutil/mem.h>
+}
+
+#include <stdio.h>
+
+using std::string_view_literals::operator""sv;
+
+struct TrackOffsets
+{
+	int firstTrackNumber = 1;
+	int lastTrackNumber = -1;
+	int leadIn = -1;
+	std::vector<int> frameOffsets;
+
+	bool isValid () const
+	{
+		return lastTrackNumber >= firstTrackNumber
+			&& leadIn > 0
+			&& frameOffsets.size() > 0;
+	}
+};
+
+static const std::string
+makeMusicBrainzIdWithOffsets (const TrackOffsets& trackOffsets)
+{
+	const size_t numBitsSha1 = 160;
+	const size_t digestSize = numBitsSha1 / 8;
+	uint8_t digest[digestSize];
+	const size_t tempSize = 32;
+	char temp[tempSize];
+	struct AVSHA* sha1 = av_sha_alloc();
+
+	AtScopeExit(sha1) { av_free(sha1); };
+
+	av_sha_init(sha1, numBitsSha1);
+
+	snprintf(temp, tempSize, "%02X", trackOffsets.firstTrackNumber);
+	av_sha_update(sha1, (const uint8_t*)temp, strlen(temp));
+
+	snprintf(temp, tempSize, "%02X", trackOffsets.lastTrackNumber);
+	av_sha_update(sha1, (const uint8_t*)temp, strlen(temp));
+
+	const size_t numTracksNeeded = 100; 
+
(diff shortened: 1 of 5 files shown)


test/MusicbrainzCDID.cxx
```diff
@@ -0,0 +1,234 @@
...
+			&& frameOffsets.size() > 0;
+	}
+};
+
+static const std::string
+makeMusicBrainzIdWithOffsets (const TrackOffsets& trackOffsets)
+{
+	const size_t numBitsSha1 = 160;
+	const size_t digestSize = numBitsSha1 / 8;
+	uint8_t digest[digestSize];
```
MaxKellermann, line 55: This STILL is a variable-length array. Please do not do that, it's illegal in C++!

test/MusicbrainzCDID.cxx
```diff
@@ -0,0 +1,234 @@
...
+	const size_t tempSize = 32;
+	char temp[tempSize];
+	struct AVSHA* sha1 = av_sha_alloc();
+
+	AtScopeExit(sha1) { av_free(sha1); };
+
+	av_sha_init(sha1, numBitsSha1);
+
+	snprintf(temp, tempSize, "%02X", trackOffsets.firstTrackNumber);
```
MaxKellermann, line 64: Use fmt instead of stdio

test/MusicbrainzCDID.cxx
```diff
@@ -0,0 +1,234 @@
...
+	av_sha_update(sha1, (const uint8_t*)temp, strlen(temp));
+
+	const size_t numTracksNeeded = 100; 
+
+	if (trackOffsets.frameOffsets.size() > numTracksNeeded)
+	{
+		throw std::runtime_error("Too many tracks found");
+	}
+
+	for (size_t i = 0; i < trackOffsets.frameOffsets.size(); ++i)
```
MaxKellermann, line 77: Use range-based for

test/MusicbrainzCDID.cxx
```diff
@@ -0,0 +1,234 @@
...
+		int offset = 0;
+
+		snprintf(temp, tempSize, "%08X", offset);
+		av_sha_update(sha1, (const uint8_t*)temp, strlen(temp));
+	}
+
+	av_sha_final(sha1, digest);
+
+	const size_t outputSize = 128;
+	char output[outputSize] = { 0 };
```
MaxKellermann, line 96: Another VLA!!!

test/MusicbrainzCDID.cxx
```diff
@@ -0,0 +1,234 @@
...
+
+	av_base64_encode(output, outputSize, digest, digestSize);
+
+	std::string outputString(output);
+
+	std::replace(outputString.begin(), outputString.end(), '=', '-');
+	std::replace(outputString.begin(), outputString.end(), '/', '_');
+	std::replace(outputString.begin(), outputString.end(), '+', '.');
```
MaxKellermann, line 104: What the .... is this?

MaxKellermann (changes requested):
I stopped after a few dozens because I got tired of this piece of code.

MaxKellermann:
Another note: code that is supposed to be used in production does not belong in `test/`.

tinoucas:
You could have said:

- I don't like the variable names: 'x' isn't correct/meaningful, 'y' would be better. You should rename all of them in the same manner/fashion.
- You didn't understand how the test program work: the test program in test/ should call the code present in MPD sources, not duplicate it.
- There are string copies that could have been avoided. Also, you should call empty() on a string instead of comparing to an new object. Same for the map, empty() is more relevant.
- Don't replace one character at a time, use this 'z' method instead. See [file] for an example.

I think you get the point: I'm a human being, not an AI chat bot you can yell at.

But since you're now picking on (all) the variable names (which is new), I guess I'm making some kind of progress.
I'll see if I can improve next week-end. And hopefully, at some point, it will be less tiresome to you.

BTW I have no clue what a "good" variable name is to you. Unless you have clear guidelines this might be tricky to fix.

Also, a variable length array compiles without warning, and runs as expected. It don't get your point. And again, yelling without giving clues on an accepted alternative...

The failed build with clang++ is about an unrelated failure to link with FmtRuntimeError. Which I don't understand but might be worth a look.

MaxKellermann:
> BTW I have no clue what a "good" variable name is to you. Unless you have clear guidelines this might be tricky to fix.

I do. https://mpd.readthedocs.io/en/latest/developer.html#code-style

PR #2306, opened 2025-05-21
decoder/vgmstream: new plugin

leo60228:
This adds a decoder plugin using [vgmstream](https://vgmstream.org/), a library that implements support for many containers and codecs used by video games. It includes support for vgmstream's [out-of-band tagging format](https://github.com/vgmstream/vgmstream/blob/master/doc/USAGE.md#tagging).

The implementation was broadly inspired by the existing `gme` plugin. I attempted implementing support for using streams rather than files, but ran into severe performance issues. This seemed to be a result of severe lock contention caused by vgmstream's access patterns. I'm not sure that fixing this is worth the effort that would be required, particularly considering that libvgmstream is explicitly intended for use with fast backing storage anyway.

Note that vgmstream's CMakeLists.txt did not install libvgmstream.so or its headers until https://github.com/vgmstream/vgmstream/pull/1722, which has not yet appeared in a release. `libvgmstream.h` is intended as a stable public API/ABI, however.


src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -0,0 +1,332 @@
...
+VgmstreamGetFormat(libvgmstream_t *lib, libstreamfile_t *file)
+{
+	uint32_t sample_rate = lib->format->sample_rate;
+	auto sample_format = SampleFormat::UNDEFINED;
+	uint8_t channels = lib->format->channels;
+
+	if (lib->format->sample_format == LIBVGMSTREAM_SFMT_PCM16) {
+		sample_format = SampleFormat::S16;
+	} else if (lib->format->sample_format == LIBVGMSTREAM_SFMT_PCM24) {
+		// mpd doesn't support S24 samples, so reopen the stream as S32
```
MaxKellermann, line 80: Oh yes, MPD does.
https://github.com/MusicPlayerDaemon/MPD/blob/f43d7de99f84f32099f0c9ef92ea96e2306a9d13/src/pcm/SampleFormat.hxx#L22-L26

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -0,0 +1,332 @@
...
+	int subsong;
+};
+
+static constexpr Domain vgmstream_domain("vgmstream");
+
+static libvgmstream_config_t cfg = {};
+
+static void
+VgmstreamGetDouble(const ConfigBlock &block, double *out, const char *name,
+		   double default_value)
```
MaxKellermann, line 40: This seems like a useful method for the `ConfigBlock` class. But it shouldn't write to a pointer; instead, return the value, like all the other `GetBlockValue( )` methods do.

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -0,0 +1,332 @@
...
+#define SUBSONG_PREFIX "song_"
+
+struct VgmstreamContainerPath {
+	AllocatedPath path;
+	int subsong;
+};
+
+static constexpr Domain vgmstream_domain("vgmstream");
+
+static libvgmstream_config_t cfg = {};
```
MaxKellermann, line 36: The variable name `cfg` is rather short and undescriptive for a global variable. What about `vgmstream_config`?

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -0,0 +1,332 @@
...
+
+	return true;
+}
+
+static AudioFormat
+VgmstreamGetFormat(libvgmstream_t *lib, libstreamfile_t *file)
+{
+	uint32_t sample_rate = lib->format->sample_rate;
+	auto sample_format = SampleFormat::UNDEFINED;
+	uint8_t channels = lib->format->channels;
```
MaxKellermann, line 75: Better use one `AudioFormat` variable instead of three variables?

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -0,0 +1,332 @@
...
+
+	if (lib->format->sample_format == LIBVGMSTREAM_SFMT_PCM16) {
```
MaxKellermann, line 77: Can we make this `switch/case`?

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -0,0 +1,332 @@
...
+		const int subsong = lib->format->subsong_index;
+
+		libvgmstream_close_stream(lib);
+
+		libvgmstream_config_t new_cfg = cfg;
+		new_cfg.force_sfmt = LIBVGMSTREAM_SFMT_PCM32;
+		libvgmstream_setup(lib, &new_cfg);
+
+		const int err = libvgmstream_open_stream(lib, file, subsong);
+		assert(err == 0);
```
MaxKellermann, line 90: Uh, not a good way to handle errors!

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -0,0 +1,332 @@
...
+
+static bool
+VgmstreamScanTags(Path path, libvgmstream_t *lib, TagHandler &handler,
+		  ReplayGainInfo *rgi) noexcept
+{
+	bool found_replay_gain = false;
+	bool found_title = false;
+
+	const auto tags_path = AllocatedPath::Build(path.GetDirectoryName(), "!tags.m3u");
+	libstreamfile_t *tags_file = libstreamfile_open_from_stdio(tags_path.c_str());
```
MaxKellermann, line 141: This will break on Windows because `AllocatedPath::c_str()` returns a `const wchar_t*` there.

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -0,0 +1,332 @@
...
+	if (tags_file == nullptr) {
+		if (*lib->format->stream_name)
+			handler.OnTag(TAG_TITLE, lib->format->stream_name);
```
MaxKellermann, line 144: What's the point of returning a tag if this function fails?

MaxKellermann (changes requested):

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
leo60228: Most formats with subsongs supported by vgmstream will provide a human-readable identifier for each subsong. This can't be used as the filename, because it's not necessarily a *unique* identifier, but it's much nicer to show "BGM_SHOP" than "song_002.bank".

leo60228:
(will address review further after work)

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
MaxKellermann: That doesn't answer my question. This is the failure code path.

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
MaxKellermann: Most of the time yes, but most of the time, no error occurs.
What if the file was deleted meanwhile or the NFS server has failed? MPD shouldn't ever crash, unless there is a bug. And assertions are supposed to check for bugs. But what you're checking here isn't a bug, it's a runtime I/O error.

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
leo60228: vgmstream normally stores tags out-of-band in a separate file, but some supported formats have an internal name field. If you think it would make the code clearer I can try to find a way to separate that handling out.

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
leo60228: From what I can tell, vgmstream's PCM24 format is *packed* 24-bit samples. This seemed easier than writing code to pad to 32 bits, though I can try doing that instead.

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
leo60228: I thought it made the code more readable to use if/else-if than to move variable declarations into the parent function. I can change it if you disagree (though this would stop being a problem if I switched to manually unpacking the 24-bit sample, of course).

leo60228:
Huh, I'm not sure what that problem building the library is. Maybe try explicitly specifying `CMAKE_INSTALL_PREFIX`?

You'll need to specify `-DBUILD_SHARED_LIBS=ON` to build/install libvgmstream and not just the CLI tools, by the way. I edited the PR description to note this.

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
MaxKellermann: I think I don't understand at all what this piece of code is supposed to do. If the function fails (returns false), it must not invoke any handler methods. It's a bug.
So whatever you're trying to do, you're doing it wrong. But until I understand the purpose of this, I can't advise what else to do.

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
MaxKellermann: Yes, please. There is `pcm_unpack_24()` which does this for you.

MaxKellermann:
> Huh, I'm not sure what that problem building the library is. Maybe try explicitly specifying `CMAKE_INSTALL_PREFIX`?

I did that, but it turns out the bug was in your vgmstream commit, and fixed by this PR: https://github.com/vgmstream/vgmstream/pull/1723

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
leo60228: Does the refactor I did make the intended logic clearer?

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -113,6 +115,59 @@ ParseContainerPath(Path path_fs)
...
+static void
+VgmstreamScanTags(Path path, const libvgmstream_t *lib, TagHandler &handler,
+		  ReplayGainInfo *rgi) noexcept
+{
+	// check if out-of-band metadata exists
+	const bool found_title = VgmstreamScanTagFile(path, handler, rgi);
+
+	// while out-of-band metadata is preferable, fall back to checking if there's a
+	// title stored in-band. this is particularly important for subsongs, which would
+	// otherwise have no useful title available
```
MaxKellermann, line 166: Don't use the single-line comment syntax if you want multi-line comments.

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -113,6 +115,59 @@ ParseContainerPath(Path path_fs)
...
+		return false;
+
+	AtScopeExit(tags) { libvgmstream_tags_free(tags); };
+
+	libvgmstream_tags_find(tags, path.GetBase().c_str());
+
+	while (libvgmstream_tags_next_tag(tags)) {
+		handler.OnPair(tags->key, tags->val);
+
+		if (StringIsEqualIgnoreCase(tags->key, "TITLE"))
```
MaxKellermann, line 143: Parsing this is redundant and adds unnecessary overhead. You already parse the key in that for loop below.

src/decoder/plugins/VgmstreamDecoderPlugin.cxx
```diff
@@ -113,6 +115,59 @@ ParseContainerPath(Path path_fs)
...
+			found_title = true;
+
+		if (rgi)
+			ParseReplayGainTag(*rgi, tags->key, tags->val);
+
+		for (unsigned i = 0; i < TAG_NUM_OF_ITEM_TYPES; ++i)
+			if (StringIsEqualIgnoreCase(tags->key, tag_item_names[i]))
+				handler.OnTag(static_cast<TagType>(i), tags->val);
```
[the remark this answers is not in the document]
leo60228: Thanks, I didn't see that function. I was looking at the Vorbis comment logic.

PR #2343, opened 2025-07-30
systemd: add DNS-SD configuration file

haansn08:
If mDNS is enabled, this allows the user to announce the presence of MPD on the local network. Remote control apps could then find the MPD service automatically, without the need to enter an local IP address:
```
$ resolvectl query -t PTR _mpd._tcp.local
_mpd._tcp.local IN PTR myhostname._mpd._tcp.local         -- link: enp3s0

$ resolvectl service myhostname._mpd._tcp.local
myhostname._mpd._tcp.local: myhostname.local:6600 [priority=0, weight=0]
                              192.168.1.101                 -- link: enp3s0
                              fe80::9ec7:2bb0:ecfa:41c8%2   -- link: enp3s0
                              (myhostname/_mpd._tcp/local)

-- Information acquired via protocol mDNS/IPv4 in 528us.
```

### References:
* [man systemd.dnssd](https://man.archlinux.org/man/systemd.dnssd.5.en)
* [DNS-Based Service Discovery](https://www.rfc-editor.org/rfc/rfc6763)

### DIFF
--- a/systemd/dnssd/mpd.dnssd
+++ b/systemd/dnssd/mpd.dnssd
@@ -1,0 +1,4 @@
+[Service]
+Name=%H
+Type=_mpd._tcp
+Port=6600


MaxKellermann:
MPD can already do that - is this the same feature but with some systemd-resolved instead of avahi?

haansn08:
I did not realize this! Maybe it would a better idea then to use the D-Bus API from resolved (it provides `RegisterService` and `UnregisterService`, see https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.resolve1.html) instead of this config file. This would also respect the `zeroconf_enabled` option...

MaxKellermann:
Yes, that would be preferable, because it would allow MPD to announce its actual port.
Unfortunately, systemd-resolved lacks the capability to announce a service only on a specific interface. avahi-daemon can do that.

haansn08:
@MaxKellermann I now wrote some logic to register MPD with systemd-resolved: https://github.com/MusicPlayerDaemon/MPD/commit/9db7dad248de2058287c852c75adf47147abb16e

To run this code, you must be authorized to (un-)register services by polkit:
```
# cat /etc/polkit-1/rules.d/90-allow-dnssd-register.rules

polkit.addRule(function(action, subject){
	if ("org.freedesktop.resolve1.register-service" == action.id ||
	    "org.freedesktop.resolve1.unregister-service" == action.id) {
		return polkit.Result.YES;
	}
});
```

I'm still having some trouble with the log messages. They don't seem to show up at all, even with the log level set to "verbose". Maybe you could point me in the right direction?

MaxKellermann:
Is this code line really ever reached?
How did you configure the logger?
You could try `fmt::print(stderr, ...)" instead of using MPD's logger to preclude any MPD logger configuration problem.

haansn08:
@MaxKellermann Thank you for your quick reply!

> Is this code line really ever reached?

Yes. If for example an "mpd" service already exists, the error condition is reached. If I set a breakpoint when debugging, the code line is reached and I can look around.

>  How did you configure the logger?

This is my config file:
```
$ cat .config/mpd/mpd.conf
audio_output {
        type "pipewire"
        name "PipeWire Sound Server"
}

playlist_directory "~/.local/share/mpd"
port "6601"
zeroconf_enabled "yes"
music_directory "~/Downloads"
log_level "verbose"
```

> You could try `fmt::print(stderr, ...)" instead of using MPD's logger

fmt::print also doesn't seem to produce any output:
```
$ ./mpd --no-daemon
server_socket: bind to '0.0.0.0:6601' failed (continuing anyway, because binding to '[::]:6601' succeeded): Failed to bind socket: Address already in use
libsamplerate: libsamplerate converter "Fastest Sinc Interpolator"
vorbis: Xiph.Org libVorbis 1.3.7
opus: libopus 1.5.2
decoder: Decoder plugin "wildmidi" is unavailable: configuration file does not exist: /etc/timidity/timidity.cfg
sndfile: libsndfile-1.2.2
simple_db: reading DB
input: Input plugin "qobuz" is not configured: No Qobuz app_id configured
curl: version 8.16.0
curl: with OpenSSL/3.5.3
^C
```

MaxKellermann:
Add option `--stderr` - MPD logs to syslog by default. You'll find your log messages there.

haansn08:
Nothing in the syslog, but `--stderr` helps, thank you!

Issue #2374, opened 2025-10-16
Missing findaddpl command


FloFaber:
Since we already have these commands:
* `search`
* `searchadd`
* `searchaddpl`
* `find`
* `findadd`

I find that `findaddpl` is missing. Would it be possible to implement?

MaxKellermann:
What would that give you that searchaddpl can't?

FloFaber:
case-sensitivity

MaxKellermann:
OK, such a command would certainly be useful. Do you want to implement it?

FloFaber:
I'll have a look and probably make a PR later.

jcorporation:
Is this new command really required? Since MPD 0.24 there are new explicit case-(in)sensitivity filter expressions: https://mpd.readthedocs.io/en/latest/protocol.html#explicit-case-sensitivity

Is it eventually better to deprecate `find` and `findadd` or what is the use case that can not be accomplished with the above filter expressions?

FloFaber:
Oh, I didn't catch that. It doesn't seem to be mentioned in the changelog.

I think it would be better to deprecate `find`, `findadd` and `playlistfind` then.

PR #2375, opened 2025-10-16
Implement findaddpl command

FloFaber:
https://github.com/MusicPlayerDaemon/MPD/issues/2374


MaxKellermann:
Documentation is missing - see doc/protocol.rst

MaxKellermann:
Please amend the patch to not move code around.

MaxKellermann:
No reply

PR #2394, opened 2025-11-22
Add `gain` URL fragment to set ReplayGain tags

vpatkov:


MaxKellermann:
Nice how you found the simplest way to implement this with existing code libraries. I like that.

PR #2396, opened 2025-11-25
Persist partition functionality

john-:
I ended up separating out the integration test framework and test cases into two separate commits.

The thought was that if it is decided the use the test framework (which is required for the test cases), then maybe the framework should be further split out from the test cases.  The framework could be used by other functionality that would benefit from integration testing.

I am probably over thinking it so let me know.

### DIFF
--- a/src/StateFile.cxx
+++ b/src/StateFile.cxx
@@ -10,12 +10,16 @@
 #include "io/BufferedOutputStream.hxx"
 #include "storage/StorageState.hxx"
 #include "Partition.hxx"
+#include "config/PartitionConfig.hxx"
 #include "Instance.hxx"
 #include "SongLoader.hxx"
 #include "util/Domain.hxx"
 #include "Log.hxx"
+#include "util/StringCompare.hxx"
 
 #include <exception>
+
+#define PARTITION_STATE "partition: "
 
 static constexpr Domain state_file_domain("state_file");
 
@@ -55,14 +59,19 @@ StateFile::IsModified() const noexcept
 inline void
 StateFile::Write(BufferedOutputStream &os)
 {
-	partition.mixer_memento.SaveSoftwareVolumeState(os);
-	audio_output_state_save(os, partition.outputs);
+	for (auto &current_partition : partition.instance.partitions) {
+		if (&current_partition != &partition.instance.partitions.front())
+			os.Fmt(PARTITION_STATE "{}\n",
+			current_partition.name);
+		current_partition.mixer_memento.SaveSoftwareVolumeState(os);
+		audio_output_state_save(os, current_partition.outputs);
+
+		playlist_state_save(os, current_partition.playlist, current_partition.pc);
+	}
 
 #ifdef ENABLE_DATABASE
 	storage_state_save(os, partition.instance);
 #endif
-
-	playlist_state_save(os, partition.playlist, partition.pc);
 }
 
 inline void
@@ -106,13 +115,21 @@ try {
 	const SongLoader song_loader(nullptr, nullptr);
 #endif
 
+	Partition *current_partition = &partition;
+
+	FmtDebug(state_file_domain,
+		 "Starting main loop to read state file (initial partition: '{}')",
+		 current_partition->name);
+
 	const char *line;
 	while ((line = file.ReadLine()) != nullptr) {
-		success = partition.mixer_memento.LoadSoftwareVolumeState(line, partition.outputs) ||
-			audio_output_state_read(line, partition.outputs) ||
+		success = current_partition->mixer_memento.LoadSoftwareVolumeState(line,
+									partition.outputs) ||
+			audio_output_state_read(line, partition.outputs, current_partition) ||
 			playlist_state_restore(config, line, file, song_loader,
-					       partition.playlist,
-					       partition.pc);
+						current_partition->playlist,
+						current_partition->pc) ||
+			PartitionSwitch(line, current_partition);
 #ifdef ENABLE_DATABASE
 		success = success || storage_state_restore(line, file, partition.instance);
 #endif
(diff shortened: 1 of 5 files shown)


test/TestFullIntegration.cxx
```diff
@@ -0,0 +1,974 @@
...
+ * Provides management of:
+ * - Temporary test directories
+ * - MPD configuration files
+ * - MPD process lifecycle
+ * - Socket connections to MPD
+ * - State file manipulation
+ *
+ */
+class MpdTestFixture {
+private:
```
MaxKellermann: Redundant `private`.

test/TestFullIntegration.cxx
```diff
@@ -0,0 +1,974 @@
...
+	fs::path state_path_;
+	fs::path pid_path_;
+	fs::path db_path_;
+	fs::path socket_path_;
+	fs::path music_path_;
+
+	// MPD process and connection state
+	pid_t mpd_pid_ = -1;
+	int sock_ = -1;
+	std::string mpd_executable_ = "mpd";
```
MaxKellermann: Why is this a `std::string`? It is never even modified.

test/TestFullIntegration.cxx
```diff
@@ -0,0 +1,974 @@
...
+	void CreateDummyOggFile(const fs::path& path) {
+		const auto *plugin = encoder_plugin_get("vorbis");
+		if (plugin == nullptr) {
+			throw std::runtime_error("Vorbis encoder plugin not found");
+		}
+
+		ConfigBlock block;
+		block.AddBlockParam("quality", "0.1", -1);
+
+		std::unique_ptr<PreparedEncoder> p_encoder(encoder_init(*plugin, block));
```
MaxKellermann: A complicated coding style you have here. `const auto p_encoder = encoder_init(...)`.

test/TestFullIntegration.cxx
```diff
@@ -0,0 +1,974 @@
...
+			throw std::runtime_error("Failed to init vorbis encoder");
+		}
+
+		AudioFormat audio_format(44100, SampleFormat::S16, 1); // Mono
+		std::unique_ptr<Encoder> encoder(p_encoder->Open(audio_format));
+		if (encoder == nullptr) {
+			throw std::runtime_error("Failed to open vorbis encoder");
+		}
+
+		FileOutputStream os(Path::FromFS(path.c_str()));
```
MaxKellermann: This hand-waves the differences between `std::filesystem` and MPD's own filesystem abstractions away, but does this really work in all cases? It will certainly not even compile on some systems.

test/TestFullIntegration.cxx
```diff
@@ -0,0 +1,974 @@
...
+public:
+	/**
+	 * Construct a new test fixture.
+	 * Creates a temporary directory for the test instance.
+	 *
+	 * @throws std::runtime_error if temporary directory creation fails
+	 */
+	MpdTestFixture() {
+		// Create temporary directory
+		char tmpl[] = "/tmp/mpd_test_XXXXXX";
```
MaxKellermann: Don't hard-code `/tmp`.

test/TestFullIntegration.cxx
```diff
@@ -0,0 +1,974 @@
...
+
+	/**
+	 * Destructor ensures all resources are cleaned up.
+	 * Exception-safe - never throws.
+	 */
+	~MpdTestFixture() noexcept {
+		Cleanup();
+	}
+
+	// Disable copy and move to prevent resource management issues
```
MaxKellermann: What resource management issue?
Is this a LLM-generated comment? Was your code generated by a LLM?

test/TestFullIntegration.cxx
```diff
@@ -0,0 +1,974 @@
...
+	 * @throws std::runtime_error if file write fails
+	 */
+	void WriteConfig(const std::vector<std::string>& extra_config = {}) {
+		std::ofstream conf_file(conf_path_);
+		if (!conf_file) {
+			throw std::runtime_error("Failed to open config file for writing");
+		}
+
+		conf_file << "state_file \"" << state_path_.string() << "\"\n";
+		conf_file << "pid_file \"" << pid_path_.string() << "\"\n";
```
MaxKellermann: Why are you configuring a PID file?

test/TestFullIntegration.cxx
```diff
@@ -0,0 +1,974 @@
...
+	[[nodiscard]] bool Connect() {
+		sock_ = socket(AF_UNIX, SOCK_STREAM, 0);
+		if (sock_ < 0) {
+			perror("socket failed");
+			return false;
+		}
+
+		struct sockaddr_un addr{};
+		addr.sun_family = AF_UNIX;
+		std::strncpy(addr.sun_path, socket_path_.c_str(), sizeof(addr.sun_path) - 1);
```
MaxKellermann: strncpy!!!
Are you aware of the properties of strncpy?

test/TestFullIntegration.cxx
```diff
@@ -0,0 +1,974 @@
...
+	 * @param command The MPD protocol command to send
+	 * @return The response from MPD, or error message if not connected
+	 */
+	[[nodiscard]] std::string SendCommand(const std::string& command) {
+		if (sock_ < 0) {
+			return "ERROR: Not connected";
+		}
+
+		std::string cmd = command;
+		if (cmd.empty() || cmd.back() != '\n') {
```
MaxKellermann: If the command is empty, send a bare newline character? What's the point of that?

MaxKellermann (changes requested):
I stopped after a while. Sorry, but this looks like a LLM-generated mess.

john-:
Thanks for taking the time to comment.  Based on your feedback I agree that it is a mess.  I did not take the testing seriously enough that is clear.

I will take another pass at seeing if I can figure out how to test directly against StateFile().  I assume that is the proper way of testing the changes.

Otherwise, someone else may be in better position to work on a change like this.

MaxKellermann:
Adding unit tests is very valuable, and I welcome that you tried to have a unit test for new code, and a "big one" that really starts a MPD process is certainly also a very good thing. Unit test code doesn't need to be as perfect as actual production code, but what you have now isn't good and robust enough. A fragile unit test that fails half of the time (maybe due to timing problems) only annoys, but isn't helpful.
I wish you would improve it to a point that it can be merged, but it's really a big and complicated project to get all aspects right. I feel that might be a thing for another day.

Maybe for this one, we should figure out how to put the unit test at a lower level, to only check the actual state file code without launching a full MPD process. Of course, that has its own set of difficulties, because the state file code is intertwined with all subsystems that it affects. The `StateFile` class has a reference to a `Partition` object, and thus depends on everything that is connected to that - which is really everything. It's dirty glue code. I have no good idea, sorry.

john-:
I will explore the two options further.  I expect it will take me a chunk of time so will close this pull request.  Thanks again for your insight and for MPD.  We use it every day here.

PR #2408, opened 2026-01-05
Support UTF-8 BOM stripping in `TextInputStream`

slqy123:
Currently, `TextInputStream` treats all text as raw bytes and does not strip UTF-8 BOM (`EF BB BF`). Cue sheets saved with UTF-8 BOM can fail to parse the first line because the first token is prefixed with BOM bytes.

This PR Strips UTF-8 BOM in `TextInputStream` constructor. It tries to read the first 3 bytes into buffer, and consume those bytes if a BOM is present.

### DIFF
--- a/src/input/TextInputStream.cxx
+++ b/src/input/TextInputStream.cxx
@@ -15,6 +15,26 @@ TextInputStream::~TextInputStream() noexcept = default;
 char *
 TextInputStream::ReadLine()
 {
+	if (!bom_checked) {
+		bom_checked = true;
+
+		/* try to strip a UTF-8 BOM;
+		   keep all bytes if it's not a BOM */
+		auto dest = buffer.Write();
+		assert(dest.size() >= 3);
+		dest = dest.first(3);
+		size_t nbytes = is->LockRead(std::as_writable_bytes(dest));
+		buffer.Append(nbytes);
+
+		auto r = buffer.Read();
+		if (r.size() >= 3 &&
+		    static_cast<unsigned char>(r[0]) == 0xEF &&
+		    static_cast<unsigned char>(r[1]) == 0xBB &&
+		    static_cast<unsigned char>(r[2]) == 0xBF) {
+			buffer.Consume(3);
+		}
+	}
+
 	char *line = ReadBufferedLine(buffer);
 	if (line != nullptr)
 		return line;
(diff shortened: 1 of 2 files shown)


MaxKellermann:
I don't think that's the right place to do this. This makes the constructor potentially blocking (and out of the caller's control), and the constructor can now throw on I/O error (which will crash MPD). I find it rather surprising that the constructor will implicitly do blocking I/O. Plus, it may deadlock if the caller happens to already hold a mutex lock.

slqy123:
> I don't think that's the right place to do this. This makes the constructor potentially blocking (and out of the caller's control), and the constructor can now throw on I/O error (which will crash MPD). I find it rather surprising that the constructor will implicitly do blocking I/O. Plus, it may deadlock if the caller happens to already hold a mutex lock.

Thank you for the feedback. I will move this to `TextInputStream::ReadLine` and add a flag to track if the BOM was checked.

MaxKellermann:
That's much better. Now please amend the commit message - it contains less than the PR text, and that's unfortunate, because PR text gets lost on GitHub's proprietary website.
If this fixes a problem users have currently, it should probably go to the stable branch v0.24.x, but I can easily cherry-pick it over there.

slqy123:
> If this fixes a problem users have currently, it should probably go to the stable branch v0.24.x, but I can easily cherry-pick it over there.

I didn't find an existing issue for this.
Since it only breaks the first line, most of the cuesheet still parses. it's easy to miss, but users are still affected. So I believe it should go to the stable branch.

MaxKellermann:
You edited the commit message from
"input/TextInputStream: strip UTF-8 BOM"
to
"input/TextInputStream: support UTF-8 BOM stripping in TextInputStream"

This is still
- less information than the PR
- not more than before; it just adds the (weasel) word "support" and mentions TextInputStream twice, but that doesn't add any information

slqy123:
> You edited the commit message from "input/TextInputStream: strip UTF-8 BOM" to "input/TextInputStream: support UTF-8 BOM stripping in TextInputStream"
> 
> This is still
> 
>     * less information than the PR
> 
>     * not more than before; it just adds the (weasel) word "support" and mentions TextInputStream twice, but that doesn't add any information

OK, I have updated my commit message with a detailed explaination

MaxKellermann:
Cherry-picked to v0.24.x: 98bb249d094180cbe8b8fea122aff8b6e1caf438

PR #2409, opened 2026-01-07
Persist partition state and use integration testing

john-:
This PR enables the persisting of state information for all partitions.

The testing is done by executing the minimum needed MPD code in order to test the StateFile class.  I would consider it integration testing.

The [previous pull request](https://github.com/MusicPlayerDaemon/MPD/pull/2396) for this functionality had testing that was more like system testing:  The entire mpd applcation is loaded and interacted with via its inputs/outputs (e.g. configuration file).

High level notes:

- `test/meson.build` was modified to create a testing library file that includes at least enough MPD to compile the test cases.

- Mocking (gmock) is used for loading songs into the playlist and for mounting storage.

    Stock MPD functionality requires songs be in the MPD DB and use of valid storage.  The mocking overrides this behavior for testing.

- I reviewed the comments for the previous pull request and attempted to address them with this testing approach.

If there is too much basic stuff I am missing give me the feedback you have time for.

### DIFF
--- a/test/MockPlaylistSong.cxx
+++ b/test/MockPlaylistSong.cxx
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The Music Player Daemon Project
+
+/**
+ * Mock implementation of playlist_check_translate_song for testing.
+ *
+ * This minimal mock allows songs to be loaded into the queue during tests
+ * without requiring a real database or file system. It simply returns true
+ * for all songs, bypassing the normal validation that would fail in a test
+ * environment.
+ */
+
+#include "playlist/PlaylistSong.hxx"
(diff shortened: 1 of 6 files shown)


john-:
I am closing this to investigate an issue with how this interacts with myMPD on its first use of MPD with the new state file.

john-:
It turns out issue I encountered is unrelated to these mpd changes and unrelated to myMPD.  The problem was a timing issue on boot when mounting my music library over nfs.

MaxKellermann:
So.. is this PR branch ready for review after all?
(And it confused me that you created a new PR.)

john-:
Yes it is ready for review.  As far as why I created a new pull request:  I don't have a reason.  I falsely assumed it was the right thing to do.

PR #2416, opened 2026-01-25
Update mpd.conf.5.rst: details about playlist_directory

thomas-mc-work:
Add a note that this folder (sadly) only allows for a flat file structure.


MaxKellermann:
v0.24.x a017eb6886eefb6892df9a661c3365517afb4a98
(and tweaked the wording to make it less verbose)

Issue #2418, opened 2026-01-28
Directory filter / group by / command


thjbdvlt:
## Feature request

I'm writing a client for MPD, and I need to know `album` and `artist` for a given directory, *but not recursively*. I can't find a good solution with the current commands and filters (as i try to avoid `listall` and `listallmeta` as the doc says).

I wish there was a non-recursively version of `list album "(base 'directory')"`, because for now I have to find some workarounds, such as:

```
list album "(base 'directory')" "(!(base 'dir/subdir'))" "(!(base 'dir/subdir2'))" ...
lits album "(file =~ 'directory/[^/]+$')"
```

I think that it's very usual to group music collections in structure like:

```
~/music
~/music/artist1
~/music/artist1/album1
~/music/artist1/album2
~/music/artist1/album2/pictures/
~/music/artist2
~/music/artist2/lives/
~/music/artist2/lives/album1
~/music/artist2/lives/album2
~/music/subcollection/artist3
~/music/subcollection/artist3/album1
```

The `albumart` command is based on that common directory structure.

I don't want to use `list artist album ALBUMTITLE` because many albums can have the same name. And `list album group artist` doesn't work, because in an album directory there could be many artists (e.g. for collective albums or albums containing a song by many artists), which is not the same as different albums with the same title by differents artists. I can use `list file album ALBUMTITLE` and check the file paths in my application, but it seems even worst than the previous workarounds.

What I would like to use is as simple as:

```
list album "(dir 'somedir')"
list artist group somedirectory
```

I really need the directory (album URI) because of non-songs files stored in them, such as `tags` (keywords inherited from ancestores directories) and `about` (text about the album).

But maybe you have any ideas?

(Of course I could try to make a PR for that feature.)

MaxKellermann:
I think a filter directive `directory` (similar to `base` but not recursive) would be a useful protocol feature. PR welcome, of course.

> because in an album directory there could be many artists (e.g. for collective albums or albums containing a song by many artists)

That's why the `AlbumArtist` tag exists. If you work with albums, you should always use `AlbumArtist` instead of `Artist`. Songs that don't have an `AlbumArtist` tag wil automatically use the `Artist` tag instead, therefore it always works.

thjbdvlt:
Great! I'll see what I can do. I just looked at the codebase and found where (I think) it should go :)

PR #2419, opened 2026-01-29
(wip) song/DirectorySongFilter: new filter

thjbdvlt:
Adds a new filter `directory` similar to `base` but non-recursive.

See https://github.com/MusicPlayerDaemon/MPD/issues/2418

~~(The commit adds a `GetDirectory` method on `LightSong` struct, similar to `GetURI`. It could be done from within the filter functions, but I think that other (future) features could use this, e.g. `group directory` or `list directory`.)~~

In its current state (a  partial and unoptimized copy of BasSongFilter), it's still very slow!

### DIFF
--- a/src/song/BaseSongFilter.hxx
+++ b/src/song/BaseSongFilter.hxx
@@ -8,17 +8,22 @@
 
 class BaseSongFilter final : public ISongFilter {
 	std::string value;
+  bool recursive;
 
 public:
 	BaseSongFilter(const BaseSongFilter &) = default;
 
 	template<typename V>
-	explicit BaseSongFilter(V &&_value)
-		:value(std::forward<V>(_value)) {}
+	explicit BaseSongFilter(V &&_value, bool _recursive)
+		:value(std::forward<V>(_value)), recursive(_recursive) {}
 
 	const char *GetValue() const noexcept {
 		return value.c_str();
 	}
+
+  bool IsRecursive() const noexcept {
+    return recursive;
+  }
 
 	ISongFilterPtr Clone() const noexcept override {
 		return std::make_unique<BaseSongFilter>(*this);
--- a/src/song/Filter.hxx
+++ b/src/song/Filter.hxx
@@ -29,6 +29,11 @@
 enum TagType : uint8_t;
 struct LightSong;
 
+/**
+ * Base URI and recursive flag.
+ */
+struct Base { const char *uri; bool recursive; };
+
 class SongFilter {
 	AndSongFilter and_filter;
 
@@ -80,11 +85,11 @@ public:
 	bool HasFoldCase() const noexcept;
 
 	/**
-	 * Returns the "base" specification (if there is one) or
-	 * nullptr.
+	 * Returns the "base" specification, containing URI (if any) and
+	 * recursive flag (default to true).
 	 */
 	[[gnu::pure]]
-	const char *GetBase() const noexcept;
+	const Base GetBase() const noexcept;
 
 	/**
 	 * Create a copy of the filter with the given prefix stripped
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,7 @@ ver 0.25 (not yet released)
   - implement "window" parameter for command "list"
   - new command "stringnormalization"
   - show detailed seek errors
+  - new filter "directory": non-recursive version of "base"
 * decoder
   - faad: implement seeking
   - faad: output 32 bit floating point samples instead of 16 bit integer
--- a/doc/protocol.rst
+++ b/doc/protocol.rst
@@ -213,6 +213,8 @@ of:
   songs in the given directory (relative to the music
   directory).
 
+- ``(directory 'VALUE')``: non-recursive version of ``(base 'VALUE')``.
+
 - ``(modified-since 'VALUE')``: compares the
   file's time stamp with the given value (ISO 8601 or UNIX
   time stamp).
--- a/src/db/Selection.cxx
+++ b/src/db/Selection.cxx
@@ -3,6 +3,7 @@
 
 #include "Selection.hxx"
 #include "song/Filter.hxx"
+#include "song/BaseSongFilter.hxx"
 
 DatabaseSelection::DatabaseSelection(const char *_uri, bool _recursive,
 				     const SongFilter *_filter) noexcept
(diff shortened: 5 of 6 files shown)


MaxKellermann:
Please add protocol documentation

thjbdvlt:
I'm really sorry: I thought it was easier to implement that it's actually. Thus I thought that my kinda superficial C++ knowledge would do it; it doesn't.
I've now specified "work-in-progress", and I'll come with a better proposal, maybe by suclassing or wrapping `BaseSongFilter` (because of functions `HasOtherThanBase` and others). Or maybe by adding a `recursive` field in `BaseSongFilter` if it has no impact on performance.

MaxKellermann:
I forgot about `HasOtherThanBase` - but it seems this method is obsolete and can be removed; the last user was removed 8 years ago in commit 9cc960ac5ee4b04562cd34001165f83f6c4fdb0c

I thought this PR was good, I just didn't like the unnecessary allocation.

thjbdvlt:
But it (or `WithoutBasePrefix`/`GetBase`) seems to do a great job for optimization. I didn't figured out yet how exactly and how `DirectorySongFilter` can take advantage of it without duplicating code.

My directory filter is much slower.

I'm probably not using MPD protocol very well (there's a loop in which I call `list album base DIR`, because I need an array of `Albums,Artists,Directory` tuples, and I would actually like to do `list album group albumartist group directory`), but the difference is huge; I just can't use my new filter in my client while `base` is instantaneous.

MaxKellermann:
True, your code is completely unoptimized. It's a naive approach - and it works. I like naive code that just works!

But of course, it's at the same time somewhat clumsy. A search with your filter will traverse the whole database and check the URI of each and every song. The right approach would be to go to that one `Directory` object and apply the rest of the filter on all `Song` objects inside that directory. So, for the common case, your code would never actually be called; your new class would just be a marker, just like `BaseSongFilter` is - its code is just as well not called most of the time.

First step towards optimizing it: `SongFilter::GetBase()` should recognize your new class, too. That way, you get acceptable performance because only a small branch will be searched. At this point, yes, you could consider renaming `BaseSongFilter` to `DirectorySongFilter` and add the `recursive` flag - you had that idea, and I think it's a good idea. That would automatically give you this optimization.

Next: in `DatabaseSelection::DatabaseSelection()`, copy the `recursive` flag. Then the search will not visit subdirectories.

For experts: make your new feature work with `ProxyDatabasePlugin`. This requires adding the feature to libmpdclient, checking libmpdclient version at compile time, check other MPD version at runtime.

thjbdvlt:
Thank you so much! Your hints gave me good insights into the code. This leads to a solution that's both faster and more concise.

Now I have a new question.

To access `BaseSongFilter.recursive` flag from within `DatabaseSelection()`, I need to loop over filters in the same way `GetBase`, `HasOtherThanBase` and `WithoutBasePrefix` do:

```cpp
for (const auto &i : filter->GetItems()) {
  const auto *f = dynamic_cast<const BaseSongFilter *>(i.get());
  if (f != nullptr && !f->recursive) {
    recursive = false;
    break;
  }
}
```

I can either make `recursive` flag public and put the loop directly inside `DatabaseSelection()`, since it's used only once and is very small, or I can add a method to the `...Base...` toolkit. A new method (e.g. `HasNonRecursiveFlag`) seems more consistent and helps to keep `DatabaseSelection` code small, but also seems a bit unnecessary.

MaxKellermann:
Don't use negated names. Call it `IsRecursive()`, not `HasNonRecursiveFlag()`. It's too easy to get double negation wrong.
And I'd rather have this getter method than making the field public. Sure, getters are ugly boilerplate, but I tend to prefer that and making the field inaccessible. Whoever has a non-const reference to such an object should not be able to manipulate the field, as these objects are supposed to be immutable.

MaxKellermann:
When you do fixups for unmerged commits, please don't add fixup commits on top; instead, amend the existing commit.
This PR has 4 commits, but they do just one single feature, therefore should result in only one commit.
A mention in NEWS would be good.

thjbdvlt:
> When you do fixups for unmerged commits, please don't add fixup commits on top; instead, amend the existing commit. This PR has 4 commits, but they do just one single feature, therefore should result in only one commit. A mention in NEWS would be good.

I'm sorry.  May I `rebase` and `push --force` to have a single commit?

MaxKellermann:
> May I `rebase` and `push --force` to have a single commit?

I don't understand. How do you produce a single commit by rebasing?
(Force-pushing is of course necessary to publish amended changes.)

For me, it does not matter what tools you use to produce a proper PR branch. I don't use git for this, but my tool (stgit) does result in a proper git branch.

thjbdvlt:
I meant `git rebase -i HEAD~4` and then `squash` command to make one single commit. You actually answered my question, which was more about the `force --push` (but then I used it anyway, since there's no other solution). 

(I didn't know `stgit` and will try it!)

src/song/BaseSongFilter.hxx
```diff
@@ -8,18 +8,23 @@
 
 class BaseSongFilter final : public ISongFilter {
 	std::string value;
+  bool recursive;
```
MaxKellermann, line 11: Bad indent

src/song/BaseSongFilter.hxx
```diff
@@ -8,18 +8,23 @@
...
-	explicit BaseSongFilter(V &&_value)
-		:value(std::forward<V>(_value)) {}
+	explicit BaseSongFilter(V &&_value, bool _recursive)
+		:value(std::forward<V>(_value)), recursive(_recursive) {}
 
 	const char *GetValue() const noexcept {
 		return value.c_str();
 	}
 
+  bool IsRecursive() const noexcept {
```
MaxKellermann, line 24: Bad indent

src/song/Filter.hxx
```diff
@@ -29,6 +29,11 @@
 enum TagType : uint8_t;
 struct LightSong;
 
+/**
+ * Base URI and recursive flag.
+ */
+struct Base { const char *uri; bool recursive; };
```
MaxKellermann, line 35: Bad coding style

src/song/Filter.hxx
```diff
@@ -80,11 +85,11 @@ public:
 	bool HasFoldCase() const noexcept;
 
 	/**
-	 * Returns the "base" specification (if there is one) or
-	 * nullptr.
+	 * Returns the "base" specification, containing URI (if any) and
```
MaxKellermann, line 88: What is returned if there is no URI?

src/song/Filter.hxx
```diff
@@ -80,11 +85,11 @@ public:
...
+	 * recursive flag (default to true).
```
MaxKellermann, line 89: What's the meaning of the recursive flag if there is no URI?
What is the point of documenting the default value here? This method and this whole class doesn't care, does it?

thjbdvlt: Documenting this here wasn't a good idea. I will change that.

About the meaning of the recursive flag if there's no URI: I may have done this wrong. Instead of a `Base` struct with a `nullptr` as `uri` field, it could be better to return a `nullptr`.

src/song/Filter.hxx
thjbdvlt: (In current PR state, `nullptr` is returned, but see my answer in the next review comment.)

Issue #361, opened 2018-10-03
Stream does not resume after hibernate


skidoo23:
It's a feature request - not a bug report. MPD_ from git, Debian testing x_64, ALSA

1. play stream with MPD
2. systemctl hibernate
3. wait some time
4. restart MPDs host machine
5. MPD plays stream from buffer
6. after a while silence - MPD does not try to reconnect or skip

Expected behaviour
5. MPD detects resume
6. MPD empties buffer
7. optional: MPD sends a short notice to clients eg reconnecting to stream server
8. MPD reconnects to streaming source. If connect fails MPD continues with next item on playlist

Issue #1630, opened 2022-09-27
When a stream gets interrupted "playing silence to avoid xrun" is not a solution


GK-774:
If, for a stream, a network interruption occurs which exceeds the buffer, silence to avoid xrun is played but obviously no reconnect is attempted. This will go on forever if not manually interrupted. A stop and start reconnected to the stream immediately in all cases I have encountered so far.

e.g:
Sep 27 19:38:01 mpd8.hal9000.at mpd[328913]: alsa_output: Decoder is too slow; playing silence to avoid xrun
... and so on
Sep 27 19:47:41 mpd8.hal9000.at mpd[328913]: alsa_output: Decoder is too slow; playing silence to avoid xrun

In case of a stream just waiting for data to come in again, like in case of a static file, is obviously no solution, it won´t come in. The streaming server expects the client to reconnect (or it´s gone).
The default buffer time for (as an example) shoutcast server is 1s, so it would be sensible that after 1s of playing silence a reconnect (or a stop / start sequence) is attempted.

GK-774:
## Bug report
### Describe the bug
If a http stream is interrupted for a short time, wheter a reconnect is attempted nor a stop of playback is done. Instead silence is played endlessly. A mpc stop and mpc start immidiately reconnects to the stream and plays normally then.

## Expected Behavior
Reconnecting to the stream would be attempted (basically stop/start) or if this is not feasible to implement, playback stopped and an error reported.

## Actual Behavior
The actual error condition is obfuscated by playing silence in an endless loop.

## Version
Music Player Daemon 0.23.9 (0.23.9)
Copyright 2003-2007 Warren Dukes <warren.dukes@gmail.com>
Copyright 2008-2021 Max Kellermann <max.kellermann@gmail.com>
This is free software; see the source for copying conditions.  There is NO
warranty; not even MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Database plugins:
 simple proxy upnp

Storage plugins:
 local udisks nfs curl

Neighbor plugins:
 upnp udisks

Decoders plugins:
 [mad] mp3 mp2
 [mpg123] mp3
 [vorbis] ogg oga
 [oggflac] ogg oga
 [flac] flac
 [opus] opus ogg oga
 [sndfile] wav aiff aif au snd paf iff svx sf voc w64 pvf xi htk caf sd2
 [audiofile] wav au aiff aif
 [dsdiff] dff
 [dsf] dsf
[...]

PR #2442, opened 2026-03-02
Fix fail_timer duration multiplication

slqy123:
Remove incorrect `* 1000` multiplier in fail_timer.Check() call.

When the PeriodClock API was migrated to std::chrono in commit 401189984 , the fail_timer.Check() call was not updated. This caused the timer to check for a duration that was 1000 times longer than intended.

```c
// before
static constexpr unsigned REOPEN_AFTER = 10;
// after
static constexpr PeriodClock::Duration REOPEN_AFTER = std::chrono::seconds(10);

// It actually waits 10000 seconds
fail_timer.Check(REOPEN_AFTER * 1000);
```

### DIFF
--- a/src/output/Control.cxx
+++ b/src/output/Control.cxx
@@ -305,7 +305,7 @@ AudioOutputControl::LockUpdate(const AudioFormat audio_format,
 
 	if (enabled && really_enabled) {
 		if (force || !fail_timer.IsDefined() ||
-		    fail_timer.Check(REOPEN_AFTER * 1000)) {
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Wow, this bug has survived a loooong time!

slqy123:
> Wow, this bug has survived a loooong time!

Yes. This may fixes #361, #1630 (maybe? I still get `playing silence to avoid xrun` while debugging, but the stream reconnects successfully after the retry delay)

PR #2448, opened 2026-03-17
upnp: allow 1.14.30

PPN-SD:
ABIbreak in 1.14.26 has been reverted w/ 1.14.30, reapplied in 1.18.X
align with upstream versions and make easier packaging.

### DIFF
--- a/src/lib/upnp/meson.build
+++ b/src/lib/upnp/meson.build
@@ -13,7 +13,10 @@ endif
 if upnp_option == 'auto'
   upnp_dep = dependency('libupnp', version: '>= 1.8', required: false)
 
-  if upnp_dep.found() and upnp_dep.version().version_compare('>= 1.14.26')
+  if upnp_dep.found() and (
+    ( upnp_dep.version().version_compare('>= 1.14.26') and upnp_dep.version().version_compare('< 1.14.30') )
+    or upnp_dep.version().version_compare('>= 1.18.0')
+  )
     warning('Your libupnp version is known to be broken, see https://github.com/pupnp/pupnp/issues/528 - disabling')
     upnp_dep = dependency('', required: false)
   endif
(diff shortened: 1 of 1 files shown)


MaxKellermann:
There is still no reply from upstream to my bug report: https://github.com/pupnp/pupnp/issues/528

I don't understand your version_compare call. According to the Meson documentation, it allows only a single parameter: https://mesonbuild.com/Syntax.html#version_compare - but you pass three. What does that mean?

PPN-SD:
> I don't understand your version_compare call. According to the Meson documentation, it allows only a single parameter: 
https://mesonbuild.com/Syntax.html#version_compare

Two are allowed since [meson-1.8](https://mesonbuild.com/Release-notes-for-1-8-0.html#version_compare-now-accept-multiple-compare-strings). But maybe you won't up this for that.

> but you pass three. What does that mean?

Two bad pushes, sry.

MaxKellermann:
The stable branch (v0.24.x) must remain compatible with Meson 1.0. 
The unstable branch (master) must remain compatible with Debian stable, so we may update to 1.7 but not newer.

Your PR targets master - are you aware that this will not be part of the next bug-fix release then?

MaxKellermann:
> The unstable branch (master) must remain compatible with Debian stable, so we may update to 1.7 but not newer.

And it should be compatible with GitHub Actions runner images - as you see in the build failure, it has only Meson 1.3.2. That is not the most important thing (because GitHub often lags behind more than even Debian), it's most important to be compatible with Debian Stable - Debian is a good benchmark for "builds everywhere (unless people are reaaaaaaally on old OS)".

PPN-SD:
> The stable branch (v0.24.x) must remain compatible with Meson 1.0. The unstable branch (master) must remain compatible with Debian stable, so we may update to 1.7 but not newer.

OK.

> Your PR targets master - are you aware that this will not be part of the next bug-fix release then?

The most important thing is in my opinion to let the distributions apply version 1.14.30 and not directly block version >= 1.14.26. It doesn’t matter what happens to this PR, to be honest.

MaxKellermann:
> It doesn’t matter what happens to this PR, to be honest.

Does this mean you don't want to fix this PR?

PPN-SD:
It should be fine now.

13ilya:
https://build.opensuse.org/package/show/home:13ilya:branches:multimedia:apps/mpd
mpd 0.24.10
[   34s] Run-time dependency libupnp found: YES 1.18.5
[   34s] src/lib/upnp/meson.build:20: WARNING: Your libupnp version is known to be broken, see https://github.com/pupnp/pupnp/issues/528 - disabling

PPN-SD:
This PR allowed 1.14.30, but 1.18.X is still unsupported.

PR #2451, opened 2026-03-21
Fixed UPnP for full URLs with scheme

ypopovych:
Fixes #2450 

To be frank I have zero confidence in URL fix, I simply patched two URL methods to return relative if it has scheme.

XML parser fix is a better one. Current version breaks on URLs with parameters. Parser sends URL by parts split on each `&`. With this fix all of them will be concatenated together.

Tested on Jellyfin server - works fine (slow, but works).

### DIFF
--- a/src/TagAny.cxx
+++ b/src/TagAny.cxx
@@ -48,6 +48,9 @@ TagScanFile(const Path path_fs, TagHandler &handler)
 static std::string
 ResolveUri(std::string_view base, const char *relative)
 {
+	if (uri_has_scheme(relative)) {
+		return std::string(relative);
+	}
 	while (true) {
 		const char *rest = StringAfterPrefix(relative, "../");
 		if (rest == nullptr)
--- a/src/db/plugins/upnp/Directory.cxx
+++ b/src/db/plugins/upnp/Directory.cxx
@@ -156,6 +156,8 @@ protected:
 				if (duration != nullptr)
 					tag.SetDuration(ParseDuration(duration));
 
+				object.url.clear();
+
 				state = RES;
 			}
 
@@ -205,7 +207,7 @@ protected:
 			break;
 
 		case RES:
-			object.url.assign(s);
+			object.url.append(s);
 			break;
 
 		case CLASS:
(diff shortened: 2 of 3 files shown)


MaxKellermann:
Your commit consists of (at least) 3 distinct changes that deserve to be a separate commit, and your commit message documents none of these changes. It doesn't even describe the problem to be solved, let alone how and why. I don't understand this PR.

ypopovych:
Sorry about that - here’s a clearer explanation of what was done and why. 

These fixes were added because the UPnP database plugin does not handle URLs correctly. There are two separate issues:

### Bug 1

The server may return full URLs instead of relative paths in XML responses.

The current UPnP implementation assumes that only paths are returned. It then joins the returned value with the server’s base URL. When a full URL is returned, this results in a malformed URL like: `http://192.168.1.2/.../http://192.168.1.2/...`.

This was fixed in two places:

- In [src/lib/upnp/ContentDirectoryService.cxx](https://github.com/MusicPlayerDaemon/MPD/pull/2451/changes#diff-fba83b6b2242acdc33354d7214dfae2d03564669fb16bfe13c3095ac385dcd4e), `uri_apply_base` was replaced with `uri_apply_relative`, which correctly detects full URLs and avoids concatenating them.
- In [src/TagAny.cxx](https://github.com/MusicPlayerDaemon/MPD/pull/2451/changes#diff-bd44ecdc254e1308428b3b4c06dbf5337bc1abf11c0aaba63a08529c60735618), a check for the URL scheme was added to return early when a full URL is detected.

### Bug 2

This issue lies in the XML parser logic.

The current implementation incorrectly assumes that a URL is returned as a single string. However, when a URL contains query parameters, the parser invokes the callback separately for each segment. As a result, only the last query parameter is captured instead of the full URL.

This was fixed in [‎src/db/plugins/upnp/Directory.cxx](https://github.com/MusicPlayerDaemon/MPD/pull/2451/changes#diff-3147bdbb72c79322c2da284f9bb6c27331a97c51df62a352d5fa32fd22b13fdd). The URL buffer is now cleared at the start of each element (once per element) and then incrementally built by appending data in the character data callback.

I hope this explanation helps, Yehor.

ypopovych:
How do you want me to split it? Do a different PRs? Or to have 2(3?) commits in the one PR?

MaxKellermann:
> How do you want me to split it? Do a different PRs? Or to have 2(3?) commits in the one PR?

You can add any number of commits you want in one PR, even if they do different things. I can always choose to cherry-pick individual commits. But each commit should only do one thing, and each commit should have a proper commit message with an explanation.

PR #2477, opened 2026-05-05
Add support for DiscSubtitle tag

### LINKED ISSUE #2472: Add support for discsubtitle tag
## Feature request

`discsubtitle` is used by collection managers like beets or picard to differentiate individual discs within box-set releases. This is immensely useful for classical / jazz collectors where releases can have hundreds of disks under the same album name. Example:

https://musicbrainz.org/release/bb36d6af-8e22-4bb1-a789-291e9d0a19be

For reference, this is the mediafile implementation:

```python
disctitle = MediaField(
    MP3StorageStyle("TSST"),
    MP4StorageStyle("----:com.apple.iTunes:DISCSUBTITLE"),
    StorageStyle("DISCSUBTITLE"),
    ASFStorageStyle("WM/SetSubTitle"),
)
```

tomushkin:
Closes #2472.

Adds support for [disc subtitle](https://picard-docs.musicbrainz.org/en/latest/appendices/tag_mapping.html#disc-subtitle) tag.

The DiscSubtitle tag complements the Disc tag to label individual discs in box-set releases and is commonly used by auto taggers in large classical / jazz collections.


MaxKellermann:
Your PR contains text that is missing in the commit message.
The link in your PR text doesn't explain the tag, and I have no idea why it is useful and why it needs to be added to MPD's core tag list.

tomushkin:
My bad, the link was meant as a reference on how the tag maps to different formats.

The rationale is as following:

DiscSubtitle is standardly used by auto taggers like picard or beets when syncing metadata from musicbrainz or discogs multi-disc releases. It is essentially a label for the Disc tag.

Without builtin support, a mpd user with a 20+ disc album would have to either: retag the files by copying DiscSubtitle to Album or live with grouping the album by Disc number, which doesn't carry the name.

This is a great QOL improvement when dealing with multi-disc collections.

Consider this one, for example: https://musicbrainz.org/release/45a4786b-65f1-4814-a69f-6bd74776d7bf

It is a single album but most certainly you'd only want to play a single disc from it. Currently with mpd you can only do it by disc number.

Also, as far as I can tell, there's no equivalent tag that can be used for same purpose without clobbering other features.

MaxKellermann:
But what is the practical use of this tag as a core tag for MPD users? How do you actually use it? (In a way it cannot be used using `readcomments`)
What client will support it, and how would the UI look like?

tomushkin:
The client I'm eyeing to get this into is rmpc, which has arbitrary tag browsing WIP. Currently I'm using albumartist / album / disc / tracks, but that's not very useful with 20+ discs and I'm looking forward to replace it with discsubtitle.

Also, the tag can be useful wherever disc is used for displaying (like ncmpcpp's %d song format) to get a human readable label instead of an index.

I feel like it's no different than grouping, except it sits deeper into the hierarchy.

MaxKellermann:
You did not answer my question at all. I still have no idea how this tag will be useful because I have no idea how exactly you will be using it.

tomushkin:
The tag will be useful for browsing my collection. Even mpc benefits from it:

```
$ mpc search album "complete decca philips dg recordings" discsubtitle "schubert: piano sonatas"
```

rather rather than:

```
$ mpc search album "complete decca philips dg recordings" disc 16
```

And here's rmpc browsing by artist / album / discsubtitle:
2026-05-05_kitty-001

Rather than artist / album / disc:
2026-05-05_kitty-002

At least rmpc and ncmpcpp that I know of don't bother with readcomments. Browsing and grouping is limited to the builtin tags mpd provides.

PR #2496, opened 2026-05-20
Use snd_pcm_pause if available

### LINKED ISSUE #1136: Unpause and immediately pause skips a second
### Describe the bug
Why I toggle mpd playback twice in a short time, a second of the song is skipped. This can be most easily reproduced when running the following commands when mpd is paused:`
```sh
mpc play && mpc pause
```

## Expected Behavior
No / very little time is skipped

## Actual Behavior
1 second is skipped

## Version

```
Music Player Daemon 0.22.6 (0.22.6)
Copyright 2003-2007 Warren Dukes <warren.dukes@gmail.com>
Copyright 2008-2018 Max Kellermann <max.kellermann@gmail.com>
This is free software; see the source for copying conditions.  There is NO
warranty; not even MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Database plugins:
 simple proxy upnp

Storage plugins:
 local udisks nfs curl

Neighbor plugins:
 upnp udisks

Decoders plugins:
 [mad] mp3 mp2
 [mpg123] mp3
 [vorbis] ogg oga
 [oggflac] ogg oga
 [flac] flac
 [opus] opus ogg oga
 [sndfile] wav aiff aif au snd paf iff svx sf voc w64 pvf xi htk caf sd2
 [audiofile] wav au aiff aif
 [dsdiff] dff
 [dsf] dsf
 [hybrid_dsd] m4a
 [faad] aac
 [mpcdec] mpc
 [wavpack] wv
 [modplug] 669 amf ams dbm dfm dsm far it med mdl mod mtm mt2 okt s3m stm ult umx xm
 [mikmod] amf dsm far gdm imf it med mod mtm s3m stm stx ult uni xm
 [sidplay] sid mus str prg P00
 [wildmidi] mid
 [fluidsynth] mid
 [ffmpeg] 16sv 3g2 3gp 4xm 8svx aa3 aac ac3 adx afc aif aifc aiff al alaw amr anim apc ape asf atrac au aud avi avm2 avs bap bfi c93 cak cin cmv cpk daud dct divx dts dv dvd dxa eac3 film flac flc fli fll flx flv g726 gsm gxf iss m1v m2v m2t m2ts m4a m4b m4v mad mj2 mjpeg mjpg mka mkv mlp mm mmf mov mp+ mp1 mp2 mp3 mp4 mpc mpeg mpg mpga mpp mpu mve mvi mxf nc nsv nut nuv oga ogm ogv ogx oma ogg omg opus psp pva qcp qt r3d ra ram rl2 rm rmvb roq rpl rvc shn smk snd sol son spx str swf tak tgi tgq tgv thp ts tsp tta xa xvid uv uv2 vb vid vob voc vp6 vmd wav webm wma wmv wsaud wsvga wv wve rtp:// rtsp:// rtsps://
 [gme] ay gbs gym hes kss nsf nsfe sap spc vgm vgz
 [pcm]

Filters:
 libsamplerate soxr

Tag plugins:
 id3tag

Output plugins:
 shout null fifo pipe alsa ao oss openal solaris pulse jack httpd recorder

Encoder plugins:
 null vorbis opus lame twolame wave flac

Archive plugins:
 [bz2] bz2
 [zzip] zip
 [iso] iso

Input plugins:
 file io_uring archive alsa tidal qobuz curl ffmpeg nfs mms cdio_paranoia

Playlist plugins:
 extm3u m3u pls xspf asx rss soundcloud flac cue embcue

Protocols:
 file:// alsa:// cdda:// ftp:// ftps:// gopher:// hls+http:// hls+https:// http:// https:// mms:// mmsh:// mmst:// mmsu:// nfs:// qobuz:// rtmp:// rtmpe:// rtmps:// rtmpt:// rtmpte:// rtmpts:// rtp:// rtsp:// rtsps:// scp:// sftp:// smb:// srtp:// tidal://

Other features:
 avahi dbus udisks epoll icu inotify ipv6 systemd tcp un
```

## Log
Command output
```sh
❯ mpc play && mpc pause
Muse - Uprising
[playing] #22/23   0:02/4:12 (0%)
volume: n/a   repeat: off   random: off   single: off   consume: off
Muse - Uprising
[paused]  #22/23   0:02/4:12 (0%)
volume: n/a   repeat: off   random: off   single: off   consume: off
❯ mpc play && mpc pause
Muse - Uprising
[playing] #22/23   0:03/4:12 (1%)
volume: n/a   repeat: off   random: off   single: off   consume: off
Muse - Uprising
[paused]  #22/23   0:03/4:12 (1%)
volume: n/a   repeat: off   random: off   single: off   consume: off
❯ mpc play && mpc pause
Muse - Uprising
[playing] #22/23   0:04/4:12 (1%)
volume: n/a   repeat: off   random: off   single: off   consume: off
Muse - Uprising
[paused]  #22/23   0:04/4:12 (1%)
volume: n/a   repeat: off   random: off   single: off   consume: off
```

mpd output
```sh
config_file: loading file /home/friso/.config/mpd/mpd.conf
Mar 30 14:37 : exception: bind to '0.0.0.0:6600' failed (continuing anyway, because binding to '[::]:6600' succeeded): Failed to bind socket: Address already in use
Mar 30 14:37 : libsamplerate: libsamplerate converter 'Fastest Sinc Interpolator'
Mar 30 14:37 : vorbis: Xiph.Org libVorbis 1.3.7
Mar 30 14:37 : opus: libopus 1.3.1
Mar 30 14:37 : sndfile: libsndfile-1.0.31
Mar 30 14:37 : hybrid_dsd: The Hybrid DSD decoder is disabled because it was not explicitly enabled
Mar 30 14:37 : exception: Decoder plugin 'wildmidi' is unavailable: configuration file does not exist: /etc/timidity/timidity.cfg
Mar 30 14:37 : simple_db: reading DB
Mar 30 14:37 : output: No 'audio_output' defined in config file
Mar 30 14:37 : output: Attempt to detect audio output device
Mar 30 14:37 : output: Attempting to detect a alsa audio device
Mar 30 14:37 : output: Successfully detected a alsa audio device
Failed to initialize io_uring: io_uring_queue_init() failed: Cannot allocate memory
Mar 30 14:37 : exception: Input plugin 'tidal' is not configured: No Tidal application token configured
Mar 30 14:37 : exception: Input plugin 'qobuz' is not configured: No Qobuz app_id configured
Mar 30 14:37 : curl: version 7.75.0
Mar 30 14:37 : curl: with OpenSSL/1.1.1k
```

Mikachu:
This should fix #1136 but note I have only tried this on hardware that does support pause (quite old pci soundblaster live). It is also quite possible I do silly things in the patch, I'm not very familiar with the codebase, but pausing/unpausing works, stopping/playing while paused or not paused works, with and without the always_on and/or close_on_pause options (the latter should fully disable hardware pausing and revert to the old behaviour, and it's on by default). It didn't seem worth adding another option explicitly to disable hardware pause since this option already exists, but I don't have whatever DSD is, so maybe I do break that usecase if that needs silence streamed actively?

### DIFF
--- a/src/output/Interface.hxx
+++ b/src/output/Interface.hxx
@@ -40,12 +40,22 @@ public:
 		return flags & FLAG_PAUSE;
 	}
 
-	virtual bool SupportsPauseWithoutCancel() const noexcept {
-		return false;
-	}
-
 	bool GetNeedFullyDefinedAudioFormat() const noexcept {
 		return flags & FLAG_NEED_FULLY_DEFINED_AUDIO_FORMAT;
+	}
+
+	/**
+	 * Returns true if this plugin can pause playback without first
+	 * discarding buffered audio via Cancel().  When this returns true,
+	 * #FilteredAudioOutput::BeginPause() will skip the Cancel() call,
+	 * allowing the plugin to preserve its hardware buffer across a pause
+	 * and resume it intact when Play() is next called.
+	 *
+	 * Plugins that return true must handle buffer cleanup if
+	 * Cancel() is called while paused (e.g. on stop or seek).
+	 */
+	virtual bool SupportsHardwarePause() const noexcept {
+		return false;
 	}
 
 	/**
--- a/src/output/Interface.hxx
+++ b/src/output/Interface.hxx
@@ -40,6 +40,10 @@ public:
 		return flags & FLAG_PAUSE;
 	}
 
+	virtual bool SupportsPauseWithoutCancel() const noexcept {
+		return false;
+	}
+
 	bool GetNeedFullyDefinedAudioFormat() const noexcept {
 		return flags & FLAG_NEED_FULLY_DEFINED_AUDIO_FORMAT;
 	}
--- a/src/output/Thread.cxx
+++ b/src/output/Thread.cxx
@@ -519,6 +519,10 @@ AudioOutputControl::Task() noexcept
 				   the actual playback */
 				if (source_state == SourceState::OPEN)
 					source.Cancel();
+				{
+					const ScopeUnlock unlock(mutex);
+					output->Cancel();
+				}
 				InternalPause(lock);
 			} else {
 				InternalClose(false);
--- a/src/output/plugins/AlsaOutputPlugin.cxx
+++ b/src/output/plugins/AlsaOutputPlugin.cxx
@@ -232,6 +232,7 @@ class AlsaOutput final
 	const bool close_on_pause;
 
 	std::atomic_bool paused;
+	bool hw_can_pause = false;
 
 public:
 	AlsaOutput(EventLoop &loop, const ConfigBlock &block);
@@ -398,6 +399,11 @@ private:
 	/* virtual methods from class MultiSocketMonitor */
 	Event::Duration PrepareSockets() noexcept override;
 	void DispatchSockets() noexcept override;
+
+	bool SupportsPauseWithoutCancel() const noexcept override {
+		return !close_on_pause && hw_can_pause;
+	}
+
 };
 
 static constexpr Domain alsa_output_domain("alsa_output");
@@ -562,6 +568,8 @@ AlsaOutput::Setup(AudioFormat &audio_format,
 
 	AlsaSetupSw(pcm, hw_result.buffer_size - hw_result.period_size,
 		    hw_result.period_size);
+
+	hw_can_pause = hw_result.can_pause;
 
 	auto alsa_period_size = hw_result.period_size;
 	if (alsa_period_size == 0)
@@ -1095,6 +1103,7 @@ AlsaOutput::CancelInternal() noexcept
 	ring_buffer.Clear();
 
 	active = false;
+	paused = false;
 	waiting = false;
 
 	UnregisterSockets();
@@ -1139,13 +1148,46 @@ AlsaOutput::Cancel() noexcept
 bool
 AlsaOutput::Pause() noexcept
 {
-	std::lock_guard lock{mutex};
-	interrupted = false;
+	{
+		std::lock_guard lock{mutex};
+		interrupted = false;
 
-	if (close_on_pause)
-		return false;
+		if (close_on_pause)
+			return false;
 
-	// TODO use snd_pcm_pause()?
+		if (!hw_can_pause) {
+			paused = true;
+			return true;
+		}
+	}
+
+	if (LockIsActive()) {
+		BlockingCall(GetEventLoop(), [this](){
+			/* only pause if actually running; if the PCM is
+			   still in PREPARED state, there is nothing in
+			   the hardware buffer worth preserving */
+			if (snd_pcm_state(pcm) == SND_PCM_STATE_RUNNING) {
+				int err = snd_pcm_pause(pcm, /* enable */ 1);
+				if (err < 0) {
+					LogError(alsa_output_domain,
+							 "snd_pcm_pause() failed");
+					hw_can_pause = false;
+					return;
+				}
+			}
+
+			UnregisterSockets();
+			silence_timer.Cancel();
+
+			/* set waiting=true so that Activate() inside
+			   Play() will re-schedule the event loop on
+			   resume; without this, Activate() sees
+			   active=true/waiting=false and returns early */
+			waiting = true;
+		});
+	}
+	if (!hw_can_pause)
+		return Pause(); // try again without hw pause
 
 	paused = true;
 	return true;
@@ -1215,7 +1257,12 @@ AlsaOutput::Play(std::span<const std::byte> src)
 	assert(!src.empty());
 	assert(src.size() % in_frame_size == 0);
 
-	paused = false;
+	const bool was_paused = paused.exchange(false);
+
+	if (was_paused) {
+		std::lock_guard lock{mutex};
+		Activate();
+	}
 
 	const size_t max_frames = LockWaitWriteAvailable();
 	const size_t max_size = max_frames * in_frame_size;
@@ -1242,6 +1289,15 @@ AlsaOutput::PrepareSockets() noexcept
 	}
 
 	try {
+		/* if the PCM was paused via snd_pcm_pause() (i.e. Pause()
+		   was called with hw_can_pause), resume it now that Play()
+		   has delivered new data and re-activated the I/O thread */
+		if (snd_pcm_state(pcm) == SND_PCM_STATE_PAUSED) {
+			int err = snd_pcm_pause(pcm, /* disable */ 0);
+			if (err < 0)
+				throw Alsa::MakeError(err, "snd_pcm_pause() failed");
+		}
+
 		return non_block.PrepareSockets(*this, pcm);
 	} catch (...) {
 		ClearSocketList();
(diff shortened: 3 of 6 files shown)


src/output/Interface.hxx
```diff
@@ -40,6 +40,10 @@ public:
 		return flags & FLAG_PAUSE;
 	}
 
+	virtual bool SupportsPauseWithoutCancel() const noexcept {
```
MaxKellermann: That's a clunky method name, but lacking an idea for a better name, that's okay.
But please add some API documentation explaining what it means.

src/output/Thread.cxx
```diff
@@ -519,6 +519,10 @@ AudioOutputControl::Task() noexcept
 				   the actual playback */
 				if (source_state == SourceState::OPEN)
 					source.Cancel();
+				{
```
MaxKellermann, line 522: I can only guess why this is necessary - this is worth a comment explaining it.
But it's only really necessary if the output was paused-without-cancel? Can we skip the call on outputs that don't do that? And only do it if the output is really paused?

src/output/plugins/AlsaOutputPlugin.cxx
```diff
@@ -232,6 +232,7 @@ class AlsaOutput final
 	const bool close_on_pause;
 
 	std::atomic_bool paused;
+	bool hw_can_pause = false;
```
MaxKellermann, line 235: How is cross-thread access to this field protected?

src/output/plugins/AlsaOutputPlugin.cxx
```diff
@@ -1095,6 +1103,7 @@ AlsaOutput::CancelInternal() noexcept
 	ring_buffer.Clear();
 
 	active = false;
+	paused = false;
```
MaxKellermann: Why this?

src/output/plugins/AlsaOutputPlugin.cxx
```diff
@@ -1139,13 +1148,46 @@ AlsaOutput::Cancel() noexcept
...
+
+			/* set waiting=true so that Activate() inside
+			   Play() will re-schedule the event loop on
+			   resume; without this, Activate() sees
+			   active=true/waiting=false and returns early */
+			waiting = true;
+		});
+	}
+	if (!hw_can_pause)
+		return Pause(); // try again without hw pause
```
MaxKellermann, line 1189: Please avoid recursive self-calls. This smells badly.
What's the point of this call, anyway?

MaxKellermann (changes requested):
Thanks for the work, at a first glance, I like it. I need to take some more time to verify it and of course test it.
I have a few questions and comments.

src/output/Thread.cxx
Mikachu: Yes, this is needed to drop output if we paused, and then for example decide to stop and switch to another track. We could skip it, but then we need to add some extra boilerplate, something like the below, so that we can check if this is supported on the output. The double Cancel() call is a no-op but it is an extra blocking call, idk how much that matters? I can go either way.

```diff
diff --git i/src/output/Filtered.cxx w/src/output/Filtered.cxx
index b5e41e98e7..693783bded 100644
--- i/src/output/Filtered.cxx
+++ w/src/output/Filtered.cxx
@@ -33,12 +33,6 @@ FilteredAudioOutput::SupportsPause() const noexcept
        return output->SupportsPause();
 }

+bool
+FilteredAudioOutput::SupportsPauseWithoutCancel() const noexcept
+{
+       return output->SupportsPauseWithoutCancel();
+}
+
 std::map<std::string, std::string, std::less<>>
 FilteredAudioOutput::GetAttributes() const noexcept
 {
diff --git i/src/output/Filtered.hxx w/src/output/Filtered.hxx
index b9e3bb68f7..e9c4b27806 100644
--- i/src/output/Filtered.hxx
+++ w/src/output/Filtered.hxx
@@ -155,9 +155,6 @@ public:
        [[gnu::pure]]
        bool SupportsPause() const noexcept;

+       [[gnu::pure]]
+       bool SupportsPauseWithoutCancel() const noexcept;
+
        std::map<std::string, std::string, std::less<>> GetAttributes() const noexcept;
        void SetAttribute(std::string &&name, std::string &&value);
```

src/output/plugins/AlsaOutputPlugin.cxx
Mikachu: The idea here was that if somehow the snd_pcm_pause call above fails even though the driver said yes to can_pause(), we'll disable hw_can_pause and pause the old fashioned way. This recursive call would only ever happen once per session, but maybe it's a bit overly paranoid? Plus as you mentioned in the other comment, it might require protecting hw_can_pause with some lock, when we otherwise don't need it (it stays constant after the hw init, and I'm assuming Setup() is not run in parallel with playback :) ). I can remove this whole fallback thing if it's too silly?

src/output/Interface.hxx
Mikachu: Another idea is maybe SupportsHardwarePause() but I'm not sure how applicable that might be to other plugins.

	/**
	 * Returns true if this plugin can pause playback without first
	 * discarding buffered audio via Cancel().  When this returns true,
	 * #FilteredAudioOutput::BeginPause() will skip the Cancel() call,
	 * allowing the plugin to preserve its hardware buffer across a pause
	 * and resume it intact when Play() is next called.
	 *
	 * Plugins that return true must handle buffer cleanup if
	 * Cancel() is called while paused (e.g. on stop or seek).
	 */

does this make sense for an api doc comment? (I've literally only looked at the alsa output plugin, so i'm not exactly sure in which directions things generalize in mpd).

src/output/plugins/AlsaOutputPlugin.cxx
Mikachu: I think this was just me being a bit confused about things at one point, and then didn't realize it wasn't actually needed. I'll remove it.

src/output/plugins/AlsaOutputPlugin.cxx
Mikachu: (see other reply about return Pause()).

Mikachu:
(btw, i'll obviously squash those commits later, but it's easier for me to track what is changed because of review comments like this for now).

src/output/Thread.cxx
MaxKellermann: Even if it's effectively a no-op, it adds overhead, so we shouldn't do any calls when we already know it's a no-op.

src/output/plugins/AlsaOutputPlugin.cxx
MaxKellermann: If you believe there needs to be a fallback, then there needs to be a fallback. Don't remove it. But implement it without Pause() calling itself recursively.
If you're brave, you can update hw_can_pause without a lock, if you believe it's safe (enough). You can do anything if you know what you're doing. I'm not going to lawyer about 100% strict thread-safety; sometimes we can cut corners. But that consideration should be documented.

MaxKellermann:
Folded & merged: 49cb3b837f0713e329cc53265d139343cb1998d9

PR #2505, opened 2026-06-11
config: Prevent infinite recursion in include directives

hyder365:
## Issue

The MPD config file parser supports an "include" directive to load additional config files. The ReadConfigFile() function calls itself recursively for each included file, but has no mechanism to detect recursive includes. A config file that includes itself (directly or transitively through other files) will cause infinite recursion, eventually overflowing the stack and crashing MPD with a segfault.

## Solution

The patch adds a "depth" parameter to the internal ReadConfigFile() function. Depth is incremented on each recursive include call. If depth exceeds 16 (MAX_INCLUDE_DEPTH), a std::runtime_error is thrown with the message "Include recursion too deep".

The header src/config/File.hxx is also modified to add the depth parameter to the public ReadConfigFile() signature, defaulting to 0 for backward compatibility with existing callers.

Both "include" and "include_optional" paths pass depth + 1 to recursive calls, ensuring cycles are detected at any depth.

### DIFF
--- a/src/config/File.cxx
+++ b/src/config/File.cxx
@@ -159,10 +159,17 @@ ReadConfigParam(ConfigData &config_data, BufferedReader &reader,
 
 /**
  * @param directory the directory used to resolve relative paths
+ * @param depth the current include recursion depth (to detect cycles)
  */
 static void
-ReadConfigFile(ConfigData &config_data, BufferedReader &reader, Path directory)
+ReadConfigFile(ConfigData &config_data, BufferedReader &reader, Path directory,
+	       unsigned depth = 0)
 {
+	static constexpr unsigned MAX_INCLUDE_DEPTH = 16;
+
+	if (depth > MAX_INCLUDE_DEPTH)
+		throw std::runtime_error("Include recursion too deep");
+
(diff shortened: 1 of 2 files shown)


src/config/File.cxx
```diff
@@ -180,12 +187,10 @@ ReadConfigFile(ConfigData &config_data, BufferedReader &reader, Path directory)
 		assert(name != nullptr);
 
 		if (StringIsEqual(name, "include")) {
-			// TODO: detect recursion
-			// TODO: Config{Block,Param} have only line number but no file name
```
MaxKellermann, line 184: Explain this change

src/config/File.cxx
```diff
@@ -159,10 +159,17 @@ ReadConfigParam(ConfigData &config_data, BufferedReader &reader,
 
 /**
  * @param directory the directory used to resolve relative paths
+ * @param depth the current include recursion depth (to detect cycles)
```
MaxKellermann, line 162: How does the depth detect cycles?

MaxKellermann:
Commit message is bad here, too.

PR #2510, opened 2026-06-16
fmt 12.2.0 compatibilty

TimB87:
The [api](https://fmt.dev/12.0/api/) for fmt 12.2.0 changed

### DIFF
--- a/src/Mapper.cxx
+++ b/src/Mapper.cxx
@@ -17,7 +17,7 @@
 #include "Main.hxx"
 #endif
 
-#include <fmt/core.h>
+#include <fmt/format.h>
 
 #include <cassert>
 
--- a/src/input/plugins/AlsaInputPlugin.cxx
+++ b/src/input/plugins/AlsaInputPlugin.cxx
@@ -27,7 +27,7 @@
 
 #include <alsa/asoundlib.h>
 
-#include <fmt/core.h>
+#include <fmt/format.h>
 
 #include <cassert>
 
--- a/src/input/plugins/QobuzClient.cxx
+++ b/src/input/plugins/QobuzClient.cxx
@@ -5,7 +5,7 @@
 #include "lib/crypto/MD5.hxx"
 #include "thread/ScopeUnlock.hxx"
 
-#include <fmt/core.h>
+#include <fmt/format.h>
 
 #include <cassert>
 #include <stdexcept>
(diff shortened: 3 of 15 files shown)


MaxKellermann:
Your PR contains text that is missing in the commit message, and even that doesn't explain the API change.
You linked to the fmt 12.0.0 API documentation, but you claim this is about 12.2.0.

From the commit diff I can guess that `fmt/core.h` was removed, but is that really true? Then why does your commit change only half of the `#include` lines? I don't get it.

TimB87:
Hi Max,
sorry, I shouldn't have opened that this late.

I saw this compile issue with fmt 12.2.0:
```
../mpd-0.24.12/src/net/ToString.cxx: In function 'std::string ToString(SocketAddress)':
../mpd-0.24.12/src/net/ToString.cxx:80:37: error: 'format' is not a member of 'fmt'
   80 |                         return fmt::format("[{}]:{}", host, serv);
      |                                     ^~~~~~
../mpd-0.24.12/src/net/ToString.cxx:84:29: error: 'format' is not a member of 'fmt'
   84 |                 return fmt::format("{}:{}", host, serv);
      |                             ^~~~~~
[137/515] Compiling C++ object src/pcm/libpcm.a.p/ChannelsConverter.cxx.o
/tmp/distccd_ce931a95.ii:1:3: warning: style of line directive is a GCC extension
    1 | # 0 "../mpd-0.24.12/src/pcm/ChannelsConverter.cxx"
      |   ^
../mpd-0.24.12/src/pcm/ChannelsConverter.cxx: warning: style of line directive is a GCC extension
```

Please ignore distcc o:) Anyway, this led me to check the [changelog](https://github.com/fmtlib/fmt/blob/main/ChangeLog.md) which had the clue how to fix this:

> Made the <fmt/core.h> header equivalent to <fmt/base.h> by default. Code that relied on <fmt/core.h> pulling in <fmt/format.h> must now either include <fmt/format.h> directly or define FMT_DEPRECATED_HEAVY_CORE to opt back in.

It's possible that more need the change, its just the code paths I was able to trigger over here.

MaxKellermann:
The explanation belongs in the commit message.

MaxKellermann:
Merged https://github.com/MusicPlayerDaemon/MPD/pull/2512 instead of this one because it had a better commit message.

PR #2512, opened 2026-06-18
libfmt 12.2.0 support: use fmt/format.h not fmt/core.h

heitbaum:
Since libfmt 12.2.0, fmt/core.h is a shim that only includes fmt/base.h which does not provide fmt::format(). Include fmt/format.h directly in files that call fmt::format().

Fixes build with libfmt-12.2.0

https://github.com/fmtlib/fmt/releases/tag/12.2.0


MaxKellermann:
Backported to v0.24.x: 88d5b5248d71862815da336cfd1c6f66626bb4d4

PR #2515, opened 2026-06-20
protocol: validate relative paths

hyder365:
VerifyRelativePathUTF8() only checked that the path was valid UTF-8. Make it reject absolute paths, empty path components, and . or .. components.

This prevents archive entries and other relative paths from registering traversal-style names.


MaxKellermann:

PR #2516, opened 2026-06-20
tag: limit ID3 tag size

hyder365:
Apply the existing 4 MiB ID3 tag size limit to direct ID3v2 tag loading as well.

This avoids allocating very large buffers from the size declared in the ID3 header before reading the rest of the tag.


MaxKellermann:

PR #2517, opened 2026-06-20
input/curl: guard content length

hyder365:
Parse Content-Length more strictly and avoid wrapping when adding it to the current stream offset.

If the header is malformed or the sum would overflow, leave the stream size unknown instead of storing a wrapped value.


MaxKellermann:

PR #2518, opened 2026-06-20
decoder/mad: fix ID3 tag query

hyder365:
Fix the fallback ID3 tag query used when libid3tag is disabled.

Use unsigned bytes, reject non-synchsafe size bytes, and parse the full four-byte ID3v2 size field.


MaxKellermann:

PR #2519, opened 2026-06-20
decoder/ffmpeg: free AVIO buffer

hyder365:
Free the buffer allocated for avio_alloc_context() if creating the AVIOContext fails.

Without this cleanup, that failure path leaks the 8192-byte buffer.


MaxKellermann:

PR #2525, opened 2026-06-25
systemd: allow AF_NETLINK when upnp is enabled

abouvier:
`AF_NETLINK` is also needed for the upnp database plugin: https://bbs.archlinux.org/viewtopic.php?id=310134

### DIFF
--- a/systemd/meson.build
+++ b/systemd/meson.build
@@ -2,7 +2,7 @@ systemd_unit_conf = configuration_data()
 systemd_unit_conf.set('prefix', get_option('prefix'))
 
 address_families = ['AF_INET', 'AF_INET6', 'AF_UNIX']
-if get_option('smbclient').enabled()
+if get_option('smbclient').enabled() or conf.get('ENABLE_UPNP')
   # AF_NETLINK is required by libsmbclient, or it will exit() .. *sigh*
   address_families += 'AF_NETLINK'
 endif


MaxKellermann:
Bad commit message

abouvier:
Where is the guideline for good commit message? :p

MaxKellermann:
Rule of thumb: if the PR contains text that is missing in your commit message, then your commit message must be obviously bad.

Second rule of thumb: if your commit message describes only what is obvious enough by looking at the diff, but doesn't even try to explain why, it's bad.

MaxKellermann:
Your commit message says "when upnp is enabled" but that's not what your actual code does - it checks whether the feature is not disabled. That's different from the existing smbclient check, and that has a bad smell. That needs to be explained.

abouvier:
Because the `smbclient` option is a [boolean](https://github.com/MusicPlayerDaemon/MPD/blob/98c06a920d25390c84941654b1a0a24ddf6e7930/meson_options.txt#L104), but the `upnp` option is a [combo](https://github.com/MusicPlayerDaemon/MPD/blob/98c06a920d25390c84941654b1a0a24ddf6e7930/meson_options.txt#L67-L70) with multiple ways to be enabled, so checking if it's not disabled is simpler.

MaxKellermann:
> Because the `smbclient` option is a [boolean](https://github.com/MusicPlayerDaemon/MPD/blob/98c06a920d25390c84941654b1a0a24ddf6e7930/meson_options.txt#L104),

This is wrong. It's a `feature`.

MaxKellermann:
Explanation still missing.

abouvier:
`get_option('upnp').enabled()` is not possible because this option is not a feature but a combo, so the value returned by `get_option` will be a string.

Besides, multiple values of this option can end up enabling the upnp plugin, with the logic to determine that in [src/lib/upnp/meson.build](https://github.com/MusicPlayerDaemon/MPD/blob/master/src/lib/upnp/meson.build). So I guess that the best way to verify for sure that upnp is enabled is to check the configuration variable set in this file.

MaxKellermann:
Backported to v0.24.x: 633d8d209dccc53dc09044e75cebfbf3bfc3c671

PR #2537, opened 2026-07-12
win32: use InjectBreak() for cross-thread shutdown

0x0003:
`Break()` only sets a flag without waking `select()`, so the console handler thread and service dispatcher thread can never actually stop the event loop.  Use `InjectBreak()` instead, which writes to the wake pipe to interrupt `select()`.

Without this, ^c in a console and `net stop mpd` both fail to shut down MPD gracefully - the process must be hard-killed, losing the state file.

### DIFF
--- a/src/win32/Win32Main.cxx
+++ b/src/win32/Win32Main.cxx
@@ -89,7 +89,7 @@ console_handler(DWORD event)
 			// regardless our thread is still active.
 			// If this did not happen within 3 seconds
 			// let's shutdown anyway.
-			global_instance->Break();
+			global_instance->event_loop.InjectBreak();
 			// Under debugger it's better to wait indefinitely
 			// to allow debugging of shutdown code.
 			Sleep(IsDebuggerPresent() ? INFINITE : 3000);


src/Instance.hxx
```diff
@@ -146,7 +146,7 @@ struct Instance final
 	 * Wrapper for EventLoop::Break().  Call to initiate shutdown.
 	 */
 	void Break() noexcept {
-		event_loop.Break();
+		event_loop.InjectBreak();
```
MaxKellermann: Now the API documentation is wrong. And I don't think it's a good idea to have a Break() method that calls InjectBreak() and not Break() - it's error prone.

MaxKellermann (changes requested):

0x0003:
Dropped src/Instance.hxx change, it never fires on windows anyway ~(I think)~.

MaxKellermann:

MaxKellermann:
Backported to v0.24.x: f70785ad0546b8531ff6e970d91d097dcfe7f05a

PR #2539, opened 2026-07-12
db/update: truncate mtime to seconds during song db checks

0x0003:
FILETIME has 100ns precision but the database stores mtimes as time_t (seconds).  After restart, the comparison between the full-precision FILETIME value and the second-precision DB value always mismatches, forcing a full rescan on every startup.

### DIFF
--- a/src/db/update/UpdateSong.cxx
+++ b/src/db/update/UpdateSong.cxx
@@ -14,6 +14,15 @@
 
 #include <unistd.h>
 
+/* Truncate to seconds to match the DB round-trip through time_t */
+static inline bool
+CompareMtimeCoarse(std::chrono::system_clock::time_point info_mtime,
+		 std::chrono::system_clock::time_point song_mtime) noexcept
+{
+	return std::chrono::system_clock::to_time_t(info_mtime) ==
+	       std::chrono::system_clock::to_time_t(song_mtime);
+}
+
 inline void
 UpdateWalk::UpdateSongFile2(Directory &directory,
 			    std::string_view name, std::string_view suffix,
@@ -32,7 +41,9 @@ try {
 		return;
 	}
 
-	if (!(song != nullptr && info.mtime == song->mtime && !walk_discard) &&
+	if (!(song != nullptr &&
+	      CompareMtimeCoarse(info.mtime, song->mtime) &&
+	      !walk_discard) &&
 	    UpdateContainerFile(directory, name, suffix, info)) {
 		return;
 	}
@@ -61,7 +72,8 @@ try {
 		modified = true;
 		FmtNotice(update_domain, "added {}/{}",
 			  directory.GetPath(), name);
-	} else if (info.mtime != song->mtime || walk_discard) {
+	} else if (!CompareMtimeCoarse(info.mtime, song->mtime) ||
+		   walk_discard) {
 		FmtNotice(update_domain, "updating {}/{}",
 			  directory.GetPath(), name);
 		if (song->UpdateFile(storage, info))


MaxKellermann:
I don't think it's a good idea to tackle this at this lowest level. The truncation should bette be done where it's really buggy.

0x0003:
Could truncate at the points where mtime enters the database, but that'd mean introducing changes to a bunch of generic files.

A new function like this in ChronoUtil.hxx
```hxx
inline std::chrono::system_clock::time_point
TruncateToSeconds(std::chrono::system_clock::time_point tp)
{
	return std::chrono::system_clock::from_time_t(
		std::chrono::system_clock::to_time_t(tp));
}
```

Then e.g. in SimpleDatabasePlugin.cxx
```diff
diff --git a/src/db/plugins/simple/SimpleDatabasePlugin.cxx b/src/db/plugins/simple/SimpleDatabasePlugin.cxx
index be3ab57e6..f2ec718fe 100644
--- a/src/db/plugins/simple/SimpleDatabasePlugin.cxx
+++ b/src/db/plugins/simple/SimpleDatabasePlugin.cxx
@@ -29,6 +29,7 @@
 #include "util/Domain.hxx"
 #include "util/RecursiveMap.hxx"
 #include "Log.hxx"
+#include "time/ChronoUtil.hxx"
 
 #ifdef ENABLE_ZLIB
 #include "lib/zlib/GzipOutputStream.hxx"
@@ -143,7 +144,7 @@ SimpleDatabase::Load()
 
 	FileInfo fi;
 	if (GetFileInfo(path, fi))
-		mtime = fi.GetModificationTime();
+		mtime = TruncateToSeconds(fi.GetModificationTime());
 }
 
 void
@@ -373,7 +374,7 @@ SimpleDatabase::Save()
 
 	FileInfo fi;
 	if (GetFileInfo(path, fi))
-		mtime = fi.GetModificationTime();
+		mtime = TruncateToSeconds(fi.GetModificationTime());
 }
 
 void
```
That's a bit wasteful for platforms that aren't windows, but probably the most "correct" way I can think of solving this.

Your call.

---

For the record, I've been running a build with this PR for half a day now and haven't run into any problems.

MaxKellermann:
When you fix mistakes in an unmerged commit, amend the bad commit instead of adding fixup commits. It's very hard to review fixup commits.
Yet it looks like you're doing the truncation still on the wrong layer.

0x0003:
I'm not 100% sure what you mean by "wrong layer".
Do you want to leave database writes entierly unchanged, and only apply truncation during checks? I honestly don't think that a correct approach, but we could do that.
Or is there something wrong with the helper function?

MaxKellermann:
The actual time stamp comparison is done in src/db/update/UpdateSong.cxx, and we know that `song->mtime` may have been truncated, and so this comparison should allow truncated values. This is the place where you should fix this, and not truncate everything in the generic filesystem/storage layer.

MaxKellermann:
Does this really work? You truncated only the current time stamp, not the one stored in the song. What if you do two consecutive database updates without reloading the DB file by restarting MPD? I belive this PR will make this Windows-only bug available all operating systems, instead of fixing it.

0x0003:
The truncation should be no-op on other systems, I don't see how it'd make the bug available on them.

src/db/update/UpdateSong.cxx
```diff
@@ -32,7 +32,15 @@ try {
...
 	}
 
-	if (!(song != nullptr && info.mtime == song->mtime && !walk_discard) &&
+	/* truncate to seconds to match the DB round-trip through time_t */
+	const auto info_mtime = std::chrono::system_clock::from_time_t(
+		std::chrono::system_clock::to_time_t(info.mtime));
+	const auto song_mtime = song != nullptr
+		? std::chrono::system_clock::from_time_t(
+			std::chrono::system_clock::to_time_t(song->mtime))
+		: std::chrono::system_clock::time_point{};
```
MaxKellermann: I don't like this initializer. The whole use of ternary operator to produce a bogus fallback value makes the code very complicated to read. Bogus values should not exist at all.

I'd prefer some kind of function like:
```
static constexpr bool
CompareMtimeCoarse(a, b) noexcept ....
```
.. and use that in the two relevant lines.

MaxKellermann:
> The truncation should be no-op on other systems, I don't see how it'd make the bug available on them.

Until somebody uses the nanosecond-precision time stamps that are available everywhere (but are currently ignored by MPD). Then suddenly your sleeping bug will become active, and will look like the new guy's bug, but it's really yours.

Please don't get into this kind of argument with me. It leads nowhere. Your code was buggy, period.

MaxKellermann:
Backported to v0.24.x: 208a6b2abb69dde5c94a96adebc16001e1bb4569

PR #2540, opened 2026-07-13
storage: bound WebDAV PROPFIND responses

acts-1631:
Limit WebDAV PROPFIND response bodies, href values, and collected directory entries.

A configured WebDAV server could send an unbounded href value in a 207 response, causing MPD to append data until memory is exhausted. The new limits abort the request before further allocation.


MaxKellermann:
Backported to v0.24.x: 9eb6cafcdb7828798a71aca0a2f45f07fcb8cc36

PR #2541, opened 2026-07-13
qobuz: require HTTPS for API requests

acts-1631:
The Qobuz input plugin defaults to an HTTP API URL and includes the configured password in the login URL query. An on-path attacker can read the credentials and login response.\n\nUse the HTTPS endpoint by default and reject non-HTTPS base_url values.

### DIFF
--- a/src/input/plugins/QobuzInputPlugin.cxx
+++ b/src/input/plugins/QobuzInputPlugin.cxx
@@ -115,7 +115,9 @@ InitQobuzInput(EventLoop &event_loop, const ConfigBlock &block)
 	GlobalInitMD5();
 
 	const char *base_url = block.GetBlockValue("base_url",
-						   "http://www.qobuz.com/api.json/0.2/");
+						   "https://www.qobuz.com/api.json/0.2/");
+	if (!StringStartsWithIgnoreCase(base_url, "https://"sv))
+		throw PluginUnconfigured("Qobuz base_url must use HTTPS");
 
 	const char *app_id = block.GetBlockValue("app_id");
 	if (app_id == nullptr)


src/input/plugins/QobuzInputPlugin.cxx
```diff
@@ -115,7 +115,9 @@ InitQobuzInput(EventLoop &event_loop, const ConfigBlock &block)
 	GlobalInitMD5();
 
 	const char *base_url = block.GetBlockValue("base_url",
-						   "http://www.qobuz.com/api.json/0.2/");
+						   "https://www.qobuz.com/api.json/0.2/");
+	if (!StringStartsWithIgnoreCase(base_url, "https://"sv))
+		throw PluginUnconfigured("Qobuz base_url must use HTTPS");
```
MaxKellermann, line 120: Why this? If the user insists, who are we to forbid this? It's unnecessary code.

MaxKellermann (changes requested):

src/input/plugins/QobuzInputPlugin.cxx
acts-1631: Is there any legitimate reason a user would want their creds sent in plaintext? I just meant this to prevent a "foot gun."

MaxKellermann:
Backported to v0.24.x but without the enforcement code: be2d9103c062f99e8dd2cd9ec8b8ee91e75f0fab

acts-1631:
Alright, good enough for me :) thanks

src/input/plugins/QobuzInputPlugin.cxx
MaxKellermann: You are replying to a question with another question, and that's not a valid argument. But I'll bite anyway: what if they want to debug their own Qobuz protocol implementation?
Your check serves no real purpose. Unnecessary code should not exist (in this project).

PR #2544, opened 2026-07-14
python/build/libs.py: add mpg123 1.33.6

0x0003:
Adds mpg123 as a third-party library, giving the cross-compiled build a dedicated MP3 decoder instead of falling back on ffmpeg.

The main reason to add it is to work around a pretty old ffmpeg bug where id3v2.4 multi-value tags (e.g. multiple genres, artists) are truncated to only the first value: https://trac.ffmpeg.org/ticket/6949

As a bonus, it is also **significantly** faster for library scanning, nearly 20x on 25k MP3 files compared to ffmpeg, tested on NTFS.

### DIFF
--- a/python/build/libs.py
+++ b/python/build/libs.py
@@ -562,6 +562,17 @@ ffmpeg = FfmpegProject(
     ],
 )
 
+libmpg123 = AutotoolsProject(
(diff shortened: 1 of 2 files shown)


MaxKellermann:
Backported to v0.24.x: d3c39952cfb432376d87531cc45616a1af9d905d

PR #2545, opened 2026-07-15
wasapi: prevent audio pops on startup, resume, and seek

0x0003:
Fixes annoying audio pops in the wasapi output caused by sample waveform discontinuities. Most of them, at least. The remaining ones can only be masked by something like fade in/out, which MPD doesn't have.

### DIFF
--- a/src/output/plugins/wasapi/WasapiOutputPlugin.cxx
+++ b/src/output/plugins/wasapi/WasapiOutputPlugin.cxx
@@ -218,6 +218,14 @@ public:
 		std::size_t consumed = ring_buffer.WriteFrom(input);
 
 		if (!playing) {
+			/* Wait for enough data to fill the entire
+			   endpoint buffer before starting playback to
+			   prevent a pop on cold start.
+			   https://learn.microsoft.com/en-us/windows/win32/coreaudio/rendering-a-stream */
+			if (ring_buffer.ReadAvailable() <
+			    buffer_size_in_frames * frame_size) {
+				return consumed;
+			}
 			playing = true;
 			Play();
 		}
--- a/src/output/plugins/wasapi/WasapiOutputPlugin.cxx
+++ b/src/output/plugins/wasapi/WasapiOutputPlugin.cxx
@@ -430,10 +430,19 @@ try {
 		event.Wait();
 
 		if (cancel.load()) {
+			/* Stop the client to prevent playing stale data
+			   or silence after discard. Push() will restart
+			   once the ring buffer is full.
+			   https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-start */
+			if (started && !is_exclusive) {
+				Stop(client);
+				started = false;
+			}
 			ring_buffer.Discard();
 			cancel.store(false);
 			empty.store(true);
 			InterruptWaiter();
+			continue;
 		}
 
 		Status current_state = status.load();
--- a/src/output/plugins/wasapi/WasapiOutputPlugin.cxx
+++ b/src/output/plugins/wasapi/WasapiOutputPlugin.cxx
@@ -454,6 +454,15 @@ try {
 
 		UINT32 write_in_frames = buffer_size_in_frames;
 		DWORD mode = 0;
+
+		/* Flush any pending data left in the endpoint buffer
+		   after Stop() and reset the stream position to 0.
+		   Without this, Start() would play stale audio.
+		   https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-reset */
+		if (!started) {
+			Reset(client);
+		}
+
 		AtScopeExit(&) {
 			render_client->ReleaseBuffer(write_in_frames, mode);
 
--- a/src/output/plugins/wasapi/WasapiOutputPlugin.cxx
+++ b/src/output/plugins/wasapi/WasapiOutputPlugin.cxx
@@ -144,6 +144,14 @@ SetDSDFallback(AudioFormat &audio_format) noexcept
 
 } // namespace
 
+/**
+ * Renders audio from a ring buffer into a WASAPI endpoint buffer.
+ * To avoid pops, the client is started only when the endpoint buffer
+ * can be filled completely (see Push()), after flushing whatever is
+ * left over from a previous Stop() (see Work()).
+ *
+ * https://learn.microsoft.com/en-us/windows/win32/coreaudio/rendering-a-stream
+ */
 class WasapiOutputThread {
 	Thread thread{BIND_THIS_METHOD(Work)};
 	WinEvent event;
@@ -221,6 +229,14 @@ public:
 		std::size_t consumed = ring_buffer.WriteFrom(input);
 
 		if (!playing) {
+			/* Don't start playback until the endpoint
+			   buffer can be filled completely, or the
+			   device will play a short burst of audio
+			   followed by silence */
+			if (ring_buffer.ReadAvailable() <
+			    buffer_size_in_frames * frame_size) {
+				return consumed;
+			}
 			playing = true;
 			Play();
 		}
@@ -283,6 +299,16 @@ public:
 	}
 
 private:
+	/**
+	 * Stop the client; must be called on the worker thread while
+	 * started is true.  Anything left in the endpoint buffer is
+	 * flushed later, just before the client gets started again.
+	 */
+	void HaltClient() {
+		Stop(client);
+		started = false;
+	}
+
 	void SetStatus(Status s) noexcept {
 		status.store(s);
 		event.Set();
@@ -465,10 +491,19 @@ try {
 		}
 
 		if (cancel.load()) {
+			/* Stop the client or it will keep playing
+			   silence after the discard; Push() will
+			   restart it later.
+			   NOTE: Only in shared mode, in exclusive
+			   mode stopping here breaks playback */
+			if (started && !is_exclusive)
+				HaltClient();
+
 			ring_buffer.Discard();
 			cancel.store(false);
 			empty.store(true);
 			InterruptWaiter();
+			continue;
 		}
 
 		Status current_state = status.load();
@@ -479,16 +514,8 @@ try {
 			return;
 
 		case Status::PAUSE:
-			if (!started)
-				/* don't bother starting the
-				   IAudioClient if we're paused */
-				continue;
-
-			/* stop the IAudioClient while paused; it will
-			   be restarted as soon as we're asked to
-			   resume playback */
-			Stop(client);
-			started = false;
+			if (started)
+				HaltClient();
 			continue;
 
 		case Status::PLAY:
@@ -496,15 +523,13 @@ try {
 		}
 
 		UINT32 write_in_frames = buffer_size_in_frames;
-		DWORD mode = 0;
-		AtScopeExit(&) {
-			render_client->ReleaseBuffer(write_in_frames, mode);
 
-			if (!started) {
-				Start(client);
-				started = true;
-			}
-		};
+		/* Flush anything left over from a previous Stop(); the
+		   padding check below needs zero padding, or we would
+		   neither render nor start.  Do not do this directly
+		   after Stop(). */
+		if (!started)
+			Reset(client);
 
 		if (!is_exclusive) {
 			UINT32 data_in_frames =
@@ -522,6 +547,16 @@ try {
 		    FAILED(result)) {
 			throw MakeHResultError(result, "Failed to get buffer");
 		}
+
+		DWORD mode = 0;
+		AtScopeExit(&) {
+			render_client->ReleaseBuffer(write_in_frames, mode);
+
+			if (!started) {
+				Start(client);
+				started = true;
+			}
+		};
 
 		const UINT32 write_size = write_in_frames * frame_size;
 		std::span w{data, write_size};
--- a/src/output/plugins/wasapi/AudioClient.hxx
+++ b/src/output/plugins/wasapi/AudioClient.hxx
@@ -65,6 +65,14 @@ Stop(IAudioClient &client)
 }
 
 inline void
+Reset(IAudioClient &client)
+{
+	HRESULT result = client.Reset();
+	if (FAILED(result))
+		throw MakeHResultError(result, "Failed to reset client");
+}
+
+inline void
 SetEventHandle(IAudioClient &client, HANDLE h)
 {
 	HRESULT result = client.SetEventHandle(h);


src/output/plugins/wasapi/WasapiOutputPlugin.cxx
```diff
@@ -346,6 +346,10 @@ class WasapiOutput final : public AudioOutput {
 	bool Pause() override;
 	void Interrupt() noexcept override;
 
+	bool SupportsHardwarePause() const noexcept override {
+		return true;
+	}
```
MaxKellermann: I don't understand this change; the commit message doesn't explain it. Does this really have anything to do with the rest of this commit?

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
```diff
@@ -454,6 +458,13 @@ try {
 
 		UINT32 write_in_frames = buffer_size_in_frames;
 		DWORD mode = 0;
+
+		/* Reset to clear residual state that could
+		   cause stale samples. */
+		if (!started) {
+			Reset(client);
+		}
```
MaxKellermann: I don't understand this either. Where did stale data really come from?
And why do we need to call Reset() here (adding overhead to every iteration) and not at the start of the pause state?

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
```diff
@@ -426,10 +426,15 @@ try {
 		event.Wait();
 
 		if (cancel.load()) {
+			if (started && !is_exclusive) {
+				Stop(client);
```
MaxKellermann: Why? The commit message doesn't say.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
```diff
@@ -426,10 +426,15 @@ try {
...
+				started = false;
+			}
 			ring_buffer.Discard();
 			cancel.store(false);
 			empty.store(true);
 			InterruptWaiter();
+			continue;
```
MaxKellermann, line 506: What's the point of "getting a full cycle" and what does "getting a full cycle" mean? The commit message does't say.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
```diff
@@ -164,6 +164,7 @@ class WasapiOutputThread {
 	bool playing = false;
 
 	bool started = false;
+	bool primed = false;
```
MaxKellermann: What does this field mean and how it it protected across threads?

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
```diff
@@ -220,6 +221,8 @@ class WasapiOutputThread {
 		if (!playing) {
 			playing = true;
 			Play();
+		} else if (!primed) {
```
MaxKellermann: This field is written by another thread - without protection, you must not read it here.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
```diff
@@ -220,6 +221,8 @@ class WasapiOutputThread {
...
+			event.Set();
```
MaxKellermann: Why?

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
```diff
@@ -470,6 +473,16 @@ try {
 			Reset(client);
 		}
 
+		/* On cold start, wait for enough data to fill
+		   the entire endpoint buffer to avoid a pop. */
+		if (!primed) {
+			const UINT32 needed = buffer_size_in_frames * frame_size;
```
MaxKellermann: I believe waiting until the buffer is completely full is excessive, adds too much delay. We should only wait long enough to prevent the pop, but not more. How much do we need for that?

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
```diff
@@ -470,6 +473,16 @@ try {
...
+			if (ring_buffer.ReadAvailable() < needed) {
+				continue;
```
MaxKellermann: What will the device do if we skip the rest of the loop and don't write anything to the device's buffer?

MaxKellermann (changes requested):

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: Yes, it is needed. Without it Cancel() fires on pause and discards the ring buffer, which makes the Reset fix on resume useless. I did test without this block compiled in, the pops were in.

MaxKellermann: This sounds like trial'n'error - i.e. you did not understand what you were doing, and randomly changed things until they worked.
So, do you want to discard buffers or not? The other part of this commit deals with clearing buffers, but here you disabled it. This sounds like a contradiction to me, not a proper fix.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: "Stale data" is misleading indeed, I'll update the comment so it's clearer.

Reset() must run between Stop() and Start(), not at pause time, because the cancel (seek) path also calls Stop() independently.
This:
```diff
-   /* stop the IAudioClient while paused; it will
-      be restarted as soon as we're asked to
-      resume playback */
+   /* stop and reset the IAudioClient while
+      paused; it will be restarted as soon as
+      we're asked to resume playback */
    Stop(client);
+   Reset(client);
    started = false;
    continue;
```
creates pops on seek.

Overhead -> it is guarded by `if (!started)`, so it should only run once per pause/play cycle. Unless I'm misunderstanding something?

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: In shared mode, after Cancel(), the framework will close and reopen the output, creating a new thread with a fresh IAudioClient. If the old client is still running when the new one is created, both clients briefly compete for the same endpoint. One is playing discarded data, and new one starting fresh. This will cause a pop.

!is_exclusive guard is there because the audio simply doesn't play in exclusive mode with that fix in. I don't use exclusive mode, so I don't know the reason why it happens honestly. I just threw it in, seemed to work.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: Without it, execution falls through to the PLAY processing below, which would call GetBuffer() on the just-stopped client, fill it with zeros and call Start() -> pop.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: The field means the device has been primed, obviously. It's on line 476. I stole the idea from MPV, they do something similar in there so that there's no pop on startup.

I'll make it atomic.

MaxKellermann: I feared you'd say that, but atomic isn't a general-purpose magic spell that will solve all threading bugs. Think twice before using it. It's too easy to introduce new concurrency bugs with it.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: A full buffer is definitely needed to prevent the pop entirely. I didn't ~see~ hear any perceptible delay during testing. I'll experiment with only filling up to a point though.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
MaxKellermann: Oh, is this true, two can exist at the same time? That sounds dangerous, and we should ensure that the old threads finishes before reporting completion. In this context, your change sounds like a bad workaround.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
MaxKellermann: But the client was only stopped if it's not exclusive.
How will zeroes produce a pop? Zero is full silence.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
MaxKellermann: Is this behavior documented?
If this is true, then we can implement a big code simplification: then we don't need a ring buffer, and instead only send blocks of full buffers to the API.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: Pop isn't the zeroes, it's the transition from silence to audio, and continue prevents the worker from rendering a 0filled buffer in the first place.
At least that's how I intuited it, fact of the matter is that it **does** help with eliminating pops.
I'm sorry I can't give a better explanation. Let's call it another trial and error :)

Also should probably reset primed here somewhere.

MaxKellermann: Trail'n'error is not a good strategy, because it leads to an incomprehensible and fragile mountain of spaghetti code. We need a full understanding of what happens, or else we can't find a solid fix.

Where's the code that fills zeroes, anyway?

0x0003:
Converting to draft until I can come up with something better with clearer explanations. Most of the current state was done by "throw shit at the wall and see what sticks" principle, and even though it works, I don't understand most of what's going on behind the scenes really.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: WasapiOutputPlugin.cxx:519

So the flow is, as I understand it
```
1. Cancel handler: Stop(), started = false, ring_buffer.Discard() ->
2. Falls through to PLAY ->
3. !started, Reset(client)
4. (primed == true, my code) ->
5. ReadTo() reads 0 bytes (ring buffer empty) ->
6. std::fill_n fills entire buffer with zeros ->
7. ReleaseBuffer(buffer_size_in_frames, 0) ->
8. Start(), DAC plays full buffer of zeros ->
9. Next render, real audio, waveform discontinuity if the zeroes are behind
```
continue prevents steps 5-8.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: It is documented, yes.
https://learn.microsoft.com/en-us/windows/win32/coreaudio/rendering-a-stream
```
In its initial calls to the IAudioRenderClient::GetBuffer and IAudioRenderClient::ReleaseBuffer  
methods, the function fills the entire buffer before calling the IAudioClient::Start method  
to begin playing the buffer.
```

https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-initialize
```
Each time the thread awakens, it should call IAudioClient::GetCurrentPadding to determine  
 how much data to write to a rendering buffer
```

I don't believe getting rid of the ring buffer is at all feasible. How is output gonna call wasapi? It doesn't even have COM context. Among other things. Is this a tongue-in-cheek statement?
primed check only gates Start()

---

I fed [chromium wasapi implementation](https://chromium.googlesource.com/chromium/chromium/+/4cecfb2ac11410fc21f86d96802c9da6acd90bcd/media/audio/win/audio_low_latency_output_win.cc) to my local qwen llm (too lazy to dig through all of it myself, not like I'm working on chromium code anyway), and it said there's no ring buffer in there, so maybe you are right after all. I don't know how much of that is "simplification" though, looks like a massive overhaul to me. Definitely beyond my abilities.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: No, I was wrong. That was ass-backwards train of though, going from the symptom (pop on seek), and I just *assumed* it's two threads competing. Adding a change, compiling and testing got rid of the pop, so it "confirmed" my hypothesis, but that's not really what's happening there. Gotta be something way simpler.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: I think atomic is the correct solution here after all.

The alternatives are:
Removing !primed guard from Push() - it's gonna wake the worker even after priming is done which is not only wasteful, it also reintroduces pops. So not really an alternative, even.
Or a separate signaling mechanism, that's just extra complexity for no real benefit.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: Nothing, it hasn't started. Start() is at 501, just below this block.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
MaxKellermann: It prevents that by ignoring the device's request for more data (which is not good), and if you ignore that, what will lead to waking up the next loop iteration?

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
MaxKellermann: OK, it's good that we have documentation, and you should add a code comment explaining why you're waiting for the ring buffer to become full before starting the device (with a link to MSDN).
But it could be so much simpler, without the "primed" flag: just do the "full" check from inside Push() and inhibit the Play() call until the buffer is full. No new code in Work() at all. Simpler and more robust.

I think we still need the ring buffer, after all. Just wait for it to initially become full before starting the device.

MaxKellermann:
Throwing shit is a good approach for the first iteration, but before merging, we need a better understanding of the situation.

Certainly the existing code isn't as good as it should be, and suffers from the same problem; I guess the author didn't have a full understanding either. But new code solves problems, and never introduces new problems. But when I merge changes to existing code, that may introduce new problems, therefore merging code changes needs more thorough review. Because likely you will be gone when the bug reports start coming in, and then it's my problem.

I suggest you try to finish the "primed" patch first, but [a simpler version as I described in the comment](https://github.com/MusicPlayerDaemon/MPD/pull/2545#discussion_r3593174951). It is easy to see that this needs to be improved and how.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: Found justification to keep continue in there
https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-start

```
To avoid start-up glitches with rendering streams, clients should not call Start until the
audio engine has been initially loaded with data by calling the IAudioRenderClient::GetBuffer
and IAudioRenderClient::ReleaseBuffer methods on the rendering interface.
```

So ignoring "device's request" doesn't really apply, no events will fire after `Stop()` (the conversation above) anyway. We're just waiting for Push().

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-start
same quote I've given in the conversation below.

After Cancel() discards the ring buffer, the buffer is empty. If the client keeps running, it requests data via hw events and we'd haev fill with 0s, after that when real audio comes, the transition creates a pop. Stopping the client prevents that, we just wait for Push() to give us fresh data.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: After getting rid of "primed" and implementing your suggestion in Push() instead, this became not needed. 
I don't know why it helped before, and I'm not going to question it.

src/output/plugins/wasapi/WasapiOutputPlugin.cxx
0x0003: Got rid of it.

PR #2547, opened 2026-07-16
mpdconf.example: Fix path for sticker_file

Steap:


MaxKellermann:
backported to v0.24.x: 737edb8d523bd41b6fc9b8f5702b9ce5643a9782

PR #2549, opened 2026-07-17
output/osx: fix two format selection bugs, cleanups for the macOS code

CamilleScholtz:
The cleanup part of the macOS work; the format selection bug fixes were
split into #2550 and merged onto v0.24.x separately, so this branch was
rebased and now only carries the cleanups.

Errors now go through `Apple::ThrowOSStatus()` so they carry the
CoreAudio error description instead of a bare number, a new
`AudioObjectSetPropertyDataT()` wrapper mirrors the existing getter, the
buffer size uses `RoundUpToPowerOfTwo()`, and the macOS 12 "Main" rename
is handled with constexpr aliases for old SDKs instead of macros.

apple/meson.build switches to the "appleframeworks" dependency and
drops CoreServices, a leftover from the Component Manager API. Also
"#pragma once" in the apple headers and some unused includes removed.

Each commit builds on its own. Tested on macOS arm64 with and without
ENABLE_DSD.

### DIFF
--- a/src/apple/AudioObject.hxx
+++ b/src/apple/AudioObject.hxx
@@ -1,8 +1,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 // author: Max Kellermann <max.kellermann@gmail.com>
 
-#ifndef APPLE_AUDIO_OBJECT_HXX
-#define APPLE_AUDIO_OBJECT_HXX
+#pragma once
 
 #include "Throw.hxx"
 #include "util/AllocatedArray.hxx"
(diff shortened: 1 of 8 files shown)


MaxKellermann:
Which commits should be backported to v0.24.x? ("all" is a valid answer)
Can you move those to another PR, and rebase this one on my merge commit later?

CamilleScholtz:
Created a new commit with only the bug fixes

MaxKellermann:
Thanks, merged into both branches, you can rebase now.

PR #2553, opened 2026-07-29
decoder/dsdiff: reject overflowing chunk offsets

acts-1631:
dsdiff_read_metadata() reads a 64-bit chunk size from the DFF file and
adds it to the current input offset. If this addition overflows, the
result can point backward. For seekable input, dsdlib_skip_to() then
seeks back to the same chunk header, causing metadata scans to loop
indefinitely. Playback follows the same parsing path and can spin until
stopped.

Use checked offset addition in dsdiff_read_metadata(),
dsdiff_read_prop(), and dsdiff_read_prop_snd(). Also reject overflow
when applying DSDIFF even-byte padding and in dsdlib_skip() relative
seeks. Absolute backward seeks remain supported because the tag reader
uses them intentionally.

Verified with a 28-byte DFF file whose unknown chunk size wraps the
offset from 28 back to 16. The checked calculation now rejects the
chunk.

### DIFF
--- a/src/decoder/plugins/DsdiffDecoderPlugin.cxx
+++ b/src/decoder/plugins/DsdiffDecoderPlugin.cxx
@@ -21,6 +21,8 @@
 #include "tag/Handler.hxx"
 #include "DsdLib.hxx"
 
+#include <limits>
+
 struct DsdiffHeader {
 	DsdId id;
 	PackedBE64 size;
@@ -326,6 +328,10 @@ dsdiff_read_metadata(DecoderClient *client, InputStream &is,
 		} else {
 			/* ignore unknown chunk */
 			const offset_type chunk_size = chunk_header.GetSize();
+			if (chunk_size >
+			    std::numeric_limits<offset_type>::max() - is.GetOffset())
+				return false;
+
 			const offset_type chunk_end_offset =
 				is.GetOffset() + chunk_size;
 


src/decoder/plugins/DsdiffDecoderPlugin.cxx
```diff
@@ -46,9 +48,17 @@ struct DsdiffChunkHeader {
 	 * (http://www.sonicstudio.com/pdf/dsd/DSDIFF_1.5_Spec.pdf)
 	 * section 2.3.
 	 */
-	[[nodiscard]] constexpr
-	uint64_t GetPaddedSize() const noexcept {
-		return (GetSize() + 1) & ~uint64_t(1);
+	[[nodiscard]]
+	bool GetPaddedSize(uint64_t &padded_size) const noexcept {
```
MaxKellermann: This is horribly inelegant code and it's no longer obvious what this function really does. This needs API documentation. But probably it needs a rewrite to make it easier to understand. This whole PR looks like LLM slop.

src/decoder/plugins/DsdiffDecoderPlugin.cxx
```diff
@@ -107,15 +99,27 @@ dsdiff_read_prop_snd(DecoderClient *client, InputStream &is,
 		     offset_type end_offset)
 {
 	DsdiffChunkHeader header;
-	while (is.GetOffset() + sizeof(header) <= end_offset) {
+	while (is.GetOffset() < end_offset) {
+		if (end_offset - is.GetOffset() < sizeof(header))
```
MaxKellermann: How would it be possible for an attacker to produce an overflow here? With a file of 2^64-8 bytes size?

src/decoder/plugins/DsdiffDecoderPlugin.cxx
```diff
@@ -107,15 +99,27 @@ dsdiff_read_prop_snd(DecoderClient *client, InputStream &is,
...
-		offset_type chunk_end_offset = is.GetOffset()
-			+ header.GetPaddedSize();
-		if (chunk_end_offset > end_offset)
+		uint64_t chunk_size = header.GetSize();
+		if ((chunk_size & 1) != 0) {
+			if (chunk_size == std::numeric_limits<uint64_t>::max())
+				return false;
+
+			++chunk_size;
+		}
```
MaxKellermann: This block is very obscure code. Previously, it was obvious that this was about padding, but you removed the code that mentioned padding. This is still too complicated!

PR #2555, opened 2026-07-29
decoder/mad: bound Xing seek table allocations

acts-1631:
DecodeFirstFrame() uses the frame count from a Xing header as
max_frames. RunDecoder() then calls AllocateBuffers(), which eagerly
allocates frame_offsets and times arrays using that count.

The existing limit accepts 8,388,608 frames. On x86_64, the two arrays
request about 192 MiB. A crafted 1,189-byte MPEG file can supply this
count, and allocation failure terminates MPD because AllocateBuffers()
is noexcept.

Ignore Xing frame counts that cannot fit within the input even if every
frame were only a four-byte MPEG header. Unknown-size streams retain
the conservative estimate from FileSizeToSongLength(). Also use
nothrow allocation and clean up partial allocations so a remaining
failure stops this decoder instance instead of terminating MPD.

### DIFF
--- a/src/decoder/plugins/MadDecoderPlugin.cxx
+++ b/src/decoder/plugins/MadDecoderPlugin.cxx
@@ -690,10 +690,28 @@ MadDecoder::DecodeFirstFrame(Tag *tag) noexcept
 		mute_frame = MadDecoderMuteFrame::SKIP;
 
 		if ((xing.flags & XING_FRAMES) && xing.frames) {
-			mad_timer_t duration = frame.header.duration;
-			mad_timer_multiply(&duration, xing.frames);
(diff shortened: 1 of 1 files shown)


MaxKellermann:
This commit mixes several distinct changes that should be separate commits.

MaxKellermann:
1818bce17fcd0b82a86dd0c931a1ccc643bd9e30
and
56d4df723c273503c2f03cbf2d81d6bd682d245e

I didn't like the manual error handling you introduced, because C++ is all about not handling errors manually; this is fragile. Just lt the std::bad_alloc bubble through to the caller. The second patch now only removes the `noexcept`.

Issue #932, opened 2020-08-03
crash when disable an unavailable device


Wang-Yue:
### Describe the bug

```
audio_output {
 type "osx"
 device "E30 "
 dop "yes"
 hog_device "true"
 name "Topping E30"
 mixer_type "none"
}
```

- connect usb dac
- play some music
- disconnect usb (*)
- go to device, disable that usb device (**)

## Expected Behavior

-  In (*), music should stop. 
- In (**), it should be able to successfully disable the device.

## Actual Behavior

- In (*), the music will still play in system device
- In (**), mpd crashes.

## Version

This bug presents in 0.21.25 as well. Here I'm using the trunk version for testing (so that I can provide you a stack trace)
```
Music Player Daemon 0.22~git (v0.21.25-1041-gdbdf782e5+)
Copyright 2003-2007 Warren Dukes <warren.dukes@gmail.com>
Copyright 2008-2018 Max Kellermann <max.kellermann@gmail.com>
This is free software; see the source for copying conditions.  There is NO
warranty; not even MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Database plugins:
 simple proxy upnp

Storage plugins:
 local nfs curl

Neighbor plugins:
 upnp

Decoders plugins:
 [vorbis] ogg oga
 [oggflac] ogg oga
 [flac] flac
 [opus] opus ogg oga
 [sndfile] wav aiff aif au snd paf iff svx sf voc w64 pvf xi htk caf sd2
 [dsdiff] dff
 [dsf] dsf
 [hybrid_dsd] m4a
 [faad] aac
 [fluidsynth] mid
[...]

PR #2556, opened 2026-07-30
output/osx: move the default device query into a function

valentynkit:
Disconnect a Bluetooth headset during playback and the output never opens again, even once the headset is back and is the default output device:

```
mixer: Failed to read mixer for "CoreAudio": The operation couldn't be completed. (OSStatus error 560947818.)
exception: Failed to open "CoreAudio" (osx); The operation couldn't be completed. (OSStatus error 560947818.)
player: problems opening audio device while playing "..."
```

That is `kAudioHardwareBadObjectError`. `Create()` resolves `dev_id` once, and `osx_output_set_device()` only refreshes it for `kAudioUnitSubType_HALOutput`. With `kAudioUnitSubType_DefaultOutput` or `kAudioUnitSubType_SystemOutput` it is never updated, so once that device is gone the id stays dead and `GetVolume()`, `SetVolume()` and `Open()` all fail until a restart.

Re-resolving in `Enable()` would not help either. `InternalEnable()` returns early once `really_enabled` is set, and the open failure path never clears it, so `Enable()` runs once per process.

Since both subtypes track the default device anyway, query it on demand instead of caching it. The `hog_device` calls keep using the stored id so `Disable()` still releases the device `Enable()` hogged.

#932 hit the same error in 2020. Its case (\*), a device pinned by name disappearing, needs different handling and is not addressed here.

First commit is preparation, no functional change. Tested on macOS arm64 with `mixer_type "hardware"` and no `device` set. Before, the output stays unopenable until restart. After, playback recovers on the next open and the log stays clean. Full build, no new warnings.

Both commits apply to v0.24.x if you want a separate PR for that branch.

### DIFF
--- a/src/output/plugins/OSXOutputPlugin.cxx
+++ b/src/output/plugins/OSXOutputPlugin.cxx
@@ -143,11 +143,12 @@ OSXOutput::OSXOutput(const ConfigBlock &block)
 	}
 }
 
-AudioOutput *
-OSXOutput::Create(EventLoop &, const ConfigBlock &block)
+/**
+ * Query the current default output device.
+ */
+static AudioDeviceID
+GetDefaultOutputDevice(OSType component_subtype)
 {
-	OSXOutput *oo = new OSXOutput(block);
-
 	static constexpr AudioObjectPropertyAddress default_system_output_device{
 		kAudioHardwarePropertyDefaultSystemOutputDevice,
 		kAudioObjectPropertyScopeOutput,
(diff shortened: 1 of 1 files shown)


MaxKellermann:
The first commit looks like a good refactoring improvement, I like that.

I'm cautious about the second commit: it ignores `dev_id`, but never updates it. Is that a good idea? This seems like a bad smell. Are other uses of `dev_id` safe with this problem?

MaxKellermann:
First commit cherry-picked to v0.24.x (together with a few code cleanup backports from master to avoid conflict): 470a32485c0071332242ef40cbcb54345603425c

PR #2569, opened 2026-08-17
doc/protocol.rst: fix MPD version on `tagtypes` sub commands

SoongNoonien:


MaxKellermann:
Cherry-picked to v0.24.x: 2fb543528ac23b4ea4c19173a47e63187bc1997a

Note this doesn't "fix" anything - there was no wrong version annotation that needed fixing, there was none at all. The commit message is misleading.

SoongNoonien:
There was no version number annotated while other commands where annotated with the version number which was missing here. So, it fixes exactly this. Earlier the docs implied that these commands worked even before 0.14, which is the oldest version annotated anywhere in the docs.

But I would have been totally fine with you (or me) changing the commit message. To avoid this in the future, what would have been a better commit message in your opinion?

MaxKellermann:
Maybe "add version annotations to `tagtypes` sub commands".

Note I had merged your commit as-is with your commit message, because I think it's a good improvement.

PR #2570, opened 2026-08-18
util: allow relative URI to ascend to relative root

kmhofmann:
`uri_apply_relative()` currently cannot resolve a `../` reference when the base directory consists of exactly one relative path component.

For example:

```text
relative: ../foo
base:     bar/baz
expected: foo
```

`ConsumeLastSegment()` removes the trailing slash from `bar/`, then fails because there is no remaining slash. Consuming that single relative component is valid and should leave the relative root.

This is observable when a CUE sheet and its referenced audio file are located directly in `music_directory`. MPD indexes the virtual CUE tracks correctly, but playback fails with:

```text
Failed to decode "album.cue/track0001": no decoder plugin
```

The same CUE sheet and audio file work correctly when placed one directory below `music_directory`.

The change allows a one-component relative path to be consumed while continuing to reject attempts to ascend beyond either the relative or absolute root.

Regression tests cover both cases:

```cpp
{ "../foo", "bar/baz", "foo" },
{ "../../foo", "bar/baz", "" },
```

I also verified the fix with an actual CUE/FLAC pair located directly in `music_directory`.


src/util/UriRelative.cxx
```diff
@@ -84,14 +84,21 @@ StripLeadingSlashes(std::string_view &s) noexcept
 static bool
 ConsumeLastSegment(std::string_view &path) noexcept
 {
-	assert(!path.empty());
+	if (path.empty())
+		return false;
+
```
MaxKellermann: Why?

src/util/UriRelative.cxx
```diff
@@ -84,14 +84,21 @@ StripLeadingSlashes(std::string_view &s) noexcept
...
 	assert(path.back() == '/');
 
+	if (path.size() == 1)
```
MaxKellermann: Why?

src/util/UriRelative.cxx
```diff
@@ -84,14 +84,21 @@ StripLeadingSlashes(std::string_view &s) noexcept
...
+		return false;
+
 	path.remove_suffix(1);
 
 	const auto slash = path.rfind('/');
-	if (slash == path.npos)
-		return false;
+	if (slash == path.npos) {
+		path = path.substr(0, 0);
```
MaxKellermann: This is flawed because now the path no longer ends with a slash.

MaxKellermann (changes requested):

src/util/UriRelative.cxx
kmhofmann: Agreed. The intention was to represent the relative root as an empty path, but that breaks the existing invariant of `ConsumeLastSegment()`. I’ve revised the patch to leave `ConsumeLastSegment()` unchanged and handle the transition to the relative root in `ConsumeSpecial()` instead.

src/util/UriRelative.cxx
kmhofmann: These two changes were needed by the previous approach, where an empty path represented the relative root inside `ConsumeLastSegment()`. With the revised implementation they are no longer necessary and have been reverted.

src/util/UriRelative.cxx
kmhofmann: These two changes were needed by the previous approach, where an empty path represented the relative root inside `ConsumeLastSegment()`. With the revised implementation they are no longer necessary and have been reverted.

MaxKellermann:
Good fix, but was still too complicated for me - here's a simplification on top of it: 099971ba1ef00638835c43d9fa02bf7b1ee0f483

PR #2575, opened 2026-08-21
playlist/pls: limit input size

acts-1631:
The PLS parser materializes up to 65,536 entries before returning any songs. Each entry can retain a file URI and title near the text line limit, allowing a playlist to consume hundreds of MiB.

Add an optional byte limit to `TextInputStream` and reject PLS input over 16 MiB. Known-size streams are rejected before parsing, and the read limit also covers streams without a known size. I can adjust the 16 MiB limit to something else if you prefer.

### DIFF
--- a/src/input/LimitedInputStream.cxx
+++ b/src/input/LimitedInputStream.cxx
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The Music Player Daemon Project
+
+#include "LimitedInputStream.hxx"
+
+#include <cassert>
+#include <stdexcept>
+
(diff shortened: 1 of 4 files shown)


MaxKellermann:
I don't like that you added this feature to the `TextInputStream` class. I think we would be better off by keeping this out of the class (it's out of feature scope), and instead add a proxy `InputStream` class which limits the amount of data as its only feature.

acts-1631:
Alright. I have moved the limit into a dedicated `LimitedInputStream` proxy and restored `TextInputStream` to its original scope. The PLS parser now wraps its input with the proxy.

MaxKellermann:
Nice!

PR #2576, opened 2026-08-24
input: wake callers on errors and EOF

acts-1631:
`AsyncInputStream::Read()` waits on `caller_cond`, but EOF and error callbacks in asynchronous input implementations only called `InvokeOnAvailable()`. After `bd7830794`, that no longer wakes synchronous readers. Commands that enumerate a truncated remote playlist can therefore leave MPD's main thread blocked indefinitely.

Make `AsyncInputStream` notify `caller_cond` whenever it invokes the availability handler, covering all asynchronous implementations. Remove `noexcept` from the playlist enumeration helpers so I/O errors reach the command handler instead of terminating MPD after the reader is woken.

### DIFF
--- a/src/input/plugins/AlsaInputPlugin.cxx
+++ b/src/input/plugins/AlsaInputPlugin.cxx
@@ -244,6 +244,7 @@ try {
 	CommitWriteBuffer(nbytes);
 }
 catch (...) {
+	const std::lock_guard protect{mutex};
 	postponed_exception = std::current_exception();
 	InvokeOnAvailable();
 }
--- a/src/playlist/Length.cxx
+++ b/src/playlist/Length.cxx
@@ -28,7 +28,7 @@ static void
 playlist_provider_length(Response &r,
 			const SongLoader &loader,
 			const char *uri,
-			SongEnumerator &e) noexcept
+			SongEnumerator &e)
 {
 	const auto base_uri = uri != nullptr
 		? PathTraitsUTF8::GetParent(uri)
--- a/src/playlist/Print.cxx
+++ b/src/playlist/Print.cxx
@@ -27,7 +27,7 @@ playlist_provider_print(Response &r,
 			SongEnumerator &e,
 			unsigned start_index,
 			unsigned end_index,
-			bool detail) noexcept
+			bool detail)
 {
 	const auto base_uri = uri != nullptr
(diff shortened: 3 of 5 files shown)


src/input/plugins/AlsaInputPlugin.cxx
```diff
@@ -244,6 +244,7 @@ try {
 	CommitWriteBuffer(nbytes);
 }
 catch (...) {
+	const std::lock_guard protect{mutex};
```
MaxKellermann, line 247: This change is correct, but hiding it inside this unrelated commit is not. This should be a separate commit with a commit message explaining the fix.

src/playlist/Length.cxx
```diff
@@ -28,7 +28,7 @@ static void
 playlist_provider_length(Response &r,
 			const SongLoader &loader,
 			const char *uri,
-			SongEnumerator &e) noexcept
+			SongEnumerator &e)
```
MaxKellermann, line 31: Huh? Unrelated undocumented change?

src/playlist/Print.cxx
```diff
@@ -27,7 +27,7 @@ playlist_provider_print(Response &r,
 			SongEnumerator &e,
 			unsigned start_index,
 			unsigned end_index,
-			bool detail) noexcept
+			bool detail)
```
MaxKellermann, line 30: Huh? Unrelated undocumented change?

src/playlist/Print.cxx
```diff
@@ -61,7 +61,7 @@ playlist_provider_search_print(Response &r,
 			       SongEnumerator &e,
 			       unsigned start_index,
 			       unsigned end_index,
-			       SongFilter *filter) noexcept
+			       SongFilter *filter)
```
MaxKellermann, line 64: Huh? Unrelated undocumented change?

acts-1631:
Split these changes into separate documented commits. The `noexcept` removals are part of the reproduced failure path: after `AsyncInputStream` wakes the reader, `NextSong()` propagates the curl error, and `noexcept` would terminate MPD instead of allowing the command handler to report it. The ALSA mutex correction is now isolated in its own commit.

MaxKellermann:
Much better, thanks.

